PR #1, opened 2021-12-13
IDE support

srid:
- add `shell.nix` for VSCode nix-env-selector which doesn't support flakes
- add HLS to nix-shell, as well the required hie.yaml

If caching becomes an issue, I'd suggest that we create a cachix account for Plutonomicon (or some other shared cache).


nixinator:
i'm in the process of setting this up @srid .

PR #4, opened 2021-12-14
Run fourmolu 

srid:
- Also adds `bin/format` to do this before commits.
- The CI checks that it was run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+# This can be run manually like so:
+#
+#     gh workflow run ci.yml --ref <git_branch_name>
+#
+name: Continuous Integration
+on:
+  pull_request:
+  push:
+  workflow_dispatch:
+jobs:
+  build:
+    name: nix-build
+    runs-on: self-hosted
+    steps:
+
+    - name: Checkout repository to $GITHUB_WORKSPACE
+      uses: actions/checkout@v2.3.4
+
+    - run: nix develop --command "./bin/ci-format-check"
+    - run: nix develop --command cabal build all
+    - run: nix develop --command cabal test all
+    - name: Check Haskell Language Server 
+      run: nix develop --command haskell-language-server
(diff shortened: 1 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+#
+name: Continuous Integration
+on:
+  pull_request:
+  push:
+  workflow_dispatch:
+jobs:
+  build:
+    name: nix-build
+    runs-on: self-hosted
```
srid, line 13: @nixinator Can we setup self-hosted CI for this repo? Does that automatically give us nix cache? Otherwise `runs-on: ubuntu-latest` may be sufficient.

nixinator: were using herc-ci ,i'll give you setup details soon.

nixinator: i need to find out how to do the tests you need in herc-ci. So bear with me. I really want to go the herc route because the runners are some much more nix compatible, self hosted github runners are PITA.

srid: Sure, I think hercules-ci is generally better. Can the nix cache be shared with MLabs folks as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+
+    - name: Checkout repository to $GITHUB_WORKSPACE
+      uses: actions/checkout@v2.3.4
+
+    - run: nix develop --command "./bin/ci-format-check"
```
L-as, line 19: This is a big misuse of `nix develop` IMO. We want to test `nix build`, `nix develop` isn't pure, however, it isn't worth fixing right now as I assume that @nixinator will be doing some hercules stuff anyway.

nixinator:
can we split these commits out, to changes to the haskell code base, and the changes to CI, unless there something i'm missing that it all has to be in one commit right now?

Issue #6, opened 2021-12-14
Add CI to Plutarch


srid:
Should run the following in a nix-shell (effectively):

```
./bin/ci-format-check
haskell-language-server
```

Also run `nix build` (ensuring that `cabal test` is run when building the Haskell package for Plutarch).

The current GitHub CI may be replaced.

srid:
> ./bin/ci-format-check

There may be better way to do this.

nixinator:
This is looking like herc-ci + flakes can solve this problem of running tests is herc.

srid:
@nixinator https://hercules-ci.com/github/Plutonomicon/plutarch/jobs/1 should now succeed, now that `master` has a default.nix.

nixinator:
https://hercules-ci.com/github/Plutonomicon/plutarch/jobs/5

`Evaluation failed
Please make sure your ci.nix has at least one derivation.`

https://docs.hercules-ci.com/hercules-ci/troubleshooting/#why-is-my-attribute-not-in-eval

hmmmm, must be the structure of this flake. Investigating.

srid:
There is a `default.nix` in the repo and Hercules _should_ be using it,
https://docs.hercules-ci.com/hercules-ci-agent/evaluation/

> At the top-level of your repository, the agent will search for a Nix file to evaluate in the following order: nix/ci.nix, ci.nix or default.nix

Sounds like an upstream regression to me. Does Robert know?

srid:
Oh https://github.com/hercules-ci/flake-compat-ci

See #10

L-as:
Thank you @MatthewCroughan for https://github.com/NixOS/templates/pull/12

nixinator:
> Thank you @MatthewCroughan for [NixOS/templates#12](https://github.com/NixOS/templates/pull/12)

``` ______________ 
< nice one Las >
 -------------- 
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

```

PR #8, opened 2021-12-14
Actually evaluate a script

srid:
Quick change to evaluate a simple Plutarch script.

(This actually fails to evaluate, see #9. Also note that the fib test is failing; all this goes to indicate the importance of #6 )

### DIFF
--- a/Plutarch/Evaluate.hs
+++ b/Plutarch/Evaluate.hs
@@ -1,9 +1,12 @@
-module Plutarch.Evaluate (evaluateScript) where
+{-# LANGUAGE TypeApplications #-}
+
+module Plutarch.Evaluate (eval, evalWithArgs) where
 
 import Control.Monad.Except (MonadError, throwError)
 import Data.Text (Text)
 import Plutus.V1.Ledger.Scripts (Script)
 import qualified Plutus.V1.Ledger.Scripts as Scripts
+import qualified PlutusCore.Data as PLC
 import PlutusCore.Evaluation.Machine.ExBudget (ExBudget)
 import PlutusTx.Evaluation (evaluateCekTrace)
 import UntypedPlutusCore (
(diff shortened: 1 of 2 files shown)


L-as:
I want this to be fixed before this is merged, as it's likely `evaluateScript` that is problematic. If you look at it, you'll notice that it converts it to a named AST, but that conversion isn't actually correct, perhaps because we're calling it wrong. The evaluation should work once you fix that.

L-as (changes requested):
See comment.

srid:
> I want this to be fixed before this is merged, as it's likely `evaluateScript` that is problematic. If you look at it, you'll notice that it converts it to a named AST, but that conversion isn't actually correct, perhaps because we're calling it wrong. The evaluation should work once you fix that.

I don't think this is correct. Why do you think `evaluateScript` (named AST conversion is internal to it) is problematic here? We copied it verbatim (except for error handling, and returning result) from Plutus which uses this function to execute the validator `Script` (see [here](https://github.com/input-output-hk/plutus/blob/27e29cf134ddbf41fba6a51b173bd0e1aff71c28/plutus-ledger-api/src/Plutus/V1/Ledger/Scripts.hs#L407-L416) and [here](https://github.com/input-output-hk/plutus/blob/27e29cf134ddbf41fba6a51b173bd0e1aff71c28/plutus-ledger-api/src/Plutus/V1/Ledger/Scripts.hs#L194-L206)). Furthermore, Pluto generates the correct `Script` that executes correctly by `evaluateScript` *as well as* by Plutus' emulator and contract test ([see here](https://github.com/Plutonomicon/pluto/tree/main/examples/contracts/sample)).

My guess was that debruijn index calculation was wrong somewhere, but I'll have to poke a bit more on this.

L-as:
7661c899958c5c09760efd18fdcaef613f41acaa

PR #10, opened 2021-12-15
Add ci.nix

L-as:


L-as:
@nixinator @MatthewCroughan I don't think the cache is set up. This evaluation is taking ages. It's probably building GHC.

MatthewCroughan:
> @nixinator @MatthewCroughan I don't think the cache is set up. This evaluation is taking ages. It's probably building GHC.

Should we rely on IOHK cache? I'd like to know if things build from source or not. If it fails, that's more interesting. Once cached on our cachix, we won't need IOHK anyway.

L-as:
That is true, let's leave this then and see what happens.

L-as:
It finished!

nixinator:
@L-as the runners not pulling from cachix, so it's rebuild, investigating so bear with me

PR #11, opened 2021-12-15
Add some builtin types + utilities for constructing and working with them

TotallyNotChase:
* Add `PBuiltinString`, representing plutus core strings. Alongside instances and utilities for constructing terms of it.
* Add `PUnit`, representing plutus core unit. Alongside instances and utilities for constructing terms of it.
* Add `PEq` and `POrd` instances for `PBuiltinByteString`
* Add `preadByteStr` for reading hex strings as a `PBuiltinByteString` term.
* Add `Semigroup` and `Monoid` instances for `PBuiltinByteString` and `PBuiltinString` terms.

### Notes
* Should add some tests for these
* Probably add `PIsString`, `PSemigroup`, `PMonoid` typeclasses and use those instead.

### DIFF
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -73,6 +73,7 @@ hexDigitToWord8 :: HasCallStack => Char -> Word8
 hexDigitToWord8 = f . toLower
   where
     f '0' = 0
+    f '1' = 1
     f '2' = 2
     f '3' = 3
     f '4' = 4
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -1,6 +1,7 @@
 module Plutarch.Builtin (PData (..), pfstBuiltin, psndBuiltin, pasConstr, preadByteStr, PBuiltinPair, PBuiltinList, PBuiltinByteString, PBuiltinString) where
 
 import qualified Data.ByteString as BS
+import qualified Data.Text as Txt
 import Data.Char (toLower)
 import Data.String (IsString (..))
 import Data.Word (Word8)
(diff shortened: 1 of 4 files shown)


Plutarch/Builtin.hs
```diff
@@ -19,6 +22,22 @@ instance POrd PBuiltinByteString where
   x £<= y = punsafeBuiltin PLC.LessThanEqualsByteString £ x £ y
   x £< y = punsafeBuiltin PLC.LessThanByteString £ x £ y
 
+-- | Errors that may arise while using 'preadByteStr'.
+data ByteStringReadError = UnevenLength | InvalidHexDigit Char
+  deriving stock (Eq, Ord, Read, Show)
+
+-- | Interpret a hex string as a PBuiltinByteString.
+preadByteStr :: String -> Either ByteStringReadError (Term s PBuiltinByteString)
```
L-as, line 38: We probably want a better name. `phexByteStr`? Also, while it in general is good practice to use total functions, in this case none of the ways of constructing the AST support monadisms, hence I think it's better to have `HasCallStack =>` and make the function partial.

Plutarch/Builtin.hs
```diff
@@ -1,6 +1,7 @@
 module Plutarch.Builtin (PData (..), pfstBuiltin, psndBuiltin, pasConstr, preadByteStr, PBuiltinPair, PBuiltinList, PBuiltinByteString, PBuiltinString) where
 
 import qualified Data.ByteString as BS
+import qualified Data.Text as Txt
```
L-as, line 4: Please run `bin/format`.

L-as (changes requested):
Looks mostly solid.

Plutarch/Builtin.hs
L-as: Feel free to choose a name yourself, but it should be clear from the name how the string is parsed.

L-as:
> Probably add PIsString, PSemigroup, PMonoid typeclasses and use those instead.
We don't need to do this AFAICT.

Plutarch/Builtin.hs
TotallyNotChase: Went with `phexByteStr`. And made it partial with as suggested.

TotallyNotChase:
> > Probably add PIsString, PSemigroup, PMonoid typeclasses and use those instead.
> 
> We don't need to do this AFAICT.

We don't, yes. But perhaps it'd be more consistent with the existence `PEq` and `POrd`. It hardly matters though.

L-as:
> We don't, yes. But perhaps it'd be more consistent with the existence `PEq` and `POrd`. It hardly matters though.

In the case of `PEq` and `POrd`, these are *actually* needed. You can not implement `Ord` and `Eq` for these types, because the result isn't a `Bool`, but rather a `Term s PBool`.

PR #13, opened 2021-12-16
Fix phoistAcyclic

L-as:


L-as:
@nixinator What is happening here? The Ci is not running.

nixinator:
it's running, but it has to build on every commit, and your about killing my machine. I'm throwing together some better infrastructure right now.....

PR #14, opened 2021-12-17
Fix `IfThenElse` not forcing sufficiently

TotallyNotChase:
```hs
test :: Term s PInteger
test = pif (pcon PTrue) 1 2

-- >>> eval test
-- Left (EvaluationException "An error has occurred:  error:\nA builtin received a term argument when something else was expected\nCaused by: (ifThenElse True)" "UnexpectedBuiltinTermArgumentMachineError")
-- >>> printTerm test
-- "(program 1.0.0 (force (ifThenElse True (delay 1) (delay 2))))"
```
`IfThenElse` was missing a force. Should be similar to- `! (! IfThenElse True (# 1) (# 2))`


L-as:
Thanks!

PR #20, opened 2021-12-20
Add some more utilities

TotallyNotChase:
* Add plutarch synonyms to `PLC.HeadList`, `PLC.TailList`, and `PLC.NullList` for working with built in lists.
* Add synonyms for the remaining `BuiltinData` deconstruction builtins.
* Add `pbyteStr` - for creating a `PByteString` term from `ByteString`.

PR #23, opened 2021-12-21
Add type-safe Builtins; as well as polymorphic List & Pair builtins

srid:


srid:
Note that `PBuiltinList` and `PBuiltinPair`, being isomorphic (and without PlutusType implementation), are replaced by `PList` and `PPair` types (which are in turn used in `Builtin` GADT) in this PR. There is a bit of overlap with PR #20 by @TotallyNotChase.

TotallyNotChase:
I am in favor of sorting this one out and getting it merged before #20

Plutarch/Builtin.hs
```diff
@@ -1,27 +1,122 @@
...
-pfstBuiltin :: Term s (PBuiltinPair a b :--> a)
-pfstBuiltin = phoistAcyclic $ pforce . pforce . punsafeBuiltin $ PLC.FstPair
+{- | Type spec for PLC's untyped builtin functions
+
+ Note: `forces` determines the repeated application of `FORCE` when evaluating
+ a polymorphic builtin function.
+
+ Example: (UnConstrData #£ someData)
+-}
+data PBuiltin (forces :: Nat) (args :: [k -> Type]) (res :: k -> Type) where
```
L-as, line 38: Why do we need this?

srid: Type-safety. Automatically gives us typed builtin functions. For example,

```haskell
chooseData ::
  forall k (s :: k) (c' :: k -> Type).
  Term s PData ->
  Term s (PDelayed c' :--> PDelayed c' :--> PDelayed c' :--> PDelayed c' :--> PDelayed c' :--> PDelayed c')
chooseData = (B.ChooseData #£)
```

Compare to untyped:

```haskell
chooseDataUntyped :: forall k (s :: k) (a :: k -> Type). Term s a
chooseDataUntyped = punsafeBuiltin PLC.ChooseData
```

Plus, the GADT also encodes the number of forces required at the type-level.

L-as: I don't see how this is better than writing the typed functions manually.

srid: Even if you explicitly add the type of `Term` to `chooseDataUntyped` and other builtins -- and I doubt people will do that (given the 21 instances of untyped uses at the moment[^1]) --  callers would still have to remember to `force` the builtin.

I find it much more ergonomic and safer to use this.

[^1]: .
       ```
       ❯ rg -I 'punsafeBuiltin PLC\.' | wc -l
       21
       ```

L-as: I'm not sure what you mean? This is the approach we've been using until now: https://github.com/Plutonomicon/plutarch/pull/20/files#diff-74f4e5889ff8a668263439d5d956f02e7f56dcf0cf7f27987f3e2e4d0672d60eR32

srid: No, that's from the PR (`main` branch still uses untyped ones). 

I think @TotallyNotChase's approach begins to address the same problem, but it is a matter of preference as to whether you want to write `(B.TailList #£)` or `ptailBuiltin`. I personally prefer the former, because I don't have to learn yet another literal, and can just reuse the Plutus Core's naming. Regardless, `punsafeBuiltin` should not be exposed outside of where the typed builtins are defined.

L-as: https://github.com/Plutonomicon/plutarch/blob/9d08fcfc4b9cf260fd197cb7688a0a29b43d71cc/Plutarch/Bool.hs#L31

srid: Sure - and there are 21 instances of untyped. Why? :-P My point still stands.

Plutarch/Builtin.hs
```diff
@@ -1,27 +1,122 @@
...
+import Plutarch.Bool
+import Plutarch.Builtin.List.Type
+import Plutarch.Builtin.Pair.Type
 import Plutarch.ByteString (PByteString)
 import Plutarch.Integer (PInteger)
 import Plutarch.Prelude
+import Plutarch.String
 import qualified PlutusCore as PLC
 
-data PBuiltinPair (a :: k -> Type) (b :: k -> Type) (s :: k)
```
L-as, line 9: Please don't change the name nor remove this. We want a separate `PPair` eventually.

Plutarch/Builtin.hs
```diff
@@ -1,27 +1,122 @@
...
+  IData :: PBuiltin Nat0 '[PInteger] POpaque
+  UnIData :: PBuiltin Nat0 '[POpaque] PInteger
+  Trace :: PBuiltin Nat1 '[PString, a] a
+
+-- Haskell function type for a Plutus builtin.
+type family PBuiltinType (args :: [k -> Type]) (res :: k -> Type) where
+  PBuiltinType '[] res = res
+  PBuiltinType (a ': as) res = a :--> PBuiltinType as res
+
+pBuiltinTerm ::
```
L-as, line 58: All of this is much too complex for negligible benefit.

Plutarch/Builtin.hs
```diff
@@ -1,27 +1,122 @@
...
 
-psndBuiltin :: Term s (PBuiltinPair a b :--> b)
-psndBuiltin = phoistAcyclic $ pforce . pforce . punsafeBuiltin $ PLC.SndPair
+pTrace :: Text -> Term s a -> Term s a
+pTrace s f = Trace #£ pfromText s £ f
 
-pasConstr :: Term s (PData :--> PBuiltinPair PInteger (PBuiltinList PData))
-pasConstr = punsafeBuiltin PLC.UnConstrData
+(!£) :: forall k (s :: k) (a :: k -> Type). Text -> Term s a -> Term s a
+(!£) = pTrace
```
L-as, line 121: I don't see why we want an operator for this.

Plutarch/Builtin/List.hs
```diff
@@ -0,0 +1,111 @@
...
+import Plutarch
+import Plutarch.Bool
+import Plutarch.Builtin
+import Plutarch.Builtin.List.Type
+import Plutarch.Integer (PInteger)
+import Plutarch.Prelude
+import qualified PlutusCore as PLC
+import Prelude hiding (head)
+
+instance ListElemUni (a :: k -> Type) => PlutusType (PList a) where
```
L-as, line 30: If this is the only way of doing this, then we should not do it and instead use Scott-encoded lists in general.

Plutarch/Builtin/List.hs
```diff
@@ -0,0 +1,111 @@
...
+      (punsafeCoerce isEmpty)
+      (f PNil)
+      $ plet
+        (HeadList #£ list)
+        ( \head ->
+            plet (TailList #£ list) $ \tail ->
+              f $ PCons head tail
+        )
+
+cons :: forall k (s :: k) (a :: k -> Type). ListElemUni a => Term s a -> Term s (PList a) -> Term s (PList a)
```
L-as, line 52: Please follow the naming convention.

L-as: The type is in addition also incorrect AFAICT.

https://github.com/input-output-hk/plutus/blob/master/plutus-metatheory/src/Algorithmic.lagda#L154

Plutarch/Builtin/List.hs
```diff
@@ -0,0 +1,111 @@
...
+        perror
+      PCons x xs ->
+        pif
+          (n' £== 0)
+          x
+          (self £ (n' - 1) £ xs)
+
+-- | Append two lists
+append :: ListElemUni a => ClosedTerm (PList a :--> PList a :--> PList a)
+append =
```
L-as, line 105: You need to hoist this and basically all other functions too.

Plutarch/Builtin/List/Type.hs
```diff
@@ -0,0 +1,48 @@
...
+import Plutarch.Bool (PBool)
+import Plutarch.ByteString (PByteString)
+import Plutarch.Integer (PInteger)
+import Plutarch.Prelude (Type)
+import Plutarch.String (PString)
+import Plutarch.Unit (PUnit)
+import qualified PlutusCore as PLC
+import qualified PlutusCore.Data as PLC
+
+data PList a s
```
L-as, line 19: Please don't move just the data type definition into a separate file. This means you make orphan instances, in addition you generally always want to use both modules in tandem, never just one.

examples/Main.hs
```diff
@@ -44,14 +47,39 @@ fib = phoistAcyclic $
...
+      $ pif
+        (n £== 1)
+        (BL.mkList [1, 0])
+        $ plet (self £ (n - 1)) $ \a ->
+          plet (self £ (n - 2)) $ \b ->
+            BL.cons
+              (BL.head a + BL.head b)
+              a
+
+swap :: (a ~ POpaque, b ~ POpaque) => Term s (B.PPair a b :--> B.PPair b a)
```
L-as, line 65: Why are you introducing type parameters just to require an equality?

flake.nix
```diff
@@ -43,6 +43,7 @@
 
             tools = {
               haskell-language-server = {};  # Must use haskell.nix, because the compiler version should match
+              ghcid = {};
```
L-as, line 46: In general I want to reduce the number of tools we add here, as it can severely increase the time needed to do `nix develop`. Can't you get `ghcid` externally?

L-as (changes requested):

Plutarch/Builtin/List.hs
TotallyNotChase: We will need to be able to deconstruct built in lists either way. I understand that this machinery is for *building* lists rather than deconstructing them. But, IMO, we will need a way for building builtin lists along the way as well (see: `ListData`).

Unless we are 100% sure no one will ever ask for the ability to create builtin lists - I sincerely suggest we keep a *complete* builtin list implementation with great ergonomics for both building and deconstructing.

TotallyNotChase:
Do you guys think the indirections/abstractions (all for ergonomics and type safety, of course) will be detrimental to performance? I am going to run the numbers on this branch and compare them with the implementation I have currently in a client's repo, just to get an idea. I am all for safety abstractions but I am curious about its implications.

To be clear, I'm talking about stuff like-
```hs
-- | Return the head of a list; fails on empty list.
head :: forall k (s :: k) (c :: k -> Type). Term s (PList c) -> Term s c
head list =
  pmatchList list $ \case
    PNil -> perror
    PCons x _ -> x
```
compared to-
```hs
pheadBuiltin :: Term s (PBuiltinList a :--> a)
pheadBuiltin = phoistAcyclic $ pforce $ punsafeBuiltin PLC.HeadList
```

I really hope we can have sufficient abstractions without caring too much about the generated term.

L-as:
`pheadBuiltin` is much much smaller, since it's only a single term. It's not about "abstraction" per se, it's about how those abstractions are implemented.

Plutarch/Builtin/List.hs
L-as: For deconstructing lists you don't need `ListElemUni`.

TotallyNotChase: I understand that - but do we *really* not need to have a way of constructing builtin lists?

srid:
@TotallyNotChase You are right; `head` should directly use the `HeadList` builtin here. I plan to add benchmarks for tracking script size and cpu/mem cost; I think it is worth doing this from get go, so we can avoid introducing changes that break performance over time.

> Edit: By the way, may I ask why `head` is a Haskell level function? Still trying to wrap my head around the benefits and disadvantages of that :D

The body of the Haskell-level function effectively gets inlined. The function DSL on the other hand becomes a LamAbs that is bound somewhere and then applied at call sites; @L-as - what role does `phoistAcyclic` precisely play here for non-recursive functions?

srid:
I'm closing this, because I decided to work on the DSL types outside of Plutarch repository. I will consider upstreaming the changes latter.

L-as:
`phoistAcyclic` doesn't distinguish between recursive and non-recursive functions, since all functions are non-recursive from an external perspective. Recursion is done through the Y combinator after all.

srid:
> `phoistAcyclic` doesn't distinguish between recursive and non-recursive functions, since all functions are non-recursive from an external perspective. Recursion is done through the Y combinator after all.

Under what scenario (outside of `pfix` itself) would we use `plam` without `phoistAcyclic`?

L-as:
Higher order functions.

Plutarch/Builtin/List.hs
L-as: We can make a fully polymorphic `pmatch` since it's a separate type class now. We can still make a `PCon` instance, but we might still want a separate `pcons` that is fully polymorphic.

Plutarch/Builtin/List.hs
L-as: I've since this found out that the signature in Agda is incorrect. I'll probably make an issue upstream.

PR #25, opened 2021-12-21
Derive PlutusType for PData

srid:
This PR is based on top of #23 

For a review-friendly diff, see https://github.com/srid/plutarch/compare/srid/builtins...srid:srid/builtins-more


srid:
https://github.com/Plutonomicon/plutarch/pull/23#issuecomment-999655446

PR #26, opened 2021-12-22
Use # instead of £ and release 1.0.0

L-as:


L-as:
This was a common complaint from many people. Seemingly, the British keyboard is the outlier in that it has £ right next to $.

PR #27, opened 2021-12-22
Add some tests for UPLC behaviour

L-as:
Notably, it has become evident that lists are in fact
homogeneous, and that since no polymorphic `nil` exists,
it is entirely impossible to use built-in lists of functions.

### DIFF
--- a/examples/Main.hs
+++ b/examples/Main.hs
@@ -1,20 +1,26 @@
+{-# LANGUAGE PartialTypeSignatures #-}
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+{-# OPTIONS_GHC -Wno-partial-type-signatures #-}
 
 module Main (main) where
 
 import Test.Tasty
 import Test.Tasty.HUnit
 
-import Plutarch (ClosedTerm, compile, printScript, printTerm)
+import Plutarch (ClosedTerm, POpaque, compile, printScript, printTerm, punsafeBuiltin, punsafeConstant)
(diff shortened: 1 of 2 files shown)


L-as:
The lines are really long and fourmolu doesn't fix the formatting somehow.

L-as:
https://github.com/tweag/ormolu/issues/695 Haskell formatters are really behind

srid:
Polymorphic nil does work, the last time I checked. See https://github.com/Plutonomicon/plutarch/pull/25/files#diff-338f08d37e64445f86bd861acd5ad294fa7193fb2a9a37ae183ae41be4ba2f34R51-R64

L-as:
Where do you have a polymorphic nil of type `Term s (PBuiltinList a)` for any arbitrary `a`?

srid:
`a` can only be within the range of PLC's [default universe](https://github.com/input-output-hk/plutus/blob/e995df9a339b69523e34bad35816ee1e4ddd9669/plutus-core/plutus-core/src/PlutusCore/Default/Universe.hs#L78-L87) of types of anyway; but I'd consider that to be a given. `Term s (PBuiltinList a)` is not impossible to write at the eDSL, if we accept the aforementioned constraint (which understanding is encoded by the ListElemUni typeclass). 

https://github.com/Plutonomicon/pluton/blob/a2cab16ad591997ce73bbd0e01a09933daa32529/src/Pluton/Types/Builtin/List.hs#L103-L162

L-as:
Hence it isn't useful as we can't put functions in it.

PR #30, opened 2021-12-24
Add functions for working with booleans, and more synonyms to built in functions

TotallyNotChase:
This should finish up with adding almost all the remaining builtin functions as plutarch synonyms. The only ones that are left are about building lists (`MkCons`, `MkNilData`, `MkNilPairData`), building pairs (`MkPairData`), and building `Data` (`ConstrData`, `MapData`, `ListData`, `IData`, `BData`).

* Add `pnot`, `#&&`, and `#||`. Functions for working with `PBool` terms.
* Implement `PEq`, `POrd`, `Semigroup`, and `Monoid` for `PUnit` terms.
* Add the `PIntegral` typeclass alongside its `PInteger` instance. (uses `DivideInteger`, `ModInteger` etc)
* Add synonyms to the cryptography builtins. (see: `Plutarch.Crypto`)
* Add synonyms to builtins that work with `PByteString`, `ConsByteString`, `LengthByteString` etc.
* ~~Add `ppack :: [Word8] -> Term s PByteString`~~
* Add synonyms to `(en|de)codeUTF8` builtins.


Plutarch/Bool.hs
```diff
@@ -35,3 +35,19 @@ pif :: Term s PBool -> Term s a -> Term s a -> Term s a
...
+infixr 3 #&&
+
+(#&&) :: Term s PBool -> Term s PBool -> Term s PBool
+x #&& y = pif x (pif y (pcon PTrue) $ pcon PFalse) $ pcon PFalse
+
+-- | Lazily evaluated boolean or for 'PBool' terms.
+infixr 2 #||
+
+(#||) :: Term s PBool -> Term s PBool -> Term s PBool
+x #|| y = pif x (pcon PTrue) $ pif y (pcon PTrue) $ pcon PFalse
```
L-as, line 53: I wonder whether we could have some way of having Plutarch-level operators.

Plutarch/ByteString.hs
```diff
@@ -37,6 +47,33 @@ phexByteStr = punsafeConstant . PLC.Some . PLC.ValueOf PLC.DefaultUniByteString
...
+plength :: Term s (PByteString :--> PInteger)
+plength = punsafeBuiltin PLC.LengthOfByteString
+
+-- | 'PByteString' indexing function.
+pindex :: Term s (PByteString :--> PInteger :--> PInteger)
+pindex = punsafeBuiltin PLC.IndexByteString
+
+-- | Pack a list of bytes into a 'PByteString' term.
+ppack :: [Word8] -> Term s PByteString
+ppack = punsafeConstant . PLC.Some . PLC.ValueOf PLC.DefaultUniByteString . BS.pack
```
L-as, line 75: `ppack = pbyteStr . BS.pack`, I don't think we need this.

Plutarch/ByteString.hs
```diff
@@ -37,6 +47,33 @@ phexByteStr = punsafeConstant . PLC.Some . PLC.ValueOf PLC.DefaultUniByteString
...
+pslice :: Term s (PInteger :--> PInteger :--> PByteString :--> PByteString)
+pslice = punsafeBuiltin PLC.SliceByteString
+
+-- | Find the length of a 'PByteString'.
+plength :: Term s (PByteString :--> PInteger)
+plength = punsafeBuiltin PLC.LengthOfByteString
+
+-- | 'PByteString' indexing function.
+pindex :: Term s (PByteString :--> PInteger :--> PInteger)
+pindex = punsafeBuiltin PLC.IndexByteString
```
L-as, line 71: I don't think the names should be generic, though I realise that this is how `Data.ByteString` does it.

Plutarch/Integer.hs
```diff
@@ -7,6 +7,18 @@ import qualified PlutusCore as PLC
 
 data PInteger s
 
+class PIntegral a where
+  pdiv :: Term s a -> Term s a -> Term s a
+  pmod :: Term s a -> Term s a -> Term s a
+  pquot :: Term s a -> Term s a -> Term s a
+  prem :: Term s a -> Term s a -> Term s a
```
L-as, line 14: You probably want to make these Plutarch-level functions, since e.g. `Rational` would probably hoist them. It's a trade-off though.

L-as (changes requested):
Looks mostly good to me!

Plutarch/Bool.hs
TotallyNotChase: That's certainly something to think about. But I think the best you can do is have a separate function that "lifts" plutarch level functions to let them work as "infix operators" - though that kinda ruins the idea.

I assume we wouldn't need it for `&&` and `||` though.

Plutarch/ByteString.hs
TotallyNotChase: Hmm, true - actually. Going to remove it.

Plutarch/ByteString.hs
TotallyNotChase: I spent too long debating what I should've done here. In one of the commits in this PR, these were not generic names. I made them generic after the fact thinking "hey that's how haskell containers do it".

I honestly don't know which choice is better here - going to go with whatever you (or others) suggest.

Plutarch/Integer.hs
TotallyNotChase: Another place where I debated what I should've done. But I think it might not matter for just applying builtin functions? Correct me if I am wrong but if these were Plutarch level functions - applying them with `#` would still create an `Apply` term for each application, right?

Plutarch/Bool.hs
TotallyNotChase: I thought a bit more about the efficiency of this. Do you think something like this would be any better?
```hs
(#&&) :: Term s PBool -> Term s PBool -> Term s PBool
x #&& y = _pand # pdelay x # pdelay y

_pand :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
_pand = phoistAcyclic $ plam
  $ \x y -> pif' # pforce x # pcon PTrue #$ pif' # pforce y # pcon PTrue # pcon PFalse
```

L-as: Very likely yes for both. You might want to just make it `pand`, since users may want to use it directly. We might also want a strict variant.

Plutarch/Integer.hs
L-as: It matters if you want to pass the functions to a higher order function. Wonder whether there's any way we could eliminate this headache though.

TotallyNotChase: In the meantime, let's make these Plutarch level functions? I don't quite see a problem with that here.

Are you thinking of eta reductions? A highly simplified implementation perhaps? Just to get rid of cases where we have a term `X`, which is a `Lam` holding just an `Apply`. And then if someone puts the `X` term inside an `Apply` again - perhaps we can just cut the middle man and only yield the innermost `Apply`?

Plutarch/Bool.hs
TotallyNotChase: Done!

Plutarch/Integer.hs
L-as: We essentially have two types of functions: Haskell-level functions and Plutarch-level functions.
Given a Haskell-level function `f`, `plam f` converts it to a Plutarch-level function.
To go the opposite direction, there is no shortcut, but we essentially do `\v1 ... vn -> (plam f) # v1 # ... # vn`.
A round-trip always makes the term bigger.
We can not naively reduce `plam \x -> (plam f) # x` through eta-reduction, as the language has effects (`perror`).
This code compiles down into something like `\x -> f' x`, which if reduced to `f'`, might cause the term to err if `f'` is not in WHNF.

Let's say somebody does
```haskell
f x = f' # x
```
Then somebody wants to pass it to e.g. `pmap :: Term s ((a :--> b) :--> PList a :--> PList b)`, and does the following:
```haskell
pmap # (plam f) # l
```
`plam f` would be compiled into something like the following:
```haskell
\x -> f' x
```
This is often suboptimal. In the case where `f'` is a variable (e.g. a hoisted term), then it's already in WHNF and thus we can eta-reduce it without changing the semantics. However, this becomes *much* more complicated in functions of arity above 1:
```haskell
f' :: Term s (PInteger :--> PInteger :--> PInteger)
f x y = f' # x # y
```
In this case `plam f` would be compiled into:
```haskell
\x -> \y -> f' x y
```
Even if `f'` is in WHNF, the simple algorithm doesn't work anymore as `f' x` isn't in WHNF obviously.

Regardless of this, we still want to be able to do this transformation manually, by *trusting* the user.
The transformation we want to do is essentially `\f -> plam $ \x -> plam f # x` to just `\f -> plam f`.
This transformation is always safe, however, the inner `plam f` could in fact be a reference to a hoisted
term or similar, which is what complicates this.

If we had `plamReduce` which would eta-reduce its body, we could selectively do this.

An example this would fix:
```haskell
x + y = punsafeBuiltin PLC.AddInteger # x # y

padd = plam (+) -- Compiled into `\x -> \y -> AddInteger x y`
padd' = plamReduce (+) -- Compiled into `AddInteger`
```

An example where this would go wrong:
```haskell
_padd :: Term s (PInteger :--> PInteger :--> PInteger)
_padd = perror

x + y = _padd # x # y

padd = plam (+) -- Compiled into `\x -> \y -> error`
padd' = plamReduce (+) -- Compiled into `error`, bad
```

This is potentially surprising behaviour.

Perhaps the solution is just to perform deep analysis of terms.

We could special-case hoisted terms by adding another guarantee:
- Hoisted terms must not inspect their arguments before all arguments
  are passed.

This would make the following illegal:
```haskell
padd :: Term s (PInteger :--> PInteger :--> PInteger)
padd = phoistAcyclic $ plam $ \_ -> perror
```

It would be illegal because then the following would change semantics:
```haskell
padd' = plam \x y -> padd # x # y
```

The reason being that partially applying the function will now cause
an error.

Do note that the following is already "illegal":
```haskell
padd :: Term s (PInteger :--> PInteger :--> PInteger)
padd = phoistAcyclic $ perror
```

In this case, however, it's much less confusing since the script will *always*
err before doing anything, no matter what arguments you pass it.

In addition, the following would also complicate it:
```haskell
pid :: Term s (a :--> a)
pid = phoistAcyclic $ plam $ \x -> x

f :: Term s ((a :--> b :--> c) :--> a :--> b :--> c)
f = plam $ \x y z -> pid # x # y # z

g :: Term s (PInteger :--> PInteger :--> PInteger)
g = plam $ \_ -> perror

-- This will err if `f` is eta-reduced.
h = f # g # 1
```

The above is true even if we have the restriction on hoisted terms.

In the end I think the best possible solution is to do some analysis after the fact,
where we keep track of arities of functions and built-ins.

`plamReduce` is the other possible approach, and is a stop-gap, as the above issue
would be avoided unless users explicitly opt into it, but it's very hard to know
when it's safe to use on a function `f`, as you need to know how `f` is implemented.
In addition I'm not sure how much easier it would be to implement.

L-as: For the time being we probably want to recommend Plutarch-level functions over Haskell-level functions in general.

L-as: Plutarch-level functions can also have inefficiencies, but they can be fixed by inlining. E.g. if we have a function that does `\x -> x + x`, then applying it to a variable would do `(\x -> x + x) y`, but if `y` is already small enough then we should just in-line it. We only get this opportunity though if the function *isn't* hoisted.

TotallyNotChase: Made the `PIntegral` functions Plutarch level. Also, you should probably put your comments here onto an issue or similar where its more accessible to the general public.

Plutarch/ByteString.hs
L-as: The issue IMO is that we haven't done this until now. E.g. `pheadBuiltin`. I think we should do the same here and write `pbindex` or something to make it fit with the existing naming convention.

L-as: You know what, it doesn't matter.

TotallyNotChase: How about these names-
https://github.com/Plutonomicon/plutarch/pull/30/commits/da57217b6a12fbe0f6ee87a2c4950169f6669416

Should these go back to those names?

L-as: I think we should still choose a name consistent with everything else though. E.g. `pindexBS`, to make it similar to `pheadBuiltin` and such.

TotallyNotChase: Sure, done!

L-as (approved):
LGTM

PR #36, opened 2021-12-27
Deconstruct `pdata` arg before passing to `ConstrData`

TotallyNotChase:


Plutarch/Builtin.hs
```diff
@@ -97,4 +97,4 @@ instance PIsData PByteString where
 
 instance PIsData (PBuiltinPair PInteger (PBuiltinList PData)) where
   pfromData x = pasConstr # pforgetData x
-  pdata x = punsafeBuiltin PLC.ConstrData # x
+  pdata x = punsafeBuiltin PLC.ConstrData # (pfstBuiltin # x) #$ psndBuiltin # x
```
L-as, line 100: `plet`

TotallyNotChase: Ah, of course - completely forgot `pdata` is a Haskell level function! Done.

L-as: This is another good reason to recommend Plutarch-level functions: Less accidental duplication.

PR #39, opened 2021-12-28
Add tracing utilities; with cabal flags switching

TotallyNotChase:
It's added as a separate package - `plutarch-trace`. Everything one needs to know to use `plutarch-trace` and switch between tracing/no-tracing is explained in `plutarch-trace/README`. Users can also have both tracing and no-tracing version at the same time.

Cabal >= 3.0 will be required to add `plutarch-trace` as a dependency.

Some general notes-
* Any testing of these new utilities should be done within `plutarch-tracing` itself, after instantiating a signature. This is because `plutarch-tracing` depends on `plutarch`.
* `plutarch-tracing` exposes a *generic module*. That is, a module that actually uses a signature under the hood. In fact, the module just re-exports the signature.

  Why not just have the signature? You cannot instantiate a signature directly and get to using it. You need to have a package with a module that uses said signature, then you need to instantiate that package using an implementation. You cannot do-
  ```
  plutarch-trace (Plutarch.TraceSig) requires (Plutarch.TraceSig as Plutarch.Trace.Enable),
  ```
  where `Plutarch.TraceSig` is a signature.

  TL;DR - In our case, dev experience will be better with this structure. No one should ever have to import `Plutarch.TraceSig`.
* Cabal will give off warnings about using experimental features and some things are clearly a bit finicky. But it's all good, this usecase is not too complex.


L-as:
Question: Is it necessary to have a separate cabal project `plutarch-trace`?

TotallyNotChase:
Yes. So the thing about mixins is that you have to explicitly name every single module you want to import in `mixins`. If it was part of `Plutarch`, for example, you can't do-
```cabal
plutarch (Plutarch.Trace) requires (Plutarch.TraceSig as Plutarch.Trace.Enable),
```
It would only make `Plutarch.Trace` visible. You need to list all the modules in there.

I think this is why signature packages are generally tiny and don't export many modules.

L-as:
Makes sense. As for the tests, we this should be fine. We should be able to depend on plutarch-trace in the tests.

TotallyNotChase:
I wonder if `plutarch requires (Plutarch.TraceSig as Plutarch.Trace.Enable)` works. I was under the impression it doesn't from `really-small-backpack-example`.

Also, I think you'll get a cyclic dependency error if you try adding `plutarch-trace` to the current test suite. Pretty sure I tried that. Might need to make the test suite in another package all together.

plutarch-trace/impl/Plutarch/Trace/Disable.hs
```diff
@@ -0,0 +1,24 @@
+module Plutarch.Trace.Disable (ptrace', ptrace, ptraceIfTrue, ptraceIfFalse) where
+
+import Plutarch.Bool (PBool)
+import Plutarch.Prelude
+import Plutarch.String (PString)
+
+pf :: Term s (b :--> a :--> a)
+pf = phoistAcyclic $ plam $ \_ y -> y
+
+-- | A strict version of 'ptrace'.
```
L-as, line 10: I don't think this documentation is needed.

plutarch-trace/src/Plutarch/TraceSig.hsig
```diff
@@ -0,0 +1,18 @@
...
+import Plutarch.Bool (PBool)
+
+-- | A strict version of 'ptrace'.
+ptrace' :: Term s (PString :--> a :--> a)
+
+-- | Trace a message, then evaluate and return given argument.
+ptrace :: Term s PString -> Term s a -> Term s a
+
+-- | Trace a message if given argument evaluates to true.
+ptraceIfTrue :: Term s (PString :--> PBool :--> PBool)
```
L-as, line 15: Not sure whether the signatures were originally like this, but we probably want to make this `Term s PString -> Term s PBool -> Term s PBool`, otherwise we can't elide the string after all.

plutarch-trace/README.md
```diff
@@ -0,0 +1,149 @@
...
+    base,
+    plutarch-trace,
+    plutarch-trace:enable
+  mixins:
+    plutarch-trace (Plutarch.Trace) requires (Plutarch.TraceSig as Plutarch.Trace.Enable)
+  default-language: Haskell2010
+```
+
+### Disable Tracing
+* Add `plutarch-trace:disable` to your `build-depends`.
```
L-as, line 44: You might want to write a few tips about how to make the choice available without changing the source code; I'm specifically thinking that end-users can add a flag to their `.cabal`, then have their executable `if` that flag to choose between enabling and disabling tracing. There's also the option of duplicating the `executable` section such that one has `foo-exe` and `foo-exe-debug`.

L-as (changes requested):
Please ~~remove the tracing utilities `plutarch` and~~ add some tests.

TotallyNotChase:
> Please remove the tracing utilities `plutarch` and add some tests.

Sorry, didn't get what you mean here. Remove what?

L-as:
https://github.com/haskell/cabal/pull/3422

We probably want to add a separate package for tests.

L-as:
> > Please remove the tracing utilities `plutarch` and add some tests.
> 
> Sorry, didn't get what you mean here. Remove what?

This is entirely my mistake. I was thinking of the code you wrote outside of this repository for tracing and had forgotten that that wasn't in Plutarch yet.

L-as:
Relevant: https://github.com/input-output-hk/plutus/pull/4304

L-as:
I'm honestly still not sure whether Backpack is the best choice, but since I have very little experience with it, I'm thinking it's worth trying. Worst case we rip it out and make a backward incompatible change.

TotallyNotChase:
I'm not super sure about backpack either ¯\_(ツ)_/¯

But from what I've seen so far - I don't think we'll need to rip it out. I've learnt from Drazen that cabal flags aren't exactly the best for even slightly complex dev UX. I like the ability to have both versions at once. I like the easy switching. I like the painless propagation of "generic-ness". And for our simple usecase, this *shouldn't* cause any problems. We won't really be mending much with the tracing package anyway.

TotallyNotChase:
By the way, it looks like `plutarch requires (Plutarch.TraceSig as Plutarch.Trace.Enable)` works for importing all modules! You can't use the renaming feature at all if you do that implicit import all though. So we still definitely want `plutarch-tracing` as a separate package.

plutarch-trace/src/Plutarch/TraceSig.hsig
TotallyNotChase: Do you think that may be a trade-off? (wondering about preferring Plutarch level functions - though I do get the advantage of elision)

plutarch-trace/impl/Plutarch/Trace/Disable.hs
TotallyNotChase: Do you think the traces should be removed from the implementation and only kept in the signature? I actually have no idea how hoogle works with signatures (that is, if we ever decide to generate hoogle docs for this).

Fresheyeball:
@L-as @TotallyNotChase I would be surprised if backpack gets ongoing support in GHC. It's very unpopular, and the economics are terrible. It's not really a first module system for Haskell a la Standard ML, so much as a hacked on module system. I would avoid backpack if possible. I have some alternative approaches if you want.

TotallyNotChase:
Would love to hear about alternative approaches! Not a fan of cabal flags + CPP either.

L-as:
I had some discussions on #haskell:libera.chat wrt. this. Seems like nobody is maintaining Backpack, not to mention it's still unsupported in Stack. The implementation in GHC is supposedly not even finished. I'm not sure what the correct decision to take here is. Considering it's a small part of Plutarch, we can always switch it out if needed.

L-as:
https://gitlab.haskell.org/ghc/ghc/-/wikis/Backpack-refactoring

TotallyNotChase:
I've noticed Matthew Pickering has started to make an effort at improving backpack 2 months ago. That's pretty neat!

I find it funny that the backpack *unofficial documentation*, through the examples, is *extremely* good. Getting up and running with backpack is super easy with `really-small-backpack-example`. But the official documentation is non-existent apart from [cabal](https://cabal.readthedocs.io/en/latest/cabal-package.html#backpack).

Also, I'm totally fine with switching the backpack impl out for some other machinery. I had fun learning it anyway. Although I do personally like backpack, we don't have certainty of its future.

L-as:
I think we should go with the Backpack approach for now. Doesn't make sense timewise to rewrite it now *in case* support gets dropped in the future. We can always handle this then.

plutarch-trace/README.md
TotallyNotChase: Done! You may want to review [7592706](https://github.com/Plutonomicon/plutarch/pull/39/commits/7592706c219c7115a9f25b2e89f58acda15f4b49) separately. I also added some details on hiding backpack impl details in case anyone needs that. The cabal user guide on backpack was surprisingly useful for this.

L-as: LGTM.

TotallyNotChase:
@L-as Moved out the tests suite to a separate package and added tests for tracing/no-tracing. I'm not sure if the CI ran them though.

plutarch-trace/impl/Plutarch/Trace/Disable.hs
L-as: The documentation should probably only be on the signatures.

plutarch-trace/src/Plutarch/TraceSig.hsig
L-as: It is a trade-off, but in this case it's best if we do it this way.

L-as: Do note that you can still hoist it even if it's a Haskell level function by making a helper Plutarch-level function which you hoist.

TotallyNotChase: Done!

plutarch-trace/impl/Plutarch/Trace/Disable.hs
TotallyNotChase: Done!

plutarch-trace/impl/Plutarch/Trace/Disable.hs
```diff
@@ -0,0 +1,17 @@
+module Plutarch.Trace.Disable (ptrace', ptrace, ptraceIfTrue, ptraceIfFalse) where
+
+import Plutarch.Bool (PBool)
+import Plutarch.Prelude
+import Plutarch.String (PString)
+
+ptrace' :: Term s (PString :--> a :--> a)
+ptrace' = phoistAcyclic $ plam $ \_ y -> y
```
L-as, line 8: Dead code.

plutarch-trace/src/Plutarch/TraceSig.hsig
TotallyNotChase: Ok, done for real. I do wonder though, should the non tracing versions still evaluate their argument (not the string one)?

plutarch-trace/impl/Plutarch/Trace/Disable.hs
TotallyNotChase: Do you mean `ptrace'`? Or something else?

plutarch-trace/src/Plutarch/TraceSig.hsig
```diff
@@ -0,0 +1,18 @@
+signature Plutarch.TraceSig (ptrace', ptrace, ptraceIfTrue, ptraceIfFalse) where
+
+import Prelude ()
+import Plutarch.Prelude (Term, (:-->))
+import Plutarch.String (PString)
+import Plutarch.Bool (PBool)
+
+-- | A strict version of 'ptrace'.
+ptrace' :: Term s (PString :--> a :--> a)
```
L-as, line 9: I just realised that you also have `ptrace'` here. When would you actually use this?

TotallyNotChase: Whenever anyone wants a strict version, and *possibly* slightly more efficient version - maybe. Similar to `pif'`, `pand'`, and the like. `ptrace` delays its second argument (Haskell level argument) before application, and then finally forces before returning. This was to ensure that the log within `ptrace` happens *before* the potential logs within its second argument.

If `ptrace'` has no chance of being more efficient in a case like `ptrace' # "foo" # ()` - where delaying the second argument is useless - it makes sense to remove it. But I'm not sure if that's the case.

L-as: No, it *is* more efficient in some cases, but we don't care about efficiency in the first place when traces are enabled, no? Traces are going to be completely removed in production after all.

plutarch-trace/src/Plutarch/TraceSig.hsig
L-as: They should just pass it through, which is the same as applying it to `id`.

plutarch-trace/src/Plutarch/TraceSig.hsig
L-as: https://github.com/Plutonomicon/plutarch/pull/39/commits/98b8e5f47f5bea7ac9af8513ee76734461a51742

L-as:
There's a big problem here... haskell.nix doesn't support Backpack.

https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fsw3526sw0pv4w8m5dp4mydfs8zccw1vd-examples-test-examples-1.0.0.drv/log?via-job=b0aa0a14-305b-4e3b-bca1-aacd858db38c

https://github.com/input-output-hk/haskell.nix/issues/244

L-as:
It seems like we'll have to move off of Backpack...

The only other clean solution I can think of is passing in `ptrace` using implicit arguments / type classes, but that would mean every function now needs to do `Something => Term s ...`, which would be a huge breaking change.

The alternative is that we fix support for Backpack but from what I can tell, this is far from trivial.
It's not even support by the Haskell infrastructure in Nixpkgs.

We'll probably have to use flags and CPP unfortunately.

blamario:
I suppose I'm missing something, but why do we need to elevate this to the level of Cabal? Why wouldn't something as simple as https://hackage.haskell.org/package/NoTrace work well enough?

L-as:
Is there a good programmatic way of switching automatically to that? I certainly don't think we *need* Backpack, but just adding add `Plutarch.NoTrace` doesn't solve it.

TotallyNotChase:
@L-as Ha, I expected something of this sort to happen. Let's just get this over with - switched to cabal flags + CPP.

I think this also means the CI should re-run the "Tracing" testcase with a `-f development` now. `development` flag is off by default.

L-as:
@TotallyNotChase We don't need to use CPP. We can use the flags in the cabal file to select between different source files.

TotallyNotChase:
> @TotallyNotChase We don't need to use CPP. We can use the flags in the cabal file to select between different source files. Although I'm not sure whether this is a better approach.

CPP reduces some code duplication compared to that approach. I don't think it really matters though.

L-as:
Is this missing anything? I looked at the code and it seemed fine to me.

TotallyNotChase:
@L-as It should be ready to merge.

L-as (approved):
LGTM

PR #41, opened 2021-12-28
Remove unsound "delay (force x) => x" optimisation

L-as:
Thanks @TotallyNotChase for noticing the issue in the generated code!

### DIFF
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -71,8 +71,8 @@ mapTerm f (TermResult t d) = TermResult (f t) d
 mergeTerms :: (RawTerm -> RawTerm -> RawTerm) -> TermResult -> TermResult -> TermResult
 mergeTerms f x y = TermResult (f (getTerm x) (getTerm y)) (getDeps x <> getDeps y)
 
-mkNew :: RawTerm -> TermResult
-mkNew r = TermResult r []
+mkTermRes :: RawTerm -> TermResult
+mkTermRes r = TermResult r []
 
 -- Source: Unembedding Domain-Specific Languages by Robert Atkey, Sam Lindley, Jeremy Yallop
 -- Thanks!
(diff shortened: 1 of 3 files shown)


L-as:
I'm still debating internally between view patterns and record syntax for pattern matching:
```haskell
TermResult { getTerm = RError }
```
vs
```haskell
getTerm -> RError
```

Plutarch/Internal.hs
```diff
@@ -60,6 +60,20 @@ hashRawTerm' (RHoisted (HoistedTerm hash _)) = flip hashUpdate ("8" :: BS.ByteSt
...
+  }
+
+mapTerm :: (RawTerm -> RawTerm) -> TermResult -> TermResult
+mapTerm f (TermResult t d) = TermResult (f t) d
+
+mergeTerms :: (RawTerm -> RawTerm -> RawTerm) -> TermResult -> TermResult -> TermResult
+mergeTerms f x y = TermResult (f (getTerm x) (getTerm y)) (getDeps x <> getDeps y)
+
+mkNew :: RawTerm -> TermResult
+mkNew r = TermResult r []
```
TotallyNotChase, line 75: Name this `mkTermRes` maybe? I understand that this isn't exported so it hardly matters but still.

Plutarch/Internal.hs
```diff
@@ -82,64 +96,61 @@ data PDelayed (a :: k -> Type) (s :: k)
 
 plam' :: (Term s a -> Term s b) -> Term s (a :--> b)
 plam' f = Term $ \i ->
-  let v = Term $ \j -> (RVar (j - (i + 1)), [])
-      (t, deps) = asRawTerm (f v) (i + 1)
-   in (RLamAbs $ t, deps)
+  let v = Term $ \j -> TermResult (RVar (j - (i + 1))) []
```
TotallyNotChase, line 99:
```hs
TermResult (RVar (j - (i + 1))) []
```
Is there a reason for not using `mkNew` here?

TotallyNotChase (approved):

TotallyNotChase:
I tested this on the validator rewrite and it works correctly! Looks like all the optimizations combined also shaved off some script size and cpu/mem units!

Plutarch/Internal.hs
L-as: Nice catch.

Plutarch/Internal.hs
L-as: Thanks.

TotallyNotChase (approved):

TotallyNotChase:
LGTM! Let's get this onto production!

PR #42, opened 2021-12-29
optimisation: Do some η-reductions

### LINKED ISSUE #32: Fix unnecessary η-expansions
We essentially have two types of functions: Haskell-level functions and Plutarch-level functions.
Given a Haskell-level function `f`, `plam f` converts it to a Plutarch-level function.
To go the opposite direction, there is no shortcut, but we essentially do `\v1 ... vn -> (plam f) # v1 # ... # vn`.
A round-trip always makes the term bigger.
We can not naively reduce `plam \x -> (plam f) # x` through eta-reduction, as the language has effects (`perror`).
This code compiles down into something like `\x -> f' x`, which if reduced to `f'`, might cause the term to err if `f'` is not in WHNF.

Let's say somebody does
```haskell
f x = f' # x
```
Then somebody wants to pass it to e.g. `pmap :: Term s ((a :--> b) :--> PList a :--> PList b)`, and does the following:
```haskell
pmap # (plam f) # l
```
`plam f` would be compiled into something like the following:
```haskell
\x -> f' x
```
This is often suboptimal. In the case where `f'` is a variable (e.g. a hoisted term), then it's already in WHNF and thus we can eta-reduce it without changing the semantics. However, this becomes *much* more complicated in functions of arity above 1:
```haskell
f' :: Term s (PInteger :--> PInteger :--> PInteger)
f x y = f' # x # y
```
In this case `plam f` would be compiled into:
```haskell
\x -> \y -> f' x y
```
Even if `f'` is in WHNF, the simple algorithm doesn't work anymore as `f' x` isn't in WHNF obviously.

Regardless of this, we still want to be able to do this transformation manually, by *trusting* the user.
The transformation we want to do is essentially `\f -> plam $ \x -> plam f # x` to just `\f -> plam f`.
This transformation is always safe, however, the inner `plam f` could in fact be a reference to a hoisted
term or similar, which is what complicates this.

If we had `plamReduce` which would eta-reduce its body, we could selectively do this.

An example this would fix:
```haskell
x + y = punsafeBuiltin PLC.AddInteger # x # y

padd = plam (+) -- Compiled into `\x -> \y -> AddInteger x y`
padd' = plamReduce (+) -- Compiled into `AddInteger`
```

An example where this would go wrong:
```haskell
_padd :: Term s (PInteger :--> PInteger :--> PInteger)
_padd = perror

x + y = _padd # x # y

padd = plam (+) -- Compiled into `\x -> \y -> error`
padd' = plamReduce (+) -- Compiled into `error`, bad
```

This is potentially surprising behaviour.

Perhaps the solution is just to perform deep analysis of terms.

We could special-case hoisted terms by adding another guarantee:
- Hoisted terms must not inspect their arguments before all arguments
  are passed.

This would make the following illegal:
```haskell
padd :: Term s (PInteger :--> PInteger :--> PInteger)
padd = phoistAcyclic $ plam $ \_ -> perror
```

It would be illegal because then the following would change semantics:
```haskell
padd' = plam \x y -> padd # x # y
```

The reason being that partially applying the function will now cause
an error.

Do note that the following is already "illegal":
```haskell
padd :: Term s (PInteger :--> PInteger :--> PInteger)
padd = phoistAcyclic $ perror
```

In this case, however, it's much less confusing since the script will *always*
err before doing anything, no matter what arguments you pass it.

In addition, the following would also complicate it:
```haskell
pid :: Term s (a :--> a)
pid = phoistAcyclic $ plam $ \x -> x

f :: Term s ((a :--> b :--> c) :--> a :--> b :--> c)
f = plam $ \x y z -> pid # x # y # z

g :: Term s (PInteger :--> PInteger :--> PInteger)
g = plam $ \_ -> perror

-- This will err if `f` is eta-reduced.
h = f # g # 1
```

The above is true even if we have the restriction on hoisted terms.

In the end I think the best possible solution is to do some analysis after the fact,
where we keep track of arities of functions and built-ins.

`plamReduce` is the other possible approach, and is a stop-gap, as the above issue
would be avoided unless users explicitly opt into it, but it's very hard to know
when it's safe to use on a function `f`, as you need to know how `f` is implemented.
In addition I'm not sure how much easier it would be to implement.

(from https://github.com/Plutonomicon/plutarch/pull/30#discussion_r775141470)

L-as:
We take advantage of the previously introduced concept of "arities"
to determine when this optimisation is sound. Arity represents
how many arguments can be passed before the term is no longer in WHNF.

We can't η-reduce when the wrapping lambda has a higher arity, as
that would mean e.g. `\x y z -> f x y z` where `f = \_ -> perror` would be
optimised into `\_ -> perror` which has different behaviour.

We don't catch all cases, specifically, the `f` in `\x -> f x` must be
a hoisted term (therefore closed). It doesn't have to be this restricted,
but then we'd need per-level tracking of variable usage by terms, which is
certainly possible (and also useful for automatic hoisting / CSE) but not
something I'm going to do in this commit.

Fixes #32

### DIFF
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -95,8 +95,77 @@ plam' :: (Term s a -> Term s b) -> Term s (a :--> b)
 plam' f = Term $ \i ->
   let v = Term $ \j -> mkTermRes $ RVar (j - (i + 1))
    in case asRawTerm (f v) (i + 1) of
+        -- arity 1
+        t@(getTerm -> RApply t'@(getArity -> Just _) [RVar 0]) -> t {getTerm = t'}
+        -- arity 2 + n
+        t@(getTerm -> RLamAbs n (RApply t'@(getArity -> Just n') args))
+          | (maybe False (== [0 .. n + 1]) $ traverse (\case RVar n -> Just n; _ -> Nothing) args)
+              && n' >= n + 1 ->
+            t {getTerm = t'}
         t@(getTerm -> RLamAbs n t') -> t {getTerm = RLamAbs (n + 1) t'}
         t -> mapTerm (RLamAbs 0) t
+  where
+    -- 0 is 1
+    getArity :: RawTerm -> Maybe Natural
+    -- We only do this if it's hoisted, since it's only safe if it doesn't
+    -- refer to any of the variables in the wrapping lambda.
+    getArity (RHoisted (HoistedTerm _ (RLamAbs n _))) = Just n
+    getArity (RHoisted (HoistedTerm _ t)) = getArityBuiltin t
+    getArity t = getArityBuiltin t
+
+    getArityBuiltin :: RawTerm -> Maybe Natural
+    getArityBuiltin (RBuiltin PLC.AddInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.SubtractInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.MultiplyInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.DivideInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.QuotientInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.RemainderInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.ModInteger) = Just 1
+    getArityBuiltin (RBuiltin PLC.EqualsInteger) = Just 1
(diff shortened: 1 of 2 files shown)


L-as:
Note: This will mean that the following holds:
```haskell
f = phoistAcyclic $ plam $ \x y -> _

g x y = f # x # y

f' = plam $ \x y -> g x y

f' ≡ f
```

L-as:
Since @TotallyNotChase is off, I'll merge this, as there is no one else to test this new version with real code AFAIK.

PR #47, opened 2021-12-30
add shrinker integration

Geometer1729:
I added a function compileAndShrink with the same type as compile which just runs the result of compile through shrinkScript.

### DIFF
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -23,21 +23,25 @@ module Plutarch.Internal (
   TermCont (..),
 ) where
 
+import Control.Monad (guard)
 import Crypto.Hash (Context, Digest, hashFinalize, hashInit, hashUpdate)
 import Crypto.Hash.Algorithms (Blake2b_160)
 import Crypto.Hash.IO (HashAlgorithm)
 import qualified Data.ByteString as BS
+import Data.Function ((&))
 import Data.Kind (Type)
 import Data.List (foldl', groupBy, sortOn)
 import qualified Data.Map.Lazy as M
 import qualified Data.Set as S
 import qualified Flat.Run as F
+import GHC.Natural (intToNatural)
 import GHC.Stack (HasCallStack)
 import Numeric.Natural (Natural)
 import Plutarch.Evaluate (evaluateScript)
 import Plutus.V1.Ledger.Scripts (Script (Script))
 import PlutusCore (Some, ValueOf)
 import qualified PlutusCore as PLC
+import qualified PlutusCore.Constant as PLC
 import PlutusCore.DeBruijn (DeBruijn (DeBruijn), Index (Index))
 import qualified UntypedPlutusCore as UPLC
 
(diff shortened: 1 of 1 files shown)


L-as:
Thanks! I am honestly not sure whether exposing this in `Plutarch.Internal` is worth it because:
1) Users of this library will still have to do the source-repository-package-sha256map-dance.
2) It saves users of the Shrinker library probably only ~8 characters in total. (`import Shrink (shrinkScript)` vs `import Plutarch (compileAndShrink)`, `shrinkScript $ compile t` vs `compileAndShrink t`)

However, what I think could be interesting is adding a dependency on `Shrinker` for the tests. Have you tried compiling the examples with the shrinker enabled? We want to ensure that it works with shrinking enabled after all.

Geometer1729:
Fair enough, I'll work on some tests with the plutarch examples.

Geometer1729:
I added some analogous unit tests to the existing ones which use shrinker.

examples/Main.hs
```diff
@@ -289,3 +320,118 @@ dummyCurrency :: CurrencySymbol
 dummyCurrency =
   CurrencySymbol . fromJust . Aeson.decode $
     "\"1111111111111111111111111111111111111111111111111111111111111111\""
+
+shrinkerTests :: TestTree
+shrinkerTests =
+  testGroup
+    "shrinker tests"
```
L-as, line 327: Preferably we want to avoid duplicating each test case. We probably need to restructure the examples. Do you have any opinions wrt. how to do this?

Geometer1729: Yeah I didn't feel great about that while writing it but couldn't think of a better way at the time. A data structure representing a test which could then be interpreted differently by the two sets of unit tests seems like a good way to go now that I think about it some more. If that sounds good to you I'll start working on that once I finish a bug fix in shrinker.

L-as: That sounds fine to me! Another solution is making the `fails` etc. functions be members of a type class, such that you can instantiate them differently.

srid: A simple parametrization would work, I think?

```haskell
overallTests :: TestTree 
overallTests = 
  testGroup "tests" $
    flip fmap [CompileOpts True, CompileOpts False] $ plutarchTests

plutarchTests :: CompileOpts -> TestTree
plutarchTests opts =
  testGroup
    ("plutarch tests: shrink=" <> show (shrink opts))
       [ test "foo" $ do 
            compile opts (constant @PInteger 42) @?= 42
        ...
```

Instead of `compileAndShrink` - we could have `compile :: CompileOpts -> ...`, with:

```haskell
data CompileOpts = CompileOpts = { shrink :: Bool }
```

Tests that use `printTerm` can be separated out (since they are hardcoded to compare the UPLC). We can then write a `compile` that implicitly compares unshrunk and shrunk evaluation if shrink=True. With those two things in place, we wouldn't need free monads or type classes - and the test code would remain plain to use over time.

Geometer1729: That's a little tricky. For the PrintsTo tests I don't just compile it with shrink and make sure it prints to the target string (because that would only work when shrink doesn't change anything) I just check that the output with and without shrink are the same (up to a loose notion of equivalence from shrinker-testing). Also with Equal I also add some extra equality tests between evaluations with and without shrink. 

I think all the other tests could be made more generic along those lines by replacing them all with `Test :: (Reader CompileOpts Assertion) -> TestStruct` but I think I'd still need the free monad for PrintsTo and Equal.

L-as:
I somehow  closed this by accident.

L-as:
I can not reopen this. I will make a new PR and merge the changes @Geometer1729

PR #49, opened 2022-01-03
Haddock docs, minor `examples` reorganisation

t1lde:
Added some haddock docs, and some (small) relevant examples.

WIP - some more things could use doc comments, and the `Examples` modules may not be the best place to keep the examples (would be nice to better expose them to haddock...)


L-as:
Why is this a draft?

examples/Examples/PlutusType.hs
```diff
@@ -0,0 +1,52 @@
...
+  A Sum type, which can be encoded as an Enum
+-}
+data AB (s :: k) = A | B
+  
+{- | 
+  AB is encoded as an Enum, using values of PInteger
+  internally.
+-}
+instance PlutusType AB where
+  type PInner AB _ = PInteger
```
L-as, line 22: This is quite clever.

t1lde:
Put this up as draft before I logged off last night, but it should be ready ig.

Will fix formatting.

t1lde:
Looks like there's a fourmolu idempotentcy bug, something to do with Haddock that appears to be fixed in fourmolu 0.4.0.0, which hasn't made it's way into nixpkgs yet :upside_down_face:

L-as:
We need to use `fourmolu_0_4_0_0` instead of `fourmolu` in our Flake.

L-as:
See 8e815b11eed5ac324e570a88647ae3af574d5c79 from tmp branch. You can `git reset` to this.

L-as:
Nevermind, CI still fails...

L-as:
See 47775379dacb21944043bc3ea044cb18cf1bca3d instead, apparently `fourmolu_0_4_0_0` is broken on NIxpkgs. We could include it using `haskell.nix`, but :shrug: for now.

Plutarch.hs
```diff
@@ -106,30 +173,64 @@ instance PlutusType a => PCon a where
   pcon = punsafeCoerce . pcon'
 
 class PCon a where
+  -- | Construct a Plutarch Term via a Haskell value
   pcon :: a s -> Term s a
 
 class PMatch a where
+  -- | Construct a Plutarch Term via a Haskell value
```
TotallyNotChase, line 180: I think this one needs to be updated.

TotallyNotChase (changes requested):
Looks great! The haddocks and examples are going to very useful for #55 as well.

Plutarch.hs
t1lde: Done

L-as:
I just noticed you made this merge into `staging`, but `staging` is just my personal playground currently... this is also on me lol.

PR #50, opened 2022-01-04
Add `ppairDataBuiltin`; Make `Tracing` tests structure consistent with other tests

TotallyNotChase:


Plutarch/Builtin.hs
```diff
@@ -28,8 +32,24 @@ import PlutusTx (Data)
...
+pfstBuiltin = phoistAcyclic $ pforce . pforce . punsafeBuiltin $ PLC.FstPair
+
+psndBuiltin :: Term s (PBuiltinPair a b :--> b)
+psndBuiltin = phoistAcyclic $ pforce . pforce . punsafeBuiltin $ PLC.SndPair
+
+{- | Construct a builtin pair of 'PData' elements.
+
+Uses 'PAsData' to preserve more information about the underlying 'PData'.
+-}
+ppairDataBuiltin :: Term s (PAsData a :--> PAsData b :--> PBuiltinPair (PAsData a) (PAsData b))
```
L-as, line 45: We probably want to fix the `IsData` instance for pairs then to make it use `PAsData`.

Plutarch/Builtin.hs
```diff
@@ -39,6 +59,20 @@ ptailBuiltin = phoistAcyclic $ pforce $ punsafeBuiltin PLC.TailList
...
+-}
+pnilDataBuiltin :: Term s (PBuiltinList (PAsData a))
+pnilDataBuiltin = punsafeConstant . PLC.Some $ PLC.ValueOf (PLC.DefaultUniList PLC.DefaultUniData) []
+
+{- | A nil for builtin list of builtin pairs of 'PData' elements.
+
+This uses 'PAsData' to preserve more information about the underlying 'PData'.
+-}
+pnilPairDataBuiltin :: Term s (PBuiltinList (PBuiltinPair (PAsData a) (PAsData b)))
+pnilPairDataBuiltin = punsafeConstant . PLC.Some $ PLC.ValueOf (PLC.DefaultUniList $ PLC.DefaultUniPair PLC.DefaultUniData PLC.DefaultUniData) []
```
L-as, line 74: While it's easy to add these functions, I don't think we should unfortunately, since we will have to deprecate them very soon. Notably, we want a generic `pconstant` that works for all combinations of literals as I've mentioned before. I can look into designing this if you're having a hard time imagining what it should look llike.

L-as (changes requested):
See comment. Thanks for the contribution!

Plutarch/Builtin.hs
TotallyNotChase: Not sure what you meant here. The `IsData` instance is just for `(Integer, [Data])` right? The 2 members aren't `Data` values so we can't use `PAsData` there.

Plutarch/Builtin.hs
TotallyNotChase: Definitely want `pconstant` for all this stuff. Have that on my todo list a well. These here are really just meant to be alternatives that people can use in the meantime. As I have been myself in the validator rewrite.

Plutarch/Builtin.hs
L-as: You are right.

L-as: However, why do you need this then?

Plutarch/Builtin.hs
L-as: What is this function used for?

TotallyNotChase: I often find myself needing a `nil`. On the validator rewrites, I mostly use builtin list of `Data` - where I needed a `pnilDataBuiltin` to implement `filterData :: Term s (PBuiltinList a :--> PBuiltinList a)`.

I haven't needed `pnilPairDataBuiltin` it - but we'll need it when working with plutus assoc maps. The `Map` data value, `[(Data, Data)]`. This will end up representing things like `Value` - and implementing utilities on that `[(Data, Data)]` may require a `pnilPairDataBuiltin`.

L-as: I am currently leaning toward exploring #53 a bit more before deciding on this.

TotallyNotChase: It'd be great to have #53 resolved so soon! No need to have specific `nil`s in that case. I'll remove these from this PR.

srid: With #62 you get write this functions as:

```haskell
pnil :: forall k a (s :: k). PLC.Contains PLC.DefaultUni (PDefaultUniType a) => Term s (PBuiltinList a)
pnil = pconstant @(PBuiltinList a) []

pnilPairData :: forall k (s :: k). (Data, Data) -> Term s (PBuiltinPair PData PData)
pnilPairData = pconstant @(PBuiltinPair PData PData)
```

L-as: Why do you need the type applications?

L-as:
We have `pconstant` now.

TotallyNotChase:
> We have `pconstant` now.

Awesome! Removed the `nil` stuff. Also removed the `pconsBuiltin` in favor of having that land in #63 

I guess this PR just adds `ppairDataBuiltin` now lol. Also, I moved out `Plutarch.Spec.Tracing` and put it inside `Examples.Tracing` for consistency.

L-as (approved):
LGTM

PR #60, opened 2022-01-04
Draft letrec implementation

blamario:


L-as:
#59

L-as (commented):
In general I like the idea, however, I'm wondering whether we can make this more ergonomic.

What we essentially want is to `pfix` a dependently typed function, where the first argument is `Fin n` for `n` different "bindings". The return type then depends on this argument.

blamario:
I hope you'll find it more ergonomic now:
```
    testCase "even 4" $ equal' (evenOdd #. even # (4 :: Term s PInteger)) "(program 1.0.0 True)"
```

The new `(#.)`  operator applies the Scott-encoded record on the LHS to the Scott-encoded field on the right. The types work out, but the implementation is still messy. I'll see if I can extract the logic into a new module and generalize it to other ~kinds~ sorts of ADTs that don't need any letrec-ing. There is also the additional cost of type declaration

```
type instance ScottEncoded EvenOdd a = (PInteger :--> PBool) :--> (PInteger :--> PBool) :--> a
```

but I can add some Template Haskell code to generate these.

Anyway, I don't know why I'm replying to you because you should be in bed resting. Get back to health before you look again.

examples/Main.hs
```diff
@@ -46,6 +52,53 @@ example2 = plam $ \x -> pmatch x $ \case
   PLeft n -> n + 1
   PRight n -> n - 1
 
+data SampleRecord f = SampleRecord {
```
L-as, line 55: Why not replace `f` with `s` and have all the fields be `Term s ...` to make it match other Plutarch types?

L-as:
In general this is solid, but I feel like there might be a better way of doing this.

blamario:
Thanks, I guess, but unless you have a more concrete suggestion can you either

* approve the branch, so I can merge it and try to improve the approach later, or
* devote some time now to improve it.

With the current development pace of Plutarch, I find that repeated rebasing of the branch is not trivial. If you want to do something now I can outline what I tried and gave up on. For example, if the `PlutusType` class were split I could create its instances for all record types with `PInner  = ScottEncoding`, but I can't implement the `match'` method.

L-as:
@blamario What do you mean, it's already split?

examples/Main.hs
```diff
@@ -46,6 +52,53 @@ example2 = plam $ \x -> pmatch x $ \case
...
+  sampleBool :: f PBool,
+  sampleInt :: f PInteger,
+  sampleString :: f PString}
+
+type instance ScottEncoded SampleRecord a = PBool :--> PInteger :--> PString :--> a
+
+newtype PRecord r s = PRecord{_getRecord :: r (Term s)}
```
L-as, line 62: One issue I have with this is that the name is slightly confusing. Records should be defined in another way not using this mechanism, as that is simpler.

examples/Main.hs
```diff
@@ -46,6 +52,53 @@ example2 = plam $ \x -> pmatch x $ \case
...
+  sampleBool = pcon PTrue,
+  sampleInt = 12,
+  sampleString = "Hello, World!"}
+
+data EvenOdd f = EvenOdd {
+  even :: f (PInteger :--> PBool),
+  odd :: f (PInteger :--> PBool)
+  }
+
+type instance ScottEncoded EvenOdd a = (PInteger :--> PBool) :--> (PInteger :--> PBool) :--> a
```
L-as, line 88: Can you think of any way to avoid this?

L-as:
I will take a deeper look at this and try to improve it.

examples/Main.hs
blamario: The purpose of the newtype wrappers is to juggle the kinds: The record types like `SampleRecord` have a kind `((k -> Type) -> Type) -> Type` because the HKD pattern works with `(k' -> Type) -> Type`, but the `Term s`  wants a kind `k -> Type`. I also have a vague feeling that there should be a way to simplify all the kinds, but I don't have a concrete suggestion.

examples/Main.hs
blamario: Do you mean the existence of `ScottEncoded` or the need to declare its instance for every record type? For the latter, I can add a TH splice to automate the job, it would be merged into the existing `$(Rank2.TH.deriveAll SampleRecord)`.

blamario:
> What do you mean, it's already split?

I mean it's got two methods, `pcon'` and `pmatch'` and I can only implement the former. If the latter were moved into
```
class PlutusType t => MatchablePlutusType t where
     pmatch' :: forall s c. (forall b. Term s (PInner a b)) -> (a s -> Term s c) -> Term s c
```

then I could have
```
instance  (Rank2.Foldable r, Rank2.Distributive r) => PlutusType (PRecord r) where
  type PInner (PRecord r) a = ScottEncoding r a
  pcon' :: forall a s. PRecord r s -> Term s (ScottEncoding r a)
  pcon' (PRecord r) = plam' (\f-> punsafeCoerce $ appEndo (Rank2.foldMap (Endo . applyField) r) f)
    where applyField x f = punsafeCoerce f # x
```

Another idea I was looking into was
```
class UnCurry (r :: ((k -> Type) -> Type) -> Type) where
  puncurry :: Term s (ScottEncoded r a) -> PRecord r s -> Term s a

class Curry (r :: ((k -> Type) -> Type) -> Type) where
  pcurry :: (PRecord r s -> Term s a) -> Term s (ScottEncoded r a)
```
but I wasn't sure if you'd like that.

L-as:
`pcon` is already in its own class.

examples/Main.hs
L-as: Both, preferably, but just the need to declare iet would also be an improvement.

examples/Main.hs
L-as: Yeah, I get that, however I think users would be confused if they try using `PRecord` when they just want a record and not mutual recursion.

blamario:
To be honest, I'm surprised the instance
```
instance PlutusType a => PCon a where
  pcon = punsafeCoerce . pcon'
```
even compiles, I'd have expected it to require a constraint like `(PInner a b ~ a)`. Anyway, after getting over my fears and adding lots of `punsafeCoerce` calls I have something you might like better. The `(#.)` operator now takes a `Term s (PRecord r)` on the left hand side, and there's a general instance

    instance {-# overlaps #-} Rank2.Foldable r => PCon (PRecord r)

Also I went ahead and created that `Plutarch.Rec` module now that I have enough code to fill it.

My next step will be to create that Template Haskell splice for type instance ScottEncoded` boilerplate. I may also be able to add a generic `PLift` instance.

L-as:
Can you merge in upstream?

Plutarch/Evaluate.hs
```diff
@@ -42,4 +53,18 @@ evaluateScript s = do
...
   let s' = Script $ Program () (defaultVersion ()) $ termMapNames unNameDeBruijn term
-  pure (budget, logOut, s')
+  pure (usedBudget, logOut, s')
+
+{- | Evaluate a program in the CEK machine against the given budget, with the
+ usual text dynamic builtins and tracing, additionally returning the trace
+ output.
+-}
+evaluateCekBudgetedTrace ::
+  (uni ~ PLC.DefaultUni, fun ~ PLC.DefaultFun) =>
```
L-as, line 63: These constraints are unnecessary. Just replace `uni` with the relevant types.

L-as (commented):
What do you imagine the record stuff is going to be used for?

Plutarch/Evaluate.hs
blamario: Done.

blamario:
The `letrec` functionality would be used for mutually-recursive definitions, kinda like modules.

As for the plain records, I'm hoping they may end up more ergonomic than the `PDataList` style, in which case we'd replace the existing `ScriptContext` definition with a record. They're not there yet.

L-as:
> The `letrec` functionality would be used for mutually-recursive definitions, kinda like modules.
> 
> As for the plain records, I'm hoping they may end up more ergonomic than the `PDataList` style, in which case we'd replace the existing `ScriptContext` definition with a record. They're not there yet.

I don't think that's possible FWIW. The `ScriptContext` stuff can not use records because they are in fact not records. If you for example only want to access the first element in the list, then it's a waste to access the entire list.

L-as:
In any case if we wanted records we'd use the existing mechanism probably.

blamario:
`PPair` is the special case of same mechanism as `PRecord`: `PPair a b` is isomorphic to `PRecord (RecPair a b)` where
```
data RecPair a b f = RecPair{
  fst :: f a,
  snd :: f b}
```
The only thing missing from the latter is the `PlutusType` instance, because the `pmatch'` implementation will probably require more TH.

blamario:
> I don't think that's possible FWIW. The `ScriptContext` stuff can not use records because they are in fact not records. If you for example only want to access the first element in the list, then it's a waste to access the entire list.

I'm not sure what you mean by this. The `ScriptContext` in Plutus is defined [as a record](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Contexts.html#ScriptContext)
as is [`TxInfo`](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Contexts.html#TxInfo).

Record field access is O(1).

L-as:
> I'm not sure what you mean by this. The `ScriptContext` in Plutus is defined [as a record](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Contexts.html#ScriptContext) as is [`TxInfo`](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Contexts.html#TxInfo).
> 
> Record field access is O(1).

Unfortunately this isn't actually true. It's encoded as a list of data on-chain, hence `PDataList`.

L-as:
If we are to use this new mechanism for defining records, we probably want to create type aliases for the records, such that users only have to write e.g. `Term s PSomething`. `pmatch` is also necessary for them to be useful as records.

blamario:
I didn't get around to adding `pmatch` because I was hoping to generalize it upstream, but there's no fundamental problem with generating it in TH. Note however that `(#.)` already provides a partial workaround.

L-as:
The issue with `.#` is that it adds overhead. For each record access you'll call the record with a somewhat large (hoistable) lambda. We preferably want to use it like a continuation, and put all our computation into the lambda we pass to the record such that we don't need to make so many lambdas. This way also naturally enables standard Haskell record functionality within the `pmatch`, and also allows do-notation potentially.

L-as:
Another issue with this kind of record is that `pcon` works very awkwardly. I'm not sure what to do about that.

blamario:
? The `pcon` implementation behaves exactly the same as for `PPair`, except it's generalized.

L-as:
Yeah but you have to use the `PRecord` constructor too.

L-as:
We'd be sacrificing ease of use for ease of definition.

blamario:
Anyway specify what, if anything, you want done before you'll approve the PR. I'm going off.

L-as:
I think we should decouple the definition of mutually recursive functions from records for now, since records aren't quite where we want them yet, in the sense that we should perhaps use the record stuff in the PR only for defining mutually recursive functions.

We do want the TH stuff for records, and in general all data types really, (though it may be possible with just generics?), but since it's not ready yet we shouldn't recommend it for general use records IMO.

examples/Main.hs
```diff
@@ -48,6 +56,53 @@ example2 = plam $ \x -> pmatch x $ \case
   PLeft n -> n + 1
   PRight n -> n - 1
 
+data SampleRecord f = SampleRecord
```
L-as, line 59: We probably want a `examples/Examples/LetRec.hs` file or similar. We've been putting too many things in this one module.

Plutarch/Rec.hs
```diff
@@ -0,0 +1,104 @@
...
+  mapTerm,
+ )
+import qualified Rank2
+
+{- |
+  Highest precedence infixl operator, to be used like record field accessor. e.g.:
+
+  >>> record #. field
+-}
+(#.) ::
```
L-as, line 30: We probably want to rename/remove this since it's not useful (yet?), and will likely have to have its type changed anyway if it does become useful.

examples/Main.hs
```diff
@@ -48,6 +56,53 @@ example2 = plam $ \x -> pmatch x $ \case
...
+
+evenOddBuilder :: EvenOdd (Term s) -> EvenOdd (Term s)
+evenOddBuilder EvenOdd {even, odd} =
+  EvenOdd
+    { even = plam $ \n -> pif (n #== 0) (pcon PTrue) (odd #$ n - 1)
+    , odd = plam $ \n -> pif (n #== 0) (pcon PFalse) (even #$ n - 1)
+    }
+
+evenOdd :: Term s (PRecord EvenOdd)
+evenOdd = letrec evenOddBuilder
```
L-as, line 101: This isn't really an issue, but I am wondering why you didn't just inline `evenOddBuilder` into this function. Wouldn't that be cleaner?

L-as (changes requested):
Only minor changes needed, otherwise looks great to me!

examples/Main.hs
blamario: I just wanted to have its type signature in the clear, that's the only reason. I demoted the function with ccbfd32.

Plutarch/Rec.hs
blamario: If I remove it only to inline its uses in tests, I'll end up with unsightly mess:

```
-        , testCase "even 5" $ equal' (evenOdd #. even # (5 :: Term s PInteger)) "(program 1.0.0 False)"
+        , testCase "even 5" $ equal' (punsafeCoerce evenOdd # getScott (even accessors) # (5 :: Term s PInteger)) "(program 1.0.0 False)"
```

blamario: I've got something better now:

    testCase "even 5" $ equal' (evenOdd # field even # (5 :: Term s PInteger)) "(program 1.0.0 False)"

L-as (approved):
LGTM

PR #62, opened 2022-01-04
Add `pconstant` and `plift`

### LINKED ISSUE #51: Unlifting script evaluation results to Haskell
We can kinda work around it by comparing compiled term representations, but it's going to be pretty necessary to unlift term-encoded values from `Script` evaluation results - mostly for testing.

## Some things I tried

I tried hacking around with this for a while, and got some of the way towards it with `readKnownSelf`, but as usual with plutus-core it was pretty headache inducing.

I also tried going via the `flat` serialization (which I think could also work with Scott-encoded terms), but didn't get too far with that either.

### LINKED ISSUE #53: Add `pconstant` and `plift`
```haskell
pconstant :: C a b => a -> Term s b
plift :: C a b => Term s b -> a
```

The typeclass `C` should mirror `DefaultUni`. You could use associated type families rather than functional dependencies.

### LINKED ISSUE #61: Make ImportQualifiedPost the default (or remove it entirely)
~~The project inconsistently uses `ImportQualifiedPost`,~~ which breaks idempotency of fourmolu. ~~Enable ImportQualifiedPost throughout, and add it to .cabal file, if we are using it.~~

srid:
@L-as Is this a right approach for resolving #53 ?

- [x] Initial implementation
- [x] Clarify `plift` semantics
- [x] Use `plift` in examples
- [x] Replace hand-written functions with plift/pconstant
- [x] Avoid having to (unsafely) specify Plutus type, as in: `pconstant @Integer @PInteger 3`
- [x] Fix confusing naming
- [x] Reevaluate PList in context of DerivingVia (see comment)

Resolves #53 
Resolves #61 
Closes #51

### DIFF
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -19,7 +19,6 @@ module Plutarch (
   PI.plet,
   PI.punsafeBuiltin,
   PI.punsafeCoerce,
-  PI.punsafeConstant,
   PI.Term,
   PI.TermCont (..),
   PlutusType (..),
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -21,6 +21,9 @@ module Plutarch (
   PI.punsafeCoerce,
   PI.Term,
   PI.TermCont (..),
+  PL.plift,
+  PL.pconstant,
+  PL.plift',
   PlutusType (..),
   printTerm,
   printScript,
@@ -41,6 +44,7 @@ module Plutarch (
 import Data.Kind (Type)
 import Plutarch.Internal (ClosedTerm, Term, compile, papp, phoistAcyclic, plam', punsafeCoerce, (:-->))
 import qualified Plutarch.Internal as PI
+import qualified Plutarch.Lift as PL
 import Plutus.V1.Ledger.Scripts (Script (Script))
 import PlutusCore.Pretty (prettyPlcReadableDebug)
 
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -19,6 +19,7 @@ module Plutarch (
   PI.plet,
   PI.punsafeBuiltin,
   PI.punsafeCoerce,
+  PI.punsafeConstant,
   PI.Term,
   PI.TermCont (..),
   PL.plift,
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -18,10 +18,11 @@ module Plutarch.Builtin (
   PAsData,
 ) where
 
-import Plutarch (punsafeBuiltin, punsafeCoerce, punsafeConstant)
+import Plutarch (punsafeBuiltin, punsafeCoerce)
 import Plutarch.Bool (PBool, PEq, (#==))
 import Plutarch.ByteString (PByteString)
 import Plutarch.Integer (PInteger)
+import Plutarch.Lift (Lift (pconstant), PDefaultUni (..))
 import Plutarch.Prelude
 import qualified PlutusCore as PLC
 import PlutusTx (Data)
@@ -49,6 +50,9 @@ data PData s
 instance PEq PData where
   x #== y = punsafeBuiltin PLC.EqualsData # x # y
 
+instance PDefaultUni PData where
+  type PDefaultUniType PData = Data
+
 pfstBuiltin :: Term s (PBuiltinPair a b :--> a)
 pfstBuiltin = phoistAcyclic $ pforce . pforce . punsafeBuiltin $ PLC.FstPair
 
@@ -71,7 +75,7 @@ pasByteStr :: Term s (PData :--> PByteString)
 pasByteStr = punsafeBuiltin PLC.UnBData
 
 pdataLiteral :: Data -> Term s PData
-pdataLiteral = punsafeConstant . PLC.Some . PLC.ValueOf PLC.DefaultUniData
+pdataLiteral = pconstant
 
 data PAsData (a :: k -> Type) (s :: k)
 
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -22,7 +22,7 @@ import Plutarch (punsafeBuiltin, punsafeCoerce)
 import Plutarch.Bool (PBool, PEq, (#==))
 import Plutarch.ByteString (PByteString)
 import Plutarch.Integer (PInteger)
-import Plutarch.Lift (Lift (pconstant), PDefaultUniType)
+import Plutarch.Lift
 import Plutarch.Prelude
 import qualified PlutusCore as PLC
 import PlutusTx (Data)
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -77,6 +77,7 @@ pasInt = punsafeBuiltin PLC.UnIData
 pasByteStr :: Term s (PData :--> PByteString)
 pasByteStr = punsafeBuiltin PLC.UnBData
 
+{-# DEPRECATED pdataLiteral "Use `pconstant` instead." #-}
 pdataLiteral :: Data -> Term s PData
 pdataLiteral = pconstant
 
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -16,17 +16,14 @@ module Plutarch.Builtin (
   pdataLiteral,
   PIsData (..),
   PAsData,
-  pconstantData,
-  pliftData,
 ) where
 
-import Plutarch (ClosedTerm, punsafeBuiltin, punsafeCoerce)
+import Plutarch (punsafeBuiltin, punsafeCoerce)
 import Plutarch.Bool (PBool, PEq, (#==))
 import Plutarch.ByteString (PByteString)
 import Plutarch.Integer (PInteger)
 import Plutarch.Lift
 import Plutarch.Prelude
-import qualified Plutus.V1.Ledger.Api as Ledger
 import qualified PlutusCore as PLC
 import PlutusTx (Data)
 
@@ -54,15 +51,6 @@ data PData s
   | PDataInteger (Term s PInteger)
   | PDataByteString (Term s PByteString)
 
-pconstantData :: forall k (p :: k -> Type) h (s :: k). Ledger.ToData h => h -> Term s p
-pconstantData = punsafeCoerce . pconstant @PData . Ledger.toData
-
-pliftData :: forall k (p :: k -> Type) h. Ledger.FromData h => ClosedTerm p -> Either LiftError h
-pliftData t = do
-  h <- plift' @PData (punsafeCoerce t)
-  maybeToRight "Failed to decode data" $ Ledger.fromData h
-  where
-    maybeToRight e = maybe (Left e) Right
 instance PEq PData where
   x #== y = punsafeBuiltin PLC.EqualsData # x # y
 
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 -- This should have been called Plutarch.Data...
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -16,14 +16,17 @@ module Plutarch.Builtin (
   pdataLiteral,
   PIsData (..),
   PAsData,
+  pconstantData,
+  pliftData,
 ) where
 
-import Plutarch (punsafeBuiltin, punsafeCoerce)
+import Plutarch (ClosedTerm, punsafeBuiltin, punsafeCoerce)
 import Plutarch.Bool (PBool, PEq, (#==))
 import Plutarch.ByteString (PByteString)
 import Plutarch.Integer (PInteger)
 import Plutarch.Lift
 import Plutarch.Prelude
+import qualified Plutus.V1.Ledger.Api as Ledger
 import qualified PlutusCore as PLC
 import PlutusTx (Data)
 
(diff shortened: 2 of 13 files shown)


srid:
> Avoid having to (unsafely) specify Plutus type, as in: pconstant @Integer @PInteger 3

Perhaps by using associated type families.

Plutarch.hs
```diff
@@ -19,7 +19,6 @@ module Plutarch (
   PI.plet,
   PI.punsafeBuiltin,
   PI.punsafeCoerce,
-  PI.punsafeConstant,
```
srid, line 22: I decided to unexport this to create an expectation that users should not use it, and instead should use `pconstant`.

Plutarch/Unit.hs
```diff
@@ -1,14 +1,14 @@
...
+import Plutarch.Lift
 
 data PUnit s = PUnit
 
 instance PlutusType PUnit where
-  type PInner PUnit _ = POpaque
-  pcon' PUnit = punsafeConstant . PLC.Some $ PLC.ValueOf PLC.DefaultUniUnit ()
+  type PInner PUnit _ = PUnit
+  pcon' PUnit = pconstant ()
   pmatch' _ f = f PUnit
```
srid, line 12: I don't know why this was using `POpaque`, but switching to `PUnit` passes type tetris. 🤷🏿

Plutarch/Builtin.hs
```diff
@@ -71,7 +78,7 @@ pasByteStr :: Term s (PData :--> PByteString)
 pasByteStr = punsafeBuiltin PLC.UnBData
 
 pdataLiteral :: Data -> Term s PData
-pdataLiteral = punsafeConstant . PLC.Some . PLC.ValueOf PLC.DefaultUniData
+pdataLiteral = pconstant
```
srid, line 81: Should we retain functions like `pdataLiteral`, or remove them while replacing their calls with `pconstant`?

Plutarch/Lift.hs
```diff
@@ -0,0 +1,77 @@
...
+plift :: (PLift h p, HasCallStack) => ClosedTerm p -> h
+plift prog = either (error . show) id $ plift' prog
+
+instance (PLC.DefaultUni `PLC.Contains` h, PDefaultUniType p ~ h) => PConstant p h where
+  pconstant =
+    punsafeConstant . PLC.Some . PLC.ValueOf (PLC.knownUniOf (Proxy @h))
+
+instance PLC.KnownTypeIn PLC.DefaultUni (UPLC.Term PLC.DeBruijn PLC.DefaultUni PLC.DefaultFun ()) h => PLift h p where
+  plift' prog =
+    case evaluateScript (compile prog) of
```
L-as, line 66: Can we somehow move this logic into `plift` and make `plift'` as small as possible?

Plutarch.hs
L-as: I get this, but generally I think we should preserve backward compatibility. It's better to just deprecate it IMO.

Plutarch/Builtin.hs
L-as: We should deprecate the old ones.

Plutarch.hs
srid: I think that's generally a good idea -- and I'll add deprecation notice to `pdataLiteral` and friends -- but we can't deprecate `punsafeConstant` yet, because it is used internally by Plutarch:

image

srid: By the way, in this PR `punsafeConstant` is used only in two places:
- `Plutarch.Lift`
- `uplcTests` that you added.

L-as: We can `-Wno-deprecate` the relevant modules, but that's quite ugly. A better solution is to create a new name for `punsafeConstant` (e.g. `punsafeConstantInternal`), make `punsafeConstant` be an alias for that, and only deprecate the alias.

Plutarch/Lift.hs
srid: Not sure how. It is essentially a wrapper around `evaluateScript` and `readKnownSelf` (with some error wrapping) which we need in the instance. And the type is already small enough I think:

```haskell
plift' :: ClosedTerm p -> Either LiftError h
```

Plutarch.hs
srid: Done.

Plutarch/Builtin.hs
srid: Done.

Plutarch/Lift.hs
srid: Taking a step back: do we even really need these two type classes (`PConstant` and `PLift`)? They are never instantiated outside the module.

L-as: For one, I think you can likely merge them.

L-as: I think we might also want *not* to export them. You could export them from an internal module. Downstream users should not define their own instances.

L-as: Also, you might actually be able to remove the type classes and just use the same constraints as used for `readKnownSelf` and such.

srid: Done. The type class is no more.

srid:
I've opened an identical PR at #64 where the CI runs. This one can be closed.

srid:
Oh never mind. The CI started running on this one magically after I opened that.

L-as:
If you put the commit on any branch in this repository it's run. I often push other people's work to `tmp` to get CI to run.

L-as:
Can you add a test for a nested data type? E.g. a list of pairs of lists of integers.

srid:
> Can you add a test for a nested data type? E.g. a list of pairs of lists of integers.

Good idea ... done.

fourmolu is broken with this commit, though:

```
❯ bin/format 
+ find -type f -name '*.hs' '!' -path '*/dist-newstyle/*'
+ xargs fourmolu -o-XTypeApplications -m inplace
Loaded config from: /home/srid/work/ardana/plutarch/fourmolu.yaml
The GHC parser (in Haddock mode) failed:
  examples/Main.hs:252:13-18
  parse error on input plift'
```

It doesn't like some TypeApplications.

srid:
> fourmolu is broken with this commit,

Okay, fixed. Apparently `$do` (no space between) triggers it in this context.

L-as:
Now that I think about it, one reason to perhaps keep the type classes is that we want to support `pconstant` and `plift` for `ScriptContext` and similar. It would look like `pconstant :: ScriptContext -> Term s PScriptContext` in this case. Internally it would just be a data literal obviously, using the `ToData` instance for `ScriptContext`, but the resultant type would be different. The current solution is to use a data literal then cource it, but that's unsafe.

srid:
Wouldn't every instance, for every sum/product data type (including `ScriptContext`), look _identical_? Provisionally I came up with this instance for PScriptContext:

```haskell
instance PLift PScriptPurpose ScriptPurpose where
  pconstant = punsafeCoerce . pconstant @PData . toData
  plift' t = do
    h <- plift' @PData @Data (punsafeCoerce t)
    case fromData h of
      Nothing -> Left $ LiftError_Custom "No data"
      Just v -> pure v
```

<small>(By the way, this triggers overlapping instances because the original instance is on generic types constrained by open type family, which needs some looking into ...)</small>

I can see this would allow us to do the following, so in that sense this class would be useful I suppose:

```diff
--- a/examples/Main.hs
+++ b/examples/Main.hs
@@ -183,8 +183,8 @@ plutarchTests =
     , testCase "ScriptPurpose literal" $
         let d :: ScriptPurpose
             d = Minting dummyCurrency
-            f :: Term s PData
-            f = pconstant $ toData d
+            f :: Term s PScriptPurpose
+            f = pconstant @PScriptPurpose @ScriptPurpose d
          in printTerm f @?= "(program 1.0.0 #d8799f58201111111111111111111111111111111111111111111111111111111111111111ff)"
     , testCase "decode ScriptPurpose" $
         let d :: ScriptPurpose
```

srid:
Restored `PLift` in recent commits. 

Deriving an instance for ScriptContext and friends looks like this (this instance is now in the PR):

```haskell
instance {-# OVERLAPPING #-} PLift PScriptContext Ledger.ScriptContext where
  pconstant = pconstantData
  plift' = pliftData
```

Note: the implementation is identical - since they all use `Data` - so we just reuse `pconstantData` and `pliftData`. 

(I may be missing something here in regards to avoiding even this boilerplate ... )

TotallyNotChase:
This should also close #51 correct?

Plutarch/ScriptContext.hs
```diff
@@ -41,6 +43,10 @@ data PScriptPurpose s
   | PCertifying (Term s (PDataList '[POpaque]))
   deriving (PMatch, PIsData) via (PIsDataReprInstances PScriptPurpose)
 
+instance {-# OVERLAPPING #-} PLift PScriptPurpose Ledger.ScriptPurpose where
```
TotallyNotChase, line 46: Can we use some `DerivingVia` machinery here

L-as:
Both `pconstantData` and `pliftData` are unsafe functions, in addition you probably want to use the `PIsData` class and such.

L-as:
Actually, only `pconstant` has to be unsafe.

srid:
Okay, I used DerivingVia to auto-derive `PList` for ScriptContext and friends in 404dd1f86983a32cf395143adea646f2370cf342 - but this forced me to swap the arguments to `PList`, so that I can write `deriving PLift Ledger.ScriptPurpose via ...`.

`FlipC` from constraints-extras didn't work at first pass, and I could give it more try - however, I'm wondering if I should rethink `PList` itself, and have our primitive P* types derive-via it in a similar manner. Would also avoid overlapping instances.

srid:
Basically,

```haskell
data PBuiltinType p h s 
...
data PBool s = PTrue | PFalse
  deriving (PLift Bool) via PBuiltinType PBool Bool
```

Plutarch/DataRepr.hs
```diff
@@ -124,3 +126,12 @@ instance PIsDataRepr a => PIsData (PIsDataReprInstances a) where
 
 instance PIsDataRepr a => PMatch (PIsDataReprInstances a) where
   pmatch x f = pmatchRepr (punsafeCoerce x) (f . PIsDataReprInstances)
+
+instance {-# OVERLAPPING #-} (Ledger.FromData h, Ledger.ToData h, PIsData p) => PLift h (PIsDataReprInstances p) where
```
L-as, line 130: Do you have to mark this `OVERLAPPING` even though the other instance is `OVERLAPPABLE`?

L-as:
> Okay, I used DerivingVia to auto-derive `PList` for ScriptContext and friends in [404dd1f](https://github.com/Plutonomicon/plutarch/commit/404dd1f86983a32cf395143adea646f2370cf342) - but this forced me to swap the arguments to `PList`, so that I can write `deriving PLift Ledger.ScriptPurpose via ...`. But this is not good from user point of view due to extra type application; ` pconstant @_ @PInteger`.
> 
> `FlipC` from constraints-extras didn't work at first pass, and I could give it more try - however, I'm wondering if I should rethink `PList` itself, and have our primitive P* types derive-via it in a similar manner. Would also avoid overlapping instances.

Couldn't you rename `pconstant` to `pconstant'` then make a new `pconstant` that flips around the type parameters?

Plutarch/DataRepr.hs
srid: I recall the compiler forced me to (even though I know it should not be required), but both overlap pragmas are gone now in fa0b1d02e80bc69d4a3b5e3fd1d27aac8c3083ff

srid:
@L-as Some substantial revamp of `PLift` in fa0b1d02e80bc69d4a3b5e3fd1d27aac8c3083ff - mainly it is now `PList p` ie., without the `h` argument (`h` is determined by the instance). This change also enables DerivingVia for P* builtin types:

```haskell
data PInteger s
  deriving (PLift) via PBuiltinType PInteger Integer
```

or nested types:

```haskell
data PBuiltinList (a :: k -> Type) (s :: k)

deriving via
  PBuiltinType (PBuiltinList a) [PHaskellType a]
  instance
    PLC.DefaultUni `PLC.Contains` PHaskellType a => (PLift (PBuiltinList a))
```

And we are back to one type application now, `pconstant @PInteger 42`. No more overlapping instances, which I think is a good thing.

---

You originally requested `C a b` - but think the 2nd argument is not needed, since we usually want `a` to determine `b` anyway.

L-as:
> You originally requested C a b - but think the 2nd argument is not needed, since we usually want a to determine b anyway.

Yeah, it was just for the sake of clarity, because it's harder to make associated type families clear.

Plutarch/Builtin.hs
```diff
@@ -1,3 +1,6 @@
+{-# LANGUAGE DerivingVia #-}
```
L-as, line 1: You should just add `DerivingVia` to `plutarch.cabal`

Plutarch/DataRepr.hs
```diff
@@ -112,15 +114,25 @@ pmatchDataRepr d handlers =
               handler
               $ go common (idx + 1) rest constr
 
-newtype PIsDataReprInstances a s = PIsDataReprInstances (a s)
+newtype PIsDataReprInstances a h s = PIsDataReprInstances (a s)
```
L-as, line 117: Do we need this `h`?

srid: Yes. The `h` is used by the `PLift` instance [here](https://github.com/Plutonomicon/plutarch/pull/62/files#diff-ea6f5550d24dc825b81b16e44af9bcb35390ed76d1d2522473df83429b44ac6eR83) to associate with `PHaskellType`.

This is the also reason why `PBuiltinType` has a `h` in it [here](https://github.com/Plutonomicon/plutarch/pull/62/files#diff-ea6f5550d24dc825b81b16e44af9bcb35390ed76d1d2522473df83429b44ac6eR74). I wonder if I should rename `PBuiltinType`.

srid: And at the use site:

```haskell
 deriving (PMatch, PIsData, PLift) via (PIsDataReprInstances PScriptPurpose Ledger.ScriptPurpose)
```

^ Associates the Haskell type as `h`.

srid: Alternatively, we can leave `PIsDataReprInstances` untouched, and create a new newtype (analogous to `PBuiltinType`, but for non-builtin types):

```haskell
newtype PUserType p h s = PUserType (p s)

...
 deriving (PMatch, PIsData) via (PIsDataReprInstances PScriptPurpose)
 deriving (PLift) via (PUserType PScriptPurpose Ledger.ScriptPurpose)
...
```

L-as:
Do you think anything is missing? This seems mergeable to me.

srid:
@L-as It is ready, as long as you are happy with the naming of the new types/constraints (if not let me know your suggestions).

L-as (approved):
LGTM

L-as:
Thanks! :heart:

Issue #56, opened 2022-01-04
The implications of scott encoded types in custom datums and redeemers


TotallyNotChase:
Imagine `PMaybe` - it's represented with scott encoding. As in, it is not a `BuiltinData`. In due time, people are going to be implementing `PIsDataRepr` for their custom datums and redeemers - for easier and safer matching.

What happens when they try to use `PMaybe` in their custom datum/redeemer?

Do we implement `PIsData` for `PMaybe`? Doesn't that mean there are now 2 Plutus Core representations for `PMaybe` and room for ambiguity that cannot be caught in the type system?

Do we disallow usage of `PMaybe` in custom datum/redeemer by not implementing `PIsData`? Does that mean we need another `PMaybe` represented purely as builtin data? Won't it be a pain to work with both if certain APIs require one and not the other? Will that mean we'll need a way to convert between the two as well?

L-as:
https://github.com/Plutonomicon/plutarch/pull/63#issuecomment-1006643987

PR #63, opened 2022-01-04
Add `PList` `PlutusType` wrapper for `PBuiltinList`s

emiflake:
Summary: Add `PList` type that wraps `PBuiltinList`s containing `PData`. A couple of helper functions for working with lists. I've hopefully struck a good balance between ergonomics and performance, as the produced seems to mirror what I've seen when using just the `PBuiltinList` type. Should this exist as a separate library? `Bool`/`Integer` has a bunch of helper functions too, so I assume it's okay for now.

TODO:
- [x] Figure out an efficient way to deal with wrapping and unwrapping list elements. N/A with `ScottList`s. Can convert to/from `BuiltinList` using `pconvertLists`
- [x] Ergonomic lists from Haskell types after #53.


Plutarch/List.hs
```diff
@@ -0,0 +1,189 @@
...
+      (\self x xs -> f # x # (self # xs))
+      (\_self -> z)
+
+punsafeHead :: PIsData a => Term s (PList a :--> a)
+punsafeHead =
+  phoistAcyclic $
+    plam $ \xs ->
+      pmatch xs $ \case
+        PCons x _ -> x
+        PNil -> perror
```
srid, line 135: We can use `HeadList` directly on this, as it is partial anyway? Same for `punsafeTail`.

Plutarch/List.hs
```diff
@@ -0,0 +1,189 @@
...
+--------------------------------------------------------------------------------
+
+-- | Meta recursive list eliminator
+precList :: PIsData a => (Term s (PList a :--> r) -> Term s a -> Term s (PList a) -> Term s r) -> (Term s (PList a :--> r) -> Term s r) -> Term s (PList a :--> r)
+precList mcons mnil =
+  pfix #$ plam $ \self xs ->
+    pmatch xs $ \case
+      PCons x xs -> mcons self x xs
+      PNil -> mnil self
+
```
srid, line 120: I wonder what the budget difference here would be when using [`ChooseList`](https://github.com/input-output-hk/plutus/blob/c8c5183f7facd967d48fe07b3b14465b8dd48fe7/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L286-L297) directly.

emiflake: Yeah, good point. Maybe using `ChooseList` is better for the `pmatch'` implementation. I'll figure out the budget difference and pick the best one.

Plutarch/List.hs
emiflake: Maybe we should leave these out entirely then? Or make them return `PMaybe`s.

Plutarch/List.hs
emiflake: Alright, there's a slight benefit to be gained by using `ChooseList`:

```hs
-- pif + pnullBuiltin
Right (ExBudget {exBudgetCPU = ExCPU 895561, exBudgetMemory = ExMemory 2165},[],<Script>)
-- pchooseListBuiltin
Right (ExBudget {exBudgetCPU = ExCPU 776468, exBudgetMemory = ExMemory 1764},[],<Script>)
```

L-as (commented):
In general the code is great, however, I don't think you need the `PList` "newtype". You could add the constructors to `PBuiltinList` (which would be morally backward compatible), and make all of the functions work for all lists, not just lists of data AFAICT.

See also #56 

The issue is that we will likely have two "families" of data types:
- Scott-encoded data types
- `Data`-encoded data types

We need the first one for general computation, but only the second one can be stored in datums, redeemers, and similar.

See also https://github.com/input-output-hk/plutus/issues/4209

If the above issue is fixed, we can likely merge the two families. Until then, we will likely have both. Let's e.g. say we want a red black tree on-chain. That's going to be represented with a Scott encoding, and to put that into a list you need a Scott-encoded list too (the built-in one doesn't support functions).

For this PR, i think what's best is to merge `PList` and `PBuiltinList`, call it `PBuiltinList`, possibly reexport it from `Plutarch.List`, and possibly deprecate the old export.

Another thing is that because of this, `Plutarch.List` is ambiguous. 
@srid before recommended that we make two subdirectories for the different families.
So you could make it `Plutarch.Builtin.List`.

emiflake:
> You could add the constructors to `PBuiltinList`

Yeah, this works, although it's a bit of a pain to get around the cyclic dependency (without using orphans.)

> and make all of the functions work for all lists, not just lists of data AFAICT.

Are you proposing I scott-encode `PBuiltinList`? I don't see how to put non-data in a `PBuiltinList` otherwise, even without the wrapper:

```hs
> printScript <$> eval (pconsBuiltin # (fromInteger @(Term _ PInteger) 0xdeadbeef) # pemptyListBuiltin)
*** Exception: HUnitFailure (Just (SrcLoc {srcLocPackage = "interactive", srcLocModule = "Ghci2", srcLocFile = "<interactive>", srcLocStartLine = 12, srcLocStartCol = 17, srcLocEndLine = 12, srcLocEndCol = 102})) "Script evaluation failed: EvaluationError [] \"(CekEvaluationFailure,Just (Apply () (Apply () (Force () (Builtin () MkCons)) (Constant () (Some (ValueOf integer 3735928559)))) (Constant () (Some (ValueOf list (data) [])))))\""

> printScript <$> eval (pconsBuiltin # (pdata $ fromInteger @(Term _ PInteger) 0xdeadbeef) # pemptyListBuiltin)
"(program 1.0.0 [#1adeadbeef])"
```

I agree it would be regrettable to reimplement all of these functions for both "types" of lists, though. At worst, we can implement it once for scott-encoding and have a conversion operation incurring the cost a single time. Maybe this scott/data dichotomy can be turned into a class, but that's out of scope.
```hs
ptoScottList :: PIsData a => PBuiltinList a -> PScottList a
pfromScottList :: PIsData a => PScottList a -> PBuiltinList a
```

Or maybe we need to generalize a little more and enable implementing lists more abstractly?
```hs
class PIsList (list :: (k -> Type) -> k -> Type) where
  type PElemConstraint list :: (k -> Type) -> Constraint
  pelimList :: PElemConstraint list a => Term s (a :--> list a :--> r) -> Term s (PDelayed r) -> Term s (list a :--> r)
  pconsList :: PElemConstraint list a => Term s (a :--> list a :--> list a)
  pnilList :: PElemConstraint list a => Term s (list a)
```
Then `PList` becomes a proper Scott-encoded list, and we can create instances for both. With `instance PIsList PBuiltinList where type PElemConstraint PBuiltinList = PIsData`

emiflake:
Okay, here's what it looks like with a `ListLike` typeclass. I think now with #62 both of the TODOs are completed.

Plutarch/Builtin.hs
```diff
@@ -68,9 +69,46 @@ pheadBuiltin = phoistAcyclic $ pforce $ punsafeBuiltin PLC.HeadList
...
+
+instance PListLike PBuiltinList where
+  type PElemConstraint PBuiltinList = InDefaultUni
+  pelimList match_cons match_nil =
+    plam $ \ls -> pmatch ls $ \case
+      PCons x xs -> match_cons # x # xs
+      PNil -> pforce match_nil
+  pconsList = plam $ \x xs -> pcon (PCons x xs)
+  pnilList = pcon PNil
+  punsafeHead = pheadBuiltin
```
L-as, line 106: Note: These aren't unsafe. Unsafety refers to unsoundness, when you have a type for a term that is not within that type, e.g. `punsafeCoerce (pcon PTrue) :: ClosedTerm PInteger`.

Plutarch/List.hs
```diff
@@ -0,0 +1,277 @@
...
+  -- | Canonical eliminator for list-likes.
+  --
+  -- Example:
+  -- > isEmpty :: PIsListLike list a => Term s (list a :--> PBool)
+  -- > isEmpty = pelimList (plam $ \ _x _xs -> pcon PFalse) (pdelay $ pcon PTrue)
+  pelimList :: PElemConstraint list a => Term s (a :--> list a :--> r) -> Term s (PDelayed r) -> Term s (list a :--> r)
+
+  -- | / O(1) /. Cons an element onto an existing list.
+  pconsList :: PElemConstraint list a => Term s (a :--> list a :--> list a)
```
L-as, line 73: Why not just `pcons` and `pnil`?

Plutarch/List.hs
```diff
@@ -0,0 +1,277 @@
...
+
+import Plutarch
+import Plutarch.Bool (PBool (..), PEq (..), pif, (#&&))
+import Plutarch.Integer (PInteger)
+import Plutarch.Pair (PPair (..))
+import Plutarch.Prelude
+
+import Data.Kind
+
+data PScottList (a :: k -> Type) (s :: k)
```
L-as, line 43: We have to think about naming. `PMaybe` is scott-encoded, and #66 it adding Data-encoded `PMaybe` for use within datums and similar.

L-as: Currently no prefix has been Scott-encoded, and `Builtin` has been used as a prefix for the Data-encoded ones, but that doesn't fit for e.g. `PScriptContext`. Perhaps we want to have some explicit relation between the Scott-encoded and Data-encoded variants.

L-as (changes requested):
This looks mostly fine, however, `master` needs to be merged in, and we need to think about naming.

Plutarch/List.hs
emiflake: Yeah, I suppose `PList` is fine for _this_ case, I'll rename.

L-as (approved):
LGTM

L-as:
One thing we have to think about is what to do with constraints and type classes in Plutarch.

E.g. https://github.com/Plutonomicon/plutarch/pull/63/files#diff-3e4793aafaeb7a3e085513e7fbe66023660ce529c5ba5f4de98ffc4553fec9b0R111

This is actually *monomorphised*, meaning that a separate function is generated for each `PListLike`. I'm not sure this is optimal. We have to think more about the schism between the two encodings.

Issue #43, opened 2021-12-29
Fill out all of `ScriptContext`


L-as:
We need to mirror https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Api.html#t:ScriptContext precisely.

(Do note that this is V1)

srid:
> (Do note that this is V1)

Is the list of tasks to be done in version 1 recorded somewhere? If not, perhaps we can make use of GitHub's Milestone feature (to the right).

TotallyNotChase:
Quick question, let's say I want to fill out the [signatories field in `TxInfo`](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Contexts.html#t:TxInfo), which is a list of [`PubKeyHash`](https://staging.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/Plutus-V1-Ledger-Crypto.html#t:PubKeyHash). Would something like this be *the ideal* way to do it-
```hs
newtype PPubKeyHash s = PPubKeyHash (Term s (PAsData PByteString))

instance PlutusType PPubKeyHash where
  type PInner PPubKeyHash _ = PAsData PByteString
  pcon' (PPubKeyHash x) = x
  pmatch' bdata f = f $ PPubKeyHash bdata

data PTxInfo s
  = PTxInfo
      ( Term
          s
          ( PDataList
              '[ PBuiltinList PTxInInfo
               , PBuiltinList PTxOut
               , PValue -- fee
               , PValue -- mint
               , PBuiltinList POpaque -- dcert
               , POpaque -- withdrawals
               , POpaque -- range
               ---- see here  vvv
               , PBuiltinList PPubKeyHash -- signatures
               ---- see here  ^^^
               , POpaque -- data map
               , POpaque -- tx id
               ]
          )
      )
```

L-as:
@TotallyNotChase Another possibility:
```haskell
newtype PPubKeyHash s = PPubKeyHash (Term s PByteString)
  deriving newtype ... via (PAsData PByteString)
```

PR #66, opened 2022-01-05
ScriptContext (and transitive deps) types

t1lde:
Addressing #43 Draft because I haven't tested this yet,

Derived with the help of some TH, but the end result needed a lot of manual tweaking, so the TH isn't included.

It was a bit of a headache reasoning about wrapping with PAsData vs not, and how to correctly derive based on that.

The assumption I went with is that values of `PCurrencySymbol` are unwrapped, which `PasData PCurrency` symbol would be wrapped as Data. 
This lines up with the return type of `pdhead`, too, but definitely has some gotcha potential.

Further work is to be done adding API functions such as field accessors.


L-as (changes requested):
You probably want to add support for the `pconstant` stuff.

L-as:
I want to note that if you mess up the `PAsData` stuff, you will just get type checking errors when you try to `pdhead` it, i.e. it's not unsafe if you make a mistake in that specific regard, nor is it a breaking change to fix it later.

Plutarch/Api/V2.hs
```diff
@@ -0,0 +1,89 @@
+{-# LANGUAGE DerivingVia #-}
+module Plutarch.Api.V2
```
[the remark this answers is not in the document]
t1lde: I have been under the impression that the V2 ScriptContext is always been supported on the ledger, with (compatible) differences in de-serialization...

I really hope that's the case, too. `txInfoRedeemers` is pretty necessary.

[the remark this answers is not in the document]
L-as: `txInfoRedemeers` is not real. It's never been real. It's essentially just an elaborate prank. https://github.com/input-output-hk/plutus/issues/4285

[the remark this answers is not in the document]
t1lde: Ouch.

L-as:
FYI there are still merge conflicts

Plutarch/Lift.hs
```diff
@@ -107,3 +112,28 @@ instance
 
 showEvalException :: EvaluationException CekUserError (MachineError PLC.DefaultFun) (UPLC.Term UPLC.DeBruijn PLC.DefaultUni PLC.DefaultFun ()) -> Text
 showEvalException = T.pack . show
+
+{- | 
+  DerivingVia wrapper for deriving `PLift` instances
+  via the wrapped type, while lifting to a coercible Haskell type.
+
+-}
+newtype PLiftVia (p :: k -> Type) (h :: Type) s = PLiftVia (p s)
```
L-as, line 121: Don't you want `(s :: k)`?

L-as (approved):
LGTM

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,626 @@
...
+  = PTxInfo
+      ( Term
+          s
+          ( PDataList
+              '[ PBuiltinList PTxInInfo
+               , PBuiltinList PTxOut
+               , PValue
+               , PValue
+               , PBuiltinList PDCert
+               , PBuiltinList (PBuiltinPair PStakingCredential PInteger)
```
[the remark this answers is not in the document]
TotallyNotChase: Looks like it's actually a builtin list of `Constr` data values. The Haskell tuple here has a `ToData` instance that creates a `Constr 0 [Data]`.

I think the `PIsData (PBuiltinPair PInteger (PBuiltinList PData))` should be removed in favor of having `(PIsData a, PIsData b)=> PIsData (PBuiltinPair a b)`

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,626 @@
...
+               , PPOSIXTimeRange
+               , PBuiltinList PPubKeyHash
```
TotallyNotChase, line 93: Curious question: Shouldn't this be `PBuiltinList (PAsData PPubKeyHash)` - `PPubKeyHash` is not a data value, rather a builtin bytestring - right? So its data representation would be `PAsData PPubKeyHash`? Compared to, say `PTxInfo`, which *is* a data value the entire time - not a builtin that would be converted to a data value in this context.

Plutarch/Api/V1.hs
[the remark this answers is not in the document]
L-as: The second instance wouldn't be valid AFAICT?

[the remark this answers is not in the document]
TotallyNotChase: Oh right, the `pfromData` wouldn't be possible. But the relationship seems to be `PlutusTx.toData (a, b) == Constr 0 [toData a, toData b]` and `PlutusTx.fromData (Constr 0 [toData a, toData b]) == Just (a, b)`.

Plutarch/Builtin.hs
```diff
@@ -71,6 +76,20 @@ ptailBuiltin = phoistAcyclic $ pforce $ punsafeBuiltin PLC.TailList
...
+  Note that the Plutus API doesn't use this most of the time,
+  instead encoding as a List of Tuple constructors.
+
+  Not to be confused with `PlutusTx.AssocMap.Map` / `PMap`
+
+-}
+type PBuiltinMap a b = (PBuiltinList (PBuiltinPair (PAsData a) (PAsData b)))
+
+-- | List type used for Plutus `Data`'s List constructor.
+type PList a  = PBuiltinList (PAsData a)
```
TotallyNotChase, line 91: I worry about whether or not it's redundant to write something like `PAsData PTxInfo` (in the case of `PList PTxInfo`). After all `PTxInfo` *is* `Data`, as opposed to, say `PPubKeyHash` - which does require this treatment.

L-as: The internal representation of `TxInfo` is an implementation detail.

t1lde: There's a choice I've made here for consistency -
all of the types are assumed to be 'unwrapped' by default from the Data constructor. 

As @L-as says, the internal repr is an implementation detail, the repr of `PAsData` can be arbitrarily.

t1lde:
I think this is ready, just need to get out of build hell to check the last merges...

Plutarch/DataRepr.hs
```diff
@@ -136,3 +146,36 @@ instance (Ledger.FromData h, Ledger.ToData h, PIsData p) => PLift (PIsDataReprIn
     maybeToRight "Failed to decode data" $ Ledger.fromData h
     where
       maybeToRight e = maybe (Left e) Right
+
+{- | 
+  DerivingVia wrapper for deriving `PLift` instances
+  via the wrapped type, while lifting to a coercible Haskell type.
+
+-}
+newtype PVia (p :: k -> Type) (h :: Type) (s :: k) = PVia (Term s p)
```
srid, line 155: If this is used to derive only `PLift` instances, I suggest putting it in `Plutarch.Lift` module, alongside `PBuiltinType` - because it looks like they serve, essentially, the same purpose (with the only difference being that `PVia` is used to derive new types, and `PBuiltinType` is used to derive builtin types).

I also wonder why it is named `PVia` (rather than say `PNewType` to be consistent with `PBuiltinType` naming).

t1lde: Good point, yeah. I'll move it over.

I'd say `PNewType` is less accurate than `PVia`, since it uses `Coercible`, it's a bit more powerful than just newtypes (same deal with deriving via vs deriving newtype, really).

t1lde: Actually, I'll keep `PVia` where it is, to avoid the module cycle due to the `PlutusType` instance.

Plutarch/List.hs
```diff
@@ -151,6 +152,19 @@ plength = phoistAcyclic $
     )
     $ \go -> plam $ \xs -> go # xs # 0
 
+{- |
+  Unsafely index a BuiltinList,
+  throwing an error if the index is out of bounds.
+-}
+punsafeIndex :: (PIsListLike list a) => Term s (PInteger :--> list a :--> a)
```
L-as, line 159: This isn't unsafe FWIW.

plutarch.cabal
```diff
@@ -1,39 +1,32 @@
...
-
+author:             Las Safin <me@las.rs>
+license:            MIT
 extra-source-files: README.md
 
 flag development
   description: Enable tracing functions within plutarch.
-  manual: True
-  default: False
+  manual:      True
```
L-as, line 10: Why did you change all the formatting here?

L-as (approved):
Looks mostly good to me, but I think we should merge this PR so other people can build on top of it.

I don't have many complaints wrt. the code but the PR is too big IMO. The lifting stuff could have been done in a separate PR, and would have meant we could have processed the work incrementally.

L-as:
#96

plutarch.cabal
t1lde: cabal-fmt did it

PR #70, opened 2022-01-05
getArityBuiltin: remove boilerplate using arity provided by Plutus Core

### LINKED ISSUE #69: Simplify getArityBuiltin
The current Plutarch implementation manually cases over all Plutus Builtins,

https://github.com/Plutonomicon/plutarch/blob/bf74874f7d113e27b3a784065ea525604fb3e80d/Plutarch/Internal.hs#L154-L206

But while working on `PLift` I noticed that Plutus provides the infra[^2] to simplify it as follows:

```haskell
    getArityBuiltin' :: PLC.DefaultFun -> Natural
    getArityBuiltin' builtin = 
      case PLC.toBuiltinMeaning @_ @_ @(PLC.Term PLC.TyName PLC.Name _ _ ()) builtin of 
        PLC.BuiltinMeaning sch _ _ -> intToNatural $ PLC.countTermArgs sch
```

<details>
<summary>Outdated remark</summary>
With this change, however, all but the following single test fails:

```
    λx y -> hoist mkCons x y => mkCons x y:                         2
FAIL
      examples/Main.hs:196:
      expected: "(program 1.0.0 (\\i0 -> \\i0 -> mkCons i2 i1))"
       but got: "(program 1.0.0 mkCons)"
      Use -p '/λx y -> hoist mkCons x y => mkCons x y/' to rerun this test only.
```

Is this test failure to be expected? `(program 1.0.0 mkCons)` is equivalent anyway, right? EDIT: discard this query; I did something wrong.
</summary>

[^2]: I think Plutus might even provide a way to calculate the force depth at the type-level

srid:
Resolves #69


Plutarch/Internal.hs
```diff
@@ -152,59 +156,26 @@ plam' f = Term $ \i ->
...
+    getArityBuiltin (RForce (RBuiltin builtin)) =
+      guard (getPolyVarCount builtin == 1) >> pure (getArityBuiltin' builtin)
+    getArityBuiltin (RForce (RForce (RBuiltin builtin))) =
+      guard (getPolyVarCount builtin == 2) >> pure (getArityBuiltin' builtin)
     getArityBuiltin _ = Nothing
 
+    getArityBuiltin' :: PLC.DefaultFun -> Natural
+    getArityBuiltin' builtin =
+      case PLC.toBuiltinMeaning @_ @_ @(PLC.Term PLC.TyName PLC.Name _ _ ()) builtin of
+        PLC.BuiltinMeaning sch _ _ -> intToNatural $ PLC.countTermArgs sch
```
L-as, line 170: I just made #72 and with the new tests, and this fails. You have an off-by-one error here. You need to do `-1` to get it to work.

L-as (changes requested):

Plutarch/Internal.hs
srid: Thanks; resolved now.

(Also: https://github.com/Plutonomicon/plutarch/pull/72#discussion_r779764290 )

srid:
Tests pass but the failure is in format checker. 

Unfortunate that we can't get CI to automatically run for foreign branches. I think the format check started failing since around #63 /cc @MatthewCroughan

L-as:
I fixed the formatting error in master.

srid:
Updated, and ready to be merged.

L-as (approved):
LGTM

PR #72, opened 2022-01-06
Add more η-reduction examples

L-as:


examples/Main.hs
```diff
@@ -188,24 +188,33 @@ plutarchTests =
...
-        printTerm (plet (phoistAcyclic $ plam $ \(x :: Term _ PInteger) -> x + x) $ \x -> x) @?= "(program 1.0.0 (\\i0 -> addInteger i1 i1))"
+    , testGroup
+        "η-reduction optimisations"
+        [ testCase "λx y. addInteger x y => addInteger" $
+            printTerm (plam $ \x y -> (x :: Term _ PInteger) + y) @?= "(program 1.0.0 addInteger)"
+        , testCase "λx y. hoist (force mkCons) x y => force mkCons" $
+            printTerm (plam $ \x y -> (pforce $ punsafeBuiltin PLC.MkCons) # x # y) @?= "(program 1.0.0 (force mkCons))"
+        , testCase "λx y. hoist mkCons x y => mkCons x y" $
+            printTerm (plam $ \x y -> (punsafeBuiltin PLC.MkCons) # x # y) @?= "(program 1.0.0 (\\i0 -> \\i0 -> mkCons i2 i1))"
+        , testCase "λx y. hoist (λx y. x + y - y - x) x y => λx y. x + y - y - x" $
```
srid, line 199: @L-as I understand the eta reduction in line 196 (the wrapping lambda is redundant). But why is the same not the case for line 198?

L-as: Because it changes the semantics slightly but importantly. Eta-reducing it to `λx y. hoist mkCons x y` to `mkCons` would cause different behavior when you pass an argument to it, since it would cause an error (it needs a force). If you don't η-reduce it, it would not err.

PR #76, opened 2022-01-07
Some scattered general cleanup

TotallyNotChase:
* Hoist the `ChooseList` and `MkCons` synonyms
* Export `pforgetData` - useful to get a `PData` out of a `PAsData a` for whatever `a`. This should be a safe operation regardless.
* Make `pnull` a `PListLike` class method so the builtin list instance can bind it to `pnullBuiltin`
* Remove redundant delay/force in `pelimList` - should be handled by underlying `pmatch` AFAIK
* General formatting cleanup + hoist a bunch of list utilities
* Make `IsDefaultUni` a type alias and export it - it'll be more convenient for users to write than `PLC.KnownType...` etc.
* Make `pelimList` take haskell level function arguments. `pelimList` itself is now also a haskell level function entirely.

### DIFF
--- a/Plutarch/List.hs
+++ b/Plutarch/List.hs
@@ -62,7 +62,7 @@ class PListLike (list :: (k -> Type) -> k -> Type) where
   type PElemConstraint list :: (k -> Type) -> Constraint
 
   -- | Canonical eliminator for list-likes.
-  pelimList :: PElemConstraint list a => Term s (a :--> list a :--> r) -> Term s (PDelayed r) -> Term s (list a :--> r)
+  pelimList :: PElemConstraint list a => Term s (a :--> list a :--> r) -> Term s r -> Term s (list a :--> r)
 
   -- | Cons an element onto an existing list.
   pcons :: PElemConstraint list a => Term s (a :--> list a :--> list a)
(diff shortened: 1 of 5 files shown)


Plutarch/List.hs
```diff
@@ -87,27 +90,27 @@ instance PListLike PList where
...
   pnil = pcon PSNil
 
 type PIsListLike list a = (PListLike list, PElemConstraint list a)
 
 -- | / O(n) /. Convert from any ListLike to any ListLike, provided both lists' element constraints are met.
 pconvertLists :: forall f g a s. (PElemConstraint f a, PElemConstraint g a, PListLike f, PListLike g) => Term s (f a :--> g a)
-pconvertLists =
+pconvertLists = phoistAcyclic $
   pfix #$ plam $ \self ->
     pelimList
```
TotallyNotChase, line 103: curious question: is `pelimList` sufficiently efficient to be using it in most list utilities?

Plutarch/List.hs
```diff
@@ -261,12 +260,12 @@ pzip ::
   , PElemConstraint list (PPair a b)
   ) =>
   Term s (list a :--> list b :--> list (PPair a b))
-pzip = pzipWith' $ \x y -> pcon (PPair x y)
+pzip = phoistAcyclic $ pzipWith' $ \x y -> pcon (PPair x y)
```
TotallyNotChase, line 263: curious question: would using `pzipWith #$ plam $ \x y -> pcon (PPair x y)`  be more efficient here or would it not matter?

Plutarch/List.hs
L-as: Good question. I haven't looked at it too much.

L-as: Actually, can you change it to be a zero-cost wrapper? Right now it does a bunch of unnecessary stuff. This will be a breaking change, but :shrug: it's only been a day. It should preferably incur no overhead compared to doing `pmatch`.

Plutarch/List.hs
L-as: `pzipWith'` is more efficient, since we only use the terms once. In general I feel like HOFs should be Haskell-level.

Plutarch/List.hs
TotallyNotChase: Do you mean something like (on the `PList` instance, for example)-
```hs
pelimList :: PElemConstraint list a => Term s (a :--> list a :--> r) -> Term s r -> Term s (list a) -> Term s r
pelimList match_cons match_nil ls = pmatch ls $ \case
    PSCons x xs -> match_cons # x # xs
    PSNil -> match_nil
```

I think we would ideally want to preserve the ability to take in a `match_cons` and `match_nil` separately - but I wonder whether or not they should be Haskell level functions.

TotallyNotChase: From looking at the usages of `pelimList` around the codebase - my intuition is that a haskell level `match_cons` would probably be better. But I might be wrong.

TotallyNotChase: I want to experiment with this a bit alongside a few other changed on the `pelimList-test` branch - I'll use the minswap rewrite to benchmark them.

I'll let you know when I'm happy with the results so you can look into merging this.

L-as: In general Haskell-level functions are probably better now due to the η-reduction optimisations. We still don't have any good inlining, so `(plam $ \x y -> x y) # x # y` does not reduce to `x y` unfortunately.

L-as: Yeah, we should have an issue for that.

L-as (approved):
LGTM

PR #77, opened 2022-01-07
#47

L-as:
#47


L-as:
I ended up not enabling the shrinker tests because they increased the time it took to run the tests by 1633 times.

PR #81, opened 2022-01-07
Add GHC flags for explicit foralls/ kinds

srid:
As long as we are going to be using the `k`, I thought this would be useful to have by default:

```haskell
> :k Term
Term :: forall k. k -> (k -> *) -> *
```

```haskell
> :k PBool
PBool :: forall {k}. k -> *
```

### DIFF
--- a/plutarch.cabal
+++ b/plutarch.cabal
@@ -63,7 +63,7 @@ common c
     ScopedTypeVariables
     StandaloneDeriving
     StandaloneKindSignatures
-    StarIsType
+    NoStarIsType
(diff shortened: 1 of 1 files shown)


srid:
And why not enable `NoStarIsType` while we are at it, since we use `Type` instead of `*`:

```haskell
> :k Term
Term :: forall k. k -> (k -> Type) -> Type
```

L-as:
Great, thanks!

L-as (approved):
LGTM

Issue #79, opened 2022-01-07
Benchmark list utilities and alter if needed


TotallyNotChase:
Originally raised from [this comment](https://github.com/Plutonomicon/plutarch/pull/76#discussion_r780071222).

We want to make sure the list utilities are sufficiently efficient, as they will be the basis for many ledger API functions. In particular, I'm concerned about-
* `pelemList` was benefitted by taking in Haskell level functions rather than Plutarch level functions. Are there other list utilities that might benefit from this?
* Are there any work duplication hidden in plain sight, within the abstractions?
  
  [Example of fixing work duplication](https://github.com/Plutonomicon/plutarch/pull/76/commits/44d06908ab06d2f8419f0f3e9d0b63aa34a07c96#diff-3e4793aafaeb7a3e085513e7fbe66023660ce529c5ba5f4de98ffc4553fec9b0R190).

Some essentials about Haskell level functions vs Plutarch level functions: #78

TotallyNotChase:
The plan is to do pretty basic benchmarks for singular list utilites one by one. The evaluation machine seems pretty consistent and predictable so it should be sufficient in this simple case. We also have a production level plutarch script that uses *some* of the utilities - so I sometimes cross reference there.

I *really really really* want that benchmark suite you proposed, for every big Plutarch change - it'd be really useful given Plutarch's goals. Definitely something to think about. I don't know if we should have it here of in pluton - but whatever works.

PR #84, opened 2022-01-10
Fix boolean functions not short circuiting

TotallyNotChase:
My bad, I messed up the original implementation. Should be fixed now.

`pand` and `por` (essentially synonyms of `#&&` and `#||`) were not short circuiting due to top level usage of `pif'`. Caught by checking traces.

### DIFF
--- a/Plutarch/Bool.hs
+++ b/Plutarch/Bool.hs
@@ -62,34 +62,26 @@ pnot = phoistAcyclic $ plam $ \x -> pif x (pcon PFalse) $ pcon PTrue
 infixr 3 #&&
 
 (#&&) :: Term s PBool -> Term s PBool -> Term s PBool
-x #&& y = pand # pdelay x # pdelay y
+x #&& y = pforce $ pand # x # pdelay y
 
 -- | Lazily evaluated boolean or for 'PBool' terms.
 infixr 2 #||
 
 (#||) :: Term s PBool -> Term s PBool -> Term s PBool
-x #|| y = por # pdelay x # pdelay y
+x #|| y = pforce $ por # x # pdelay y
 
 -- | Hoisted, Plutarch level, lazily evaluated boolean and function.
-pand :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
-pand = phoistAcyclic $
-  plam $
-    \x y -> pif (pforce x) (pif' # pforce y # pcon PTrue # pcon PFalse) (pcon PFalse)
+pand :: Term s (PBool :--> PDelayed PBool :--> PDelayed PBool)
+pand = phoistAcyclic $ plam $ \x y -> pif' # x # y # (phoistAcyclic $ pdelay $ pcon PFalse)
 
 -- | Hoisted, Plutarch level, strictly evaluated boolean and function.
 pand' :: Term s (PBool :--> PBool :--> PBool)
-pand' = phoistAcyclic $
-  plam $
-    \x y -> pif' # x # (pif' # y # pcon PTrue # pcon PFalse) # pcon PFalse
+pand' = phoistAcyclic $ plam $ \x y -> pif' # x # y # (pcon PFalse)
 
 -- | Hoisted, Plutarch level, lazily evaluated boolean or function.
-por :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
-por = phoistAcyclic $
-  plam $
-    \x y -> pif (pforce x) (pcon PTrue) (pif' # pforce y # pcon PTrue # pcon PFalse)
+por :: Term s (PBool :--> PDelayed PBool :--> PDelayed PBool)
+por = phoistAcyclic $ plam $ \x y -> pif' # x # (phoistAcyclic $ pdelay $ pcon PTrue) # y
 
 -- | Hoisted, Plutarch level, strictly evaluated boolean or function.
 por' :: Term s (PBool :--> PBool :--> PBool)
-por' = phoistAcyclic $
-  plam $
-    \x y -> pif' # x # pcon PTrue #$ pif' # y # pcon PTrue # pcon PFalse
+por' = phoistAcyclic $ plam $ \x y -> pif' # x # (pcon PTrue) # y
--- a/Plutarch/Bool.hs
+++ b/Plutarch/Bool.hs
@@ -74,7 +74,7 @@ x #|| y = por # pdelay x # pdelay y
 pand :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
 pand = phoistAcyclic $
   plam $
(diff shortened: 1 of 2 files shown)


Plutarch/Bool.hs
```diff
@@ -74,7 +74,7 @@ x #|| y = por # pdelay x # pdelay y
 pand :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
 pand = phoistAcyclic $
   plam $
-    \x y -> pif' # pforce x # (pif' # pforce y # pcon PTrue # pcon PFalse) # pcon PFalse
+    \x y -> pif (pforce x) (pif' # pforce y # pcon PTrue # pcon PFalse) (pcon PFalse)
```
L-as, line 77: I don't think this is entirely correct either. The reason this function takes `PDelayed PBool` is that we want to avoid unnecessary delays and forces. The function is in any case quite unoptimised. What we really want if something like
```haskell
pand :: Term s (PBool :--> PDelayed PBool :--> PBool)
pand = phoistAcyclic $ plam $ \x y -> pforce $ pif' x (phoistAcyclic $ pdelay $ pcon PTrue) y
```

TotallyNotChase: Aha, of course! That's some silly redundancy from my end

TotallyNotChase: Thanks for the fix!

Plutarch/Bool.hs
```diff
@@ -62,34 +62,26 @@ pnot = phoistAcyclic $ plam $ \x -> pif x (pcon PFalse) $ pcon PTrue
...
-x #|| y = por # pdelay x # pdelay y
+x #|| y = pforce $ por # x # pdelay y
 
 -- | Hoisted, Plutarch level, lazily evaluated boolean and function.
-pand :: Term s (PDelayed PBool :--> PDelayed PBool :--> PBool)
-pand = phoistAcyclic $
-  plam $
-    \x y -> pif' # pforce x # (pif' # pforce y # pcon PTrue # pcon PFalse) # pcon PFalse
+pand :: Term s (PBool :--> PDelayed PBool :--> PDelayed PBool)
+pand = phoistAcyclic $ plam $ \x y -> pif' # x # y # (phoistAcyclic $ pdelay $ pcon PFalse)
```
srid, line 75: Why do we need to hoist `pdelay $ pcon PFalse`  here (as well as in `por` below)?

examples/Main.hs
```diff
@@ -250,6 +250,16 @@ plutarchTests =
...
+    , testCase "True && False ≡ False" $ equal (pcon PTrue #&& pcon PFalse) (pcon PFalse)
+    , testCase "False && True ≡ False" $ equal (pcon PFalse #&& pcon PTrue) (pcon PFalse)
+    , testCase "False && perror ≡ False" $ equal (pcon PFalse #&& perror) (pcon PFalse)
+    , testCase "fails: pand False perror" $ fails $ pand # pcon PFalse # perror
+    , testCase "pand False (pdelay perror) ≡ False" $ equal (pand # pcon PFalse # pdelay perror) (pdelay $ pcon PFalse)
+    , testCase "True || False ≡ True" $ equal (pcon PTrue #|| pcon PFalse) (pcon PTrue)
+    , testCase "False || True ≡ True" $ equal (pcon PFalse #|| pcon PTrue) (pcon PTrue)
+    , testCase "True || perror ≡ True" $ equal (pcon PTrue #|| perror) (pcon PTrue)
+    , testCase "fails: por True perror" $ fails $ por # pcon PFalse # perror
+    , testCase "por True (pdelay perror) ≡ True" $ equal (por # pcon PTrue # pdelay perror) (pdelay $ pcon PTrue)
```
srid, line 262: nitpick: could group these using `testGroup`.

examples/Main.hs
```diff
@@ -250,6 +250,16 @@ plutarchTests =
             let v2 = [("IOHK", [1, 2, 3]), ("Plutus", [9, 8, 7])]
             plift' (pconstant @(PBuiltinList (PBuiltinPair PString (PBuiltinList PInteger))) v2) @?= Right v2
         ]
+    , testCase "True && False ≡ False" $ equal (pcon PTrue #&& pcon PFalse) (pcon PFalse)
+    , testCase "False && True ≡ False" $ equal (pcon PFalse #&& pcon PTrue) (pcon PFalse)
+    , testCase "False && perror ≡ False" $ equal (pcon PFalse #&& perror) (pcon PFalse)
+    , testCase "fails: pand False perror" $ fails $ pand # pcon PFalse # perror
```
srid, line 256: This doesn't seem right. Per the semantics of `fails` it expects _either_ user error (`EvaluationError`) _or_ engine error/bug (`EvaluationException`). Even `EvaluationError` is not precise:

```haskell
data EvaluationError user internal
    = InternalEvaluationError internal
      -- ^ Indicates bugs.
    | UserEvaluationError user
      -- ^ Indicates user errors.
```

What we want is to test that the result is exactly a `perror` (and not general evaluation error/exception).

Plutarch/Bool.hs
L-as: We don't need to. It will just reduce the size of the term.

examples/Main.hs
L-as: That is true.

examples/Main.hs
L-as: This feels like something for another PR. I don't touch the implementation of `fails` here.

examples/Main.hs
L-as: Should be fixed.

examples/Main.hs
srid: Filed https://github.com/Plutonomicon/plutarch/issues/85

Plutarch/Bool.hs
srid: Maybe, but it doesn't seem obvious to me that variable reference is necessarily smaller than inlined `(\ -> False)`. I suppose we can measure these precisely once some benchmarks are in place (cf. https://github.com/Plutonomicon/plutarch/issues/79#issuecomment-1007473155 )

srid (approved):

Plutarch/Bool.hs
L-as: It is indeed not obvious, and I am not 100% sure that this is the case either, but a variable reference ought to be smaller than `delay (constant ...)`, since the variable reference need not be any more than ~8 bits.

L-as: In any case it's inlined if there's only one use of it.

srid:
I overlooked something: we have another set of boolean tests here:

https://github.com/Plutonomicon/plutarch/blob/acc1b0a3be62c847aade9cf331aec6a6ac281592/examples/Main.hs#L148-L160

L-as:
lol

PR #89, opened 2022-01-11
Implement Plutarch Rationals

Geometer1729:
I'm not certain this is the best way to do Rationals in Plutarch, but we're going to need them fairly soon for Danaswap and as long as any future improvements can be backwards compatible with the Num and Fractional instances this should at least allow us to write and test code using rationals for now.


Plutarch/Rational.hs
```diff
@@ -1,28 +1,118 @@
...
-    pinl (pfst (punRational x)) $ \x1 ->
-    pinl (psnd (punRational x)) $ \x2 ->
-    pinl (pfst (punRational y)) $ \y1 ->
-    pinl (psnd (punRational y)) $ \y2 ->
-    prational (x1 * y1) (x2 * y2)
-    -}
+module Plutarch.Rational 
+  (PRational
+  ,preduce
+  ,pnumerator
```
L-as, line 6: ./bin/format

nix build .#check.x86_64-linux

Plutarch/Rational.hs
```diff
@@ -1,28 +1,118 @@
...
+  x / y = preduce #
+    pmatch x (\(PRational xn xd) ->
+    pmatch y (\(PRational yn yd) ->
+      pcon (PRational (xn*yd) (xd*yn))
+             ))
+  fromRational r = 
+    pcon $ PRational (fromInteger $ numerator r) (fromInteger $ denominator r)
+
+preduce :: Term s (PRational :--> PRational)
+preduce = plam $ \x ->
```
L-as, line 88: You need to hoist all your terms.

Plutarch/Rational.hs
```diff
@@ -1,28 +1,118 @@
...
+        1
+  fromInteger n = pcon $ PRational (fromInteger n) 1
+            
+instance Fractional (Term s PRational) where
+  recip x = 
+    pmatch x (\(PRational xn xd) ->
+      pcon (PRational xd xn))
+  x / y = preduce #
+    pmatch x (\(PRational xn xd) ->
+    pmatch y (\(PRational yn yd) ->
```
L-as, line 81: Please do `pmatch x $ ...`

Plutarch/Rational.hs
```diff
@@ -1,28 +1,118 @@
...
+pgcd' = pfix #$ plam $ f
+  where
+    f self a' b' =
+        plet (pmax # a' # b') $ \a ->
+        plet (pmin # a' # b') $ \b ->
+          pif (b #== 0)
+            a
+            (self # (pmod # a # b) # b)
+
+pmin,pmax :: POrd a => Term s (a :--> a :--> a)
```
L-as, line 109: Please don't declare multiple functions together. See github.com/mlabs-haskell/styleguide

L-as (changes requested):
Waiting for the review of @kozross

Geometer1729:
The arithmetic terms don't seem to be hoist-able because they are not term level functions. `pgcd'`,`pmin` and `pmax` are all used just once. For `pnumerator` , `pdenominator` and `pfromInteger` I'm not sure if I should hoist them. They are all pretty simple.

srid:
I think we are also going to need a `PIsData` instance, to be able to store rationals in datum/redeemer.

L-as:
> pgcd',pmin and pmax are all used just once. For pnumerator , pdenominator and pfromInteger I'm not sure if I should hoist them. They are all pretty simple.

The rule of thumb is to hoist everything you can. If it's used only once, then it will be inlined by the compiler.

Plutarch/Rational.hs
```diff
@@ -1,28 +1,125 @@
...
+
+instance PEq PRational where
+  l #== r =
+    pmatch l $ \(PRational ln ld) ->
+      pmatch r $ \(PRational rn rd) ->
+        rd * ln #== rn * ld
+
+instance POrd PRational where
+  l #<= r =
+    pmatch l $ \(PRational ln ld) ->
```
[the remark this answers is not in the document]
L-as: @TotallyNotChase this should probably also be in the guide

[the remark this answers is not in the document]
L-as: #92

kozross (dismissed):
Looks good to me. Are there plans to add `round`, `truncate` and/or `properFraction` support?

L-as:
Needs to merge in upstream.

L-as:
> Looks good to me. Are there plans to add `round`, `truncate` and/or `properFraction` support?

If you want to add them, sure.

L-as (dismissed):
LGTM

Geometer1729:
@srid I tried to write a PIsData instance but I'm not really sure if this is right.

Geometer1729:
>

Geometer1729:
I'm sorry I clicked the wrong button  and closed this.

Geometer1729:
> Looks good to me. Are there plans to add `round`, `truncate` and/or `properFraction` support?

I kinda forgot about those. Looks like we do need them. I'll start working on that.

srid:
> @srid I tried to write a PIsData instance but I'm not really sure if this is right.

You can add a test to `examples/Examples/Rationals.hs` to verify that `pfromData . pdata` is essentially `id`, on a bunch of inputs (property tests would be better, but we don't use them yet).

Geometer1729:
@srid that test revealed a bug, I fixed it and now the tests pass, hopefully that was the only bug.
@kozross I implemented the rounding functions you mentioned.

Plutarch/Rational.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
```
L-as, line 1: ?

Plutarch/Rational.hs
```diff
@@ -11,12 +12,33 @@ module Plutarch.Rational (
...
 import Plutarch.Integer (PInteger, PIntegral (pdiv, pmod))
+import Plutarch.List (PListLike (pcons, phead, pnil, ptail), pmap)
 
 data PRational s = PRational (Term s PInteger) (Term s PInteger)
 
+instance PIsData PRational where
+  pfromData x = pListToRat #$ pmap # pasInt #$ pasList # pforgetData x
+  pdata x =
+    (punsafeCoerce :: Term _ (PAsData (PBuiltinList (PAsData PInteger))) -> Term _ (PAsData PRational)) $
+      pdata $ pmap # plam pdata #$ pRatToList # x
```
L-as: Why not just do `pdata $` instead of `plam pdata #$`?

L-as: Otherwise this looks fine actually :+1: , however, there is a design choice to be made here, between this approach and another approach. @TotallyNotChase

L-as (changes requested):

Plutarch/Rational.hs
L-as: I don't think we need the other pragma either.

L-as:
In general, I recommend not adding more features to PRs. You should just make a new PR.

Plutarch/Rational.hs
L-as: #56

L-as: I just noticed these aren't hoisted.

Plutarch/Rational.hs
kozross: This is a _really_ bad thing to enable in general.

Plutarch/Rational.hs
```diff
@@ -1,28 +1,235 @@
...
+pnumerator :: Term s (PRational :--> PInteger)
+pnumerator = phoistAcyclic $ plam $ \x -> pmatch x $ \(PRational n _) -> n
+
+pdenominator :: Term s (PRational :--> PInteger)
+pdenominator = phoistAcyclic $ plam $ \x -> pmatch x $ \(PRational _ d) -> d
+
+pfromInteger :: Term s (PInteger :--> PRational)
+pfromInteger = phoistAcyclic $ plam $ \n -> pcon $ PRational n 1
+
+pround :: Term s (PRational :--> PInteger)
```
kozross, line 218: Does this match the behaviour of `round` from both Plutus and `base`? Specifically:

``round x`` returns the nearest integer to ``x``; the even integer if ``x`` is equidistant between two integers

examples/Examples/Rationals.hs
```diff
@@ -0,0 +1,65 @@
...
+    , testCase "round 5/3" $
+        expect $ pround # (5 / 3) #== 2
+    , testCase "round 4/3" $
+        expect $ pround # (4 / 3) #== 1
+    , testCase "round 5/2" $
+        expect $ pround # (5 / 2) #== 3
+    , testCase "round 9/4" $
+        expect $ pround # (9 / 4) #== 2
+    , testCase "round 11/4" $
+        expect $ pround # (11 / 4) #== 3
```
kozross, line 39: We need to cover the 'equidistant' case described above. There are also no tests dealing with negative values.

examples/Examples/Rationals.hs
```diff
@@ -0,0 +1,65 @@
...
+    , testCase "truncate 7/4" $
+        expect $ ptruncate # (7 / 4) #== 1
+    , testCase "properFraction 11/7" $
+        expect $
+          pmatch (pproperFraction # (11 / 7)) $ \(PPair x y) ->
+            x #== 1 #&& y #== (4 / 7)
+    , testCase "properFraction 13/7" $
+        expect $
+          pmatch (pproperFraction # (13 / 7)) $ \(PPair x y) ->
+            x #== 1 #&& y #== (6 / 7)
```
kozross, line 53: Similarly here: we need to at least deal with negative values properly.

kozross (changes requested):

Plutarch/Rational.hs
Geometer1729: Sorry about that, I think the haskell-language-server plugin I'm trying out added those. I'm not sure why. I've removed both pragmas.

Plutarch/Rational.hs
Geometer1729: It does not I didn't realize that was the behavior of round. I'll fix that.

L-as:
Please retarget to staging BTW>

Geometer1729:
> In general, I recommend not adding more features to PRs. You should just make a new PR.

Sorry about that, I'll keep that in mind going forward.

Geometer1729:
I think I've made all of the requested changes now.

kozross (approved):

L-as:
This has merge conflicts now unfortunately.

L-as:
I think I solved it using GH's built-in functionality.

L-as:
Formatting is wrong FWIW. `./bin/format` needs to be run. CI is broken because staging branch is broken.

PR #90, opened 2022-01-12
Add section about using Haskell level functions

### LINKED ISSUE #78: General guide for Haskell-functions vs Plutarch-level functions
I've been thinking about this wrt. #76 

I somewhat incorrectly assumed that making the functions Haskell-level would be better, but now that I think about this more, this is much more complex. See e.g. https://github.com/Plutonomicon/plutarch/blob/4528d3beb5bbcd35be0b1f48a25a195e6abe3d7c/Plutarch/List.hs#L157

In the body of `pfoldr`, we mention the function `f`. If this were a Haskell-level function, it would be unconditionally inlined.
This is inobviously often problematic, because it will mean a different top-level function is generated for each function passed, i.e. it is **monomorphised**. We often do not want monomorphisation because we care about size more than CPU.

In general, functions should generally try to be as small as possible, and that is why e.g. `pelimList` should take Haskell-level functions: https://github.com/Plutonomicon/plutarch/blob/4528d3beb5bbcd35be0b1f48a25a195e6abe3d7c/Plutarch/List.hs#L92

Were they not Haskell-level, extra `plam`s would have to be introduced.

One possible optimisation that we don't perform today is inlining the function passed to a hoisted HOF if that HOF is inlined due to only being used once. This is not trivial because it will mean we have to perform analysis of the body of the HOF to find out whether it's worth it.

CC @TotallyNotChase

TotallyNotChase:
Closes #78

### DIFF
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -59,6 +59,7 @@
   - [Plutarch functions are strict](#plutarch-functions-are-strict)
   - [Don't duplicate work](#dont-duplicate-work)
   - [Prefer Plutarch level functions](#prefer-plutarch-level-functions)
+  - [When to use Haskell level functions?](#when-to-use-haskell-level-functions)
   - [Hoisting is great - but not a silver bullet](#hoisting-is-great---but-not-a-silver-bullet)
   - [The difference between `PlutusType`/`PCon` and `PLift`'s `pconstant`](#the-difference-between-plutustypepcon-and-plifts-pconstant)
 - [Common Issues](#common-issues)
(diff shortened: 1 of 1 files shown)


docs/GUIDE.md
```diff
@@ -1102,6 +1103,57 @@ por = phoistAcyclic $
...
+```hs
+pelimList :: InDefaultUni a => Term s (a :--> PBuiltinList a :--> r) -> Term s r -> Term s (PBuiltinList a) -> Term s r
+pelimList match_cons match_nil ls = pmatch ls $ \case
+  PCons x xs -> match_cons # x # xs
+  PNil -> match_nil
+```
+It takes in a Plutarch level function, let's see a typical usage-
+```hs
+pelimList
+  (plam $ \x xs -> pcons # x # (self # xs))
```
L-as, line 1123: You should hoist this

L-as: Actually, nvm, it doesn't matter here.

L-as (approved):
LGTM

Issue #94, opened 2022-01-12
Decouple lifting stuff from plutus-core internals


L-as:
We should depend on as little of Plutus internals as possible. The internal types are changing all the time and it's a headache to keep up with.

Issue #28, opened 2021-12-24
Stabilize module exports (more exports from prelude, remove unsafe exports from `Plutarch` etc)


TotallyNotChase:
Right now, Plutarch users need to import a lot of modules to be able to work with it. I suggest making the exports more rigid and uniform.

My suggestions:-
* `Plutarch.Prelude` should export all the common *Plutarch level* types as well as common utilities corresponding to said types. Just like Haskell prelude. This should ideally not include `POpaque`.
  
  Whether or not this should also export the eDSL types (`Term`, `ClosedTerm`, `PlutusType` etc) is up for discussion.

* `Plutarch` should export functions whose primary purpose is related to the eDSL - not a particular Plutarch type. e.g. `plam`, `plet`, `pdelay`, `pforce`, `phoistAcyclic`, `pinl` etc. It **should not** export any unsafe functions or functions that users are not expected to use.
* `Plutarch.Internal` or `Plutarch.Builtin.Internal` should export unsafe functions. This includes `popaque`.
* `Plutarch.Run` for exporting `compile` and `evaluateScript`.
* The *Plutarch level* types still get their own modules with their own function exports - this shouldn't have to change.

We can discuss the exact exports for each of these in this issue. In the end, Plutarch users should be able to get started on writing basic code just by importing `Plutarch.Prelude` and possibly `Plutarch`. 

P.S - I think this issue should only be worked on after we have settled the builtin list and builtin pair creation issues.

L-as:
`punsafe*` are meont to be used by the user. There are many cases where you need them to ensure good performance.

TotallyNotChase:
For sure. But I think once we have most of the builtin utilities and builtin type building synonyms implemented - there will be little reason to use them commonly. I'm not suggesting their removal from user land - I'm suggesting they be explicitly exported by some other module.

L-as:
Users should never imports anything from an `Internal` module.

PR #99, opened 2022-01-12
Staging

L-as:


L-as:
#94 needs to be fixed before this can be merged. Right now it's all broken because some new constraints have been added, not to mention `pdefaultUniConstant` is fundamentally broken.

Plutarch/Lift.hs
```diff
@@ -1,109 +1,70 @@
...
-  fromString = LiftError_Custom . T.pack
-
-{- | Class of Plutarch types `p` that can be converted to/from a Haskell type.
 
-The Haskell type is determined by `PHaskellType p`.
--}
-class PLift (p :: k -> Type) where
-  -- | The associated Haskell type for `p`
-  type PHaskellType p :: Type
+class (PConstant (PLifted p), PConstanted (PLifted p) ~ p) => PUnsafeLiftDecl (p :: PType) where
```
TotallyNotChase, line 19: Would an open type family make more sense here? It's just an empty class with an associated type after all.

TotallyNotChase:
What's the contextual difference between `PLift` and `PUnsafeLiftDecl`. They are both the same but when does one write one over the other?

Plutarch/Monadic.hs
```diff
@@ -0,0 +1,16 @@
+module Plutarch.Monadic ((>>=), (>>), fail) where
+
+import Prelude hiding (fail, (>>), (>>=))
+
+import Data.String (fromString)
+import Plutarch.Prelude
+import Plutarch.Trace (ptraceError)
+
+(>>=) :: (x -> Term s a) -> x -> Term s a
```
blamario, line 9: Some comments would be helpful here.

blamario (dismissed):
I skipped over Builtins, DataRepr, List, Rational modules, and the Nix stuff.

Plutarch/Lift.hs
```diff
@@ -1,109 +1,70 @@
...
-  -- This will fully evaluate the arbitrary closed expression, and convert the
-  -- resulting value.
-  -- -}
-  plift' :: ClosedTerm p -> Either LiftError (PHaskellType p)
+type PLift = PUnsafeLiftDecl
 
--- | Like `pconstant'` but TypeApplication-friendly
-pconstant :: forall p s. PLift p => PHaskellType p -> Term s p
-pconstant = pconstant'
+pconstant :: forall p s. PLift p => PLifted p -> Term s p
```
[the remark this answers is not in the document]
L-as: Some of the removed haddocks happened due to bugs in fourmolu 0.3, but we're using 0.5 now.

L-as:
> What's the contextual difference between `PLift` and `PUnsafeLiftDecl`. They are both the same but when does one write one over the other?

You use `PLift` unless you can't, which is when declaring the instance.

L-as:
BTW we should make a PR to fix the issues mentioned.

Plutarch/Internal.hs
```diff
@@ -104,6 +104,15 @@ mapTerm f (TermResult t d) = TermResult (f t) d
...
 mkTermRes :: RawTerm -> TermResult
 mkTermRes r = TermResult r []
+
+-- | Type of `s`.
+data S
+
+-- | Shorthand for Plutarch types.
+type PType = S -> Type
+
+type role Term phantom representational
```
TotallyNotChase, line 114: I think the lack of this type role allowed `Term s PPubKeyHash` to be coerced to `Term s PByteString`. `coerce` no longer allows this. The change makes sense though - as it'd allow other unsafe conversions as well.

But I'm curious as to how one would do the above without `punsafeCoerce` now?

L-as: We have `pto`, but `PlutusType` isn't implemented for most things, see #126.

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 # Revision history for plutarch
 
+# 1.1.0
+
+- Bunch of new functions.
```
srid, line 5: Can we list those functions here? Or at least the modules they were added in.

Plutarch/Internal.hs
```diff
@@ -104,6 +104,15 @@ mapTerm f (TermResult t d) = TermResult (f t) d
 
 mkTermRes :: RawTerm -> TermResult
 mkTermRes r = TermResult r []
+
+-- | Type of `s`.
+data S
```
srid, line 109: It maybe worth documenting this `S`. Why do we have a `s` in `Term s`?

Plutarch/Internal.hs
```diff
@@ -161,27 +170,59 @@ plam' f = Term $ \i ->
...
+    getArityBuiltin (RBuiltin PLC.BData) = Just 0
+    getArityBuiltin (RBuiltin PLC.UnConstrData) = Just 0
+    getArityBuiltin (RBuiltin PLC.UnMapData) = Just 0
+    getArityBuiltin (RBuiltin PLC.UnListData) = Just 0
+    getArityBuiltin (RBuiltin PLC.UnIData) = Just 0
+    getArityBuiltin (RBuiltin PLC.UnBData) = Just 0
+    getArityBuiltin (RBuiltin PLC.EqualsData) = Just 1
+    getArityBuiltin (RBuiltin PLC.MkPairData) = Just 1
+    getArityBuiltin (RBuiltin PLC.MkNilData) = Just 0
+    getArityBuiltin (RBuiltin PLC.MkNilPairData) = Just 0
```
srid, line 223: There probably is a way to have `getArityBuiltin` with new plutus (and even infer builtin function types from TypeScheme); I will look at it when I get a chance.

flake.nix
```diff
@@ -1,144 +1,427 @@
...
+  inputs.flat.url = "github:Quid2/flat?rev=41a040c413351e021982bb78bd00f750628f8060";
   inputs.flat.flake = false;
-  inputs.Win32-network.url = "github:input-output-hk/Win32-network?rev=3825d3abf75f83f406c1f7161883c438dac7277d";
+  # https://github.com/input-output-hk/Win32-network/pull/10
+  inputs.Win32-network.url = "github:input-output-hk/Win32-network?rev=2d1a01c7cbb9f68a1aefe2934aad6c70644ebfea";
   inputs.Win32-network.flake = false;
+  # https://github.com/haskell-foundation/foundation/pull/555
+  inputs.foundation.url = "github:haskell-foundation/foundation?rev=0bb195e1fea06d144dafc5af9a0ff79af0a5f4a0";
+  inputs.foundation.flake = false;
+  # https://github.com/locallycompact/protolude
```
srid, line 26: This comment seems outdated. We are using upstream repo in fact.

Plutarch/Lift.hs
TotallyNotChase: Ah, right... darn.

CHANGELOG.md
L-as: Sure, if somebody makes a PR we can merge it.

Plutarch/Internal.hs
L-as: The issue is that Plutus internals change. Do we really want to depend on the internals?

Plutarch/Internal.hs
TotallyNotChase: We could simply point to the part of the users guide explaining the `s`

CHANGELOG.md
TotallyNotChase: I was working on a detailed changelog - but it probably will have to wait

flake.nix
L-as: Which repository are you talking about?

Plutarch/Internal.hs
srid: We can stop relying on the internal functions, and use plutus' types/constraints only (like `TypeScheme`), which I don't imagine will change often.

Plutarch/Lift.hs
```diff
@@ -1,109 +1,70 @@
...
+        Nothing -> Left LiftError_FromRepr
+      Left e -> Left $ LiftError_EvalException e
+  Left e -> Left $ LiftError_ScriptError e
+
+plift :: forall p. (HasCallStack, PLift p) => ClosedTerm p -> (PLifted p)
+plift prog = case plift' prog of
+  Right x -> x
+  Left e -> error $ "plift failed: " <> show e
+
+newtype DerivePConstantViaCoercible (h :: Type) (p :: PType) (r :: Type) = DerivePConstantViaCoercible h
```
TotallyNotChase, line 56: Is the `h` used for `DerivePConstantViaCoercible`? Is there an example usage somewhere perhaps?

L-as: The `h` is used because it's derived *for* the `h`. `PInteger` uses it I think.

TotallyNotChase:
> BTW we should make a PR to fix the issues mentioned.

#137

Plutarch/Lift.hs
L-as: Indeed, albeit I still think it's useful to have the representation be separate.

Plutarch/Monadic.hs
TotallyNotChase: Added via #137

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,572 @@
...
+instance PUnsafeLiftDecl PRedeemer where type PLifted PRedeemer = Plutus.Redeemer
+deriving via (DerivePConstantViaNewtype Plutus.Redeemer PRedeemer PData) instance (PConstant Plutus.Redeemer)
+
+newtype PDatumHash (s :: S)
+  = PDatumHash (Term s PByteString)
+
+instance PUnsafeLiftDecl PDatumHash where type PLifted PDatumHash = Plutus.DatumHash
+deriving via (DerivePConstantViaNewtype Plutus.DatumHash PDatumHash PByteString) instance (PConstant Plutus.DatumHash)
+
+newtype PStakeValidatorHash (s :: S)
```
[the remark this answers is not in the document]
t1lde: A singular mechanism would be undesirable - you have to think about it in a way that ought not to be fully hidden. But some naming/syntactic cleanup could make it feel a bit nicer...

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,572 @@
...
+newtype PCurrencySymbol (s :: S)
+  = PCurrencySymbol (Term s PByteString)
+instance PUnsafeLiftDecl PCurrencySymbol where type PLifted PCurrencySymbol = Plutus.CurrencySymbol
+deriving via
+  (DerivePConstantViaNewtype Plutus.CurrencySymbol PCurrencySymbol PByteString)
+  instance
+    (PConstant Plutus.CurrencySymbol)
+
+newtype PValue (s :: S)
+  = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName PInteger)))
```
t1lde, line 241: Another important newtype lacking instances...

t1lde (approved):

Plutarch/Api/V1.hs
L-as: I believe @TotallyNotChase is working on this FWIW

TotallyNotChase: I haven't started generalizing what I already have yet. Will get to it later today unless someone else picks it up.

MatthewCroughan (commented):
This PR is 105 commits, it is too large for any individual to review. I cannot comment on this. Split the changes up into multiple PRs with specific changes, rather than a large refactor which globs in both Nix and Haskell code as one, and I can review it.

MatthewCroughan:
I've been informed by @L-as that there is a workflow occurring where everything is currently being merged into this Staging PR. I did not know about this, but we should write something down about this workflow in the README or elsewhere before others end up with the same confusion as me.

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,572 @@
...
+import qualified Plutus.V1.Ledger.Api as Plutus
+import qualified PlutusTx.Builtins.Internal as PT
+
+--------------------------------------------------------------------------------
+
+type PTuple = PDataList
+
+---------- V1 Specific types, Incompatible with V2
+
+newtype PTxInfo (s :: S)
```
TotallyNotChase, line 91: I just noticed this. I think we should keep these types as `data` rather than `newtype`. Similar to their ledger api variants.

It lessens the cognitive load when I can glance at a data type to know whether or not its runtime representation (i.e runtime as in Plutus core, not haskell runtime) is just its inner field. In the case of `TxInfo`, `ScriptContext` and similar - that is not the case, their constructors does indeed persist during runtime. It'd be helpful to *semantically* reflect that in Plutarch synonyms IMO

CC @L-as

TotallyNotChase (changes requested):

Plutarch/Api/V1.hs
L-as: Well, in the case of `TxInfo`, it *is* just its inner field.

TotallyNotChase: It's not though. `data TxInfo = TxInfo ...`. It's the constructor + the fields in the plutus core. Whereas the `newtype` types conventionally don't keep their constructors.

This distinction has helped me hover through plutus ledger types and understand what their plutus core representations are like. I've used the thumb rule: If it's a `newtype`, there is no extra `Constr` step - it's just the data representation of its singular field. If it's a `data`, there is a `Constr` (unless it's an integer, bytestring, list, or map) + the field(s). It'd be nice to keep this distinction in Plutarch synonyms - though it is semantically redundant.

examples/Examples/Api.hs
```diff
@@ -0,0 +1,157 @@
...
+_getInputs :: Term s (PTxInfo :--> PAsData (PBuiltinList (PAsData PTxInInfo)))
+_getInputs = plam $ \x -> P.do
+  PTxInfo i <- pmatch x
+  ptrace "xhuawdhauywhd"
+  i' <- plet i
+  pindexDataList (Proxy @0) # i'
+
+{-
+-- | Get first validator from TxInInfo
+getValidator :: Term s (PBuiltinList (PAsData PTxInInfo) :--> PAsData PValidatorHash)
```
L-as, line 121: I think we can add this back now.

Plutarch/Builtin.hs
```diff
@@ -17,32 +20,34 @@ module Plutarch.Builtin (
   PAsData,
   pforgetData,
   ppairDataBuiltin,
-  InDefaultUni,
```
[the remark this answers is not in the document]
TotallyNotChase: This has been completed via #140

Plutarch/Api/V1.hs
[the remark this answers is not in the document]
TotallyNotChase: This has been completed via #151

Plutarch/Api/V1.hs
TotallyNotChase: This has yet to be tackled, though #151 partially solves the lacking instances problem.

Plutarch/Api/V1.hs
t1lde: That's a good point and the distinction has actually tripped me up somewhat, too

L-as: It's a newtype of `PDataList`. `PDataList` itself is a `Constr`.

TotallyNotChase: Sure, but `PDataList` is very much a Plutarch concept. Keeping the plutus `data` types as `data`, and `newtype` types as `newtype` makes it immediately obvious IMO. I don't think it costs us anything either.

examples/Examples/Api.hs
TotallyNotChase: Partially resolved by #158

Plutarch/Api/V1.hs
L-as: It does, it makes it impossible to require that the contents match something.

L-as: (Through Coercible)

srid: > It does, it makes it impossible to require that the contents match something.
> 
> (Through Coercible)

It can be done using `generics-sop` (something like `Code (a s) ~ '[ '[ PDataList XS ] ]`, but it will require that `SI` type I mentioned in Discord.

TotallyNotChase: > It does, it makes it impossible to require that the contents match something.
> 
> (Through Coercible)

Ugh, that's unfortunate.

examples/Examples/Api.hs
TotallyNotChase: Further progress by #161 
Possibly fully resolved by #159

Plutarch/Api/V1.hs
TotallyNotChase: `PValue` now has both `PlutusType` and `PIsData` instances through #161 

It's currently missing `PConstant`/`PLift` instances (need those for `PMap` too). I assume this might also be resolved by #159

Plutarch.hs
```diff
@@ -22,9 +23,8 @@ module Plutarch (
   PI.punsafeConstant,
   PI.Term,
   PI.TermCont (..),
-  PL.plift,
```
TotallyNotChase, line 25: It looks like `plift` and `pconstant` exports have been removed from the `Plutarch` module. Should probably be added back.

TotallyNotChase (changes requested):

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,658 @@
...
+  pconstantToRepr m = (\(x, y) -> (Plutus.toData x, Plutus.toData y)) <$> PlutusMap.toList m
+  pconstantFromRepr m = fmap PlutusMap.fromList $
+    flip traverse m $ \(x, y) -> do
+      x' <- Plutus.fromData x
+      y' <- Plutus.fromData y
+      Just (x', y')
+
+---------- Others
+
+data PMaybe a (s :: S)
```
TotallyNotChase, line 624: These names conflict with those of the scott encoded versions. Have we got a policy for this yet? Perhaps these should be moved out to a separate module or explicitly renamed? It's a bit of a pain to import both `Plutarch.Api.V1` and `Plutarch.Maybe` at the same time during dev.

L-as: That's also what I'm thinking.

Plutarch.hs
L-as: I'm pretty sure this was a mistake in the first place, since it makes no sense. They should be re-exported from `Plutarch.Prelude`, not `Plutarch`.

TotallyNotChase: Sounds good as well. Maybe this should be tackled alongside the whole "add more exports to prelude" thing. Related: #28

Plutarch/Lift.hs
L-as: I don't think this is important enough to change.

Plutarch/Api/V1.hs
L-as: We need to revisit this.

srid: cf. https://github.com/Plutonomicon/plutarch/issues/106#issuecomment-1012103112 

and https://github.com/Plutonomicon/plutarch/issues/106#issuecomment-1013019797

Plutarch/Prelude.hs
```diff
@@ -12,14 +14,139 @@ module Plutarch.Prelude (
...
   plet,
   pinl,
-  pcon,
-  pmatch,
   pto,
   pfix,
   Type,
+  S,
+  PType,
+  PlutusType (PInner, pcon', pmatch'),
```
TotallyNotChase, line 22: I wonder if all the members of `PlutusType` are necessary now that we have generic derivers for it.

Plutarch/Builtin.hs
```diff
@@ -1,4 +1,9 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 -- This should have been called Plutarch.Data...
```
TotallyNotChase, line 8: ?

TotallyNotChase (approved):

Plutarch/Prelude.hs
L-as: Yeah, they're not necessary anymore probably. Another good example: I don't think we even *need* to make `pmatchRepr` public.

Plutarch/Builtin.hs
L-as: It probably makes sense to rename this. Where should we put `PBuiltinList`?

TotallyNotChase: `Plutarch.Builtin` seems fine and correct to me. The comment seems outdated.

Plutarch/Prelude.hs
TotallyNotChase: Well, there might be some value in letting users manually implement `PIsDataRepr` if they so choose. Though I'm not sure.

On a related note, other than `pmatchRepr` and stuff, do you think all the other exports from `Plutarch.DataRepr` are necessary?

Plutarch/Api/V1.hs
```diff
@@ -0,0 +1,549 @@
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Plutarch.Api.V1 (
```
t1lde, line 4: we ought to split this module, I'll go ahead and do this

t1lde (approved):
It's great seeing how this has all come together, props to everyone :smile: 

I'll go ahead and split up the API module rq

Plutarch/Api/V1.hs
t1lde: done on #205

TotallyNotChase (approved):
It's time!

blamario (approved):

PR #101, opened 2022-01-12
Initialize benchmarks

srid:
[Upstreaming the bulk of benchmarks code from Pluton]

Runs benchmarks in CI, creating `bench.csv` for further processing, also dumping a table of the same - so that we can look at the benchmark numbers for any commit. Useful for assessing the performance impact of PRs.

This is a draft, because I'd like to run benchmarks on the examples, but they are in cabal executable. Need to figure out what to do here.


srid:
Uhh, what happened to Git history?

TotallyNotChase:
> Uhh, what happened to Git history?

Confused, as well.

PR #108, opened 2022-01-13
Mario/records

blamario:
Added `instance PlutusType` and a couple more examples.

### DIFF
--- a/Plutarch/Rec.hs
+++ b/Plutarch/Rec.hs
@@ -12,7 +12,7 @@ import Data.Functor.Compose (Compose)
 import Data.Kind (Type)
 import Data.Monoid (Dual (Dual, getDual), Endo (Endo, appEndo), Sum (Sum, getSum))
 import Numeric.Natural (Natural)
-import Plutarch (PCon, pcon, phoistAcyclic, plam, punsafeCoerce, (#), (:-->))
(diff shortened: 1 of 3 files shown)


L-as:
You should rebase ontop of staging

blamario:
staging is still broken

blamario:
By which I mean that tests fail after the `Tracing` test. However I've just restored the LetRec tests; the issue was the GHC upgrade, but only insofar the TH splices needed to be repositioned, nothing to do with kinds.

L-as:
It shouldn't be broken?

blamario:
My `cabal test` output on `staging` ends with

        PAsData equality:                                                      OK
        Tracing:                                                               OK
        examples: <stdout>: commitBuffer: invalid argument (invalid character)

blamario:
Any idea what could this problem be? I have submitted PR #114 on top of staging and I can continue working there by disabling the tests that fail for me, but it would be nice to have a solid base to work on. I.e. either master or properly-behaved staging.

blamario:
It's something to do with the locale, in that Greek letters break the test output, but I don't see why the same output works on master.

blamario:
export LC_ALL=C.UTF-8
fixed the problem. The output didn't work under `en_CA.UTF-8` nor `en_US.UTF-8`, I guess Greek letters are incomprehensible in North America.

PR #109, opened 2022-01-13
Rework `PLift`, `asDefaultUni`, etc. and make examples compile

L-as:
Summmary:
- `PLift` is ready for `h -> p` functional dependency, but we need to replace `k` with `S` or similar first.
- `letrec` is broken for some reason, haven't looked into it.
- Removed `PVia` unfortunately, will readd.
- Removed `AsDefaultUni` stuff, it duplicated the functionality already
  provided by `PLift`, in addition to being unsound. To make an empty list for
  any type `a :: k -> Type`, we need to be able to create a constant for it,
  hence `PLift`.
- `PLift` instances for `PAsData` were removed. It messes with functional dependencies.


L-as:
This isn't really ready for master, but by merging into staging I hope others can build on top of it, especially #108

blamario:
> letrec is broken for some reason, haven't looked into it.

I don't see anything that could have broken it except, sadly, the GHC upgrade. Perhaps fixing `k` to `S` will un-break it.

Plutarch/Api/V1.hs
```diff
@@ -71,24 +69,21 @@ import Plutarch.DataRepr (
...
-import Plutarch.Tuple (PTuple)
 import qualified Plutus.V1.Ledger.Api as Plutus
-import qualified Plutus.V1.Ledger.Crypto as Plutus
-import qualified PlutusTx.AssocMap as AssocMap
-import PlutusTx.Builtins.Internal (BuiltinByteString (..))
 import qualified PlutusTx.Prelude as PlutusTx
 
 --------------------------------------------------------------------------------
 
+type PTuple = PDataList
```
t1lde, line 85: Good point, but this does remove lifting to Haskell tuples as a side-effect :thinking:

t1lde (approved):

L-as:
We could implement lifting for PDataList

Plutarch/Lift.hs
```diff
@@ -1,147 +1,69 @@
...
+
+plift' :: forall p h. PUnsafeLiftDecl h p => ClosedTerm p -> Either LiftError h
+plift' prog = case evaluateScript (compile prog) of
+  Right (_, _, Scripts.unScript -> UPLC.Program _ _ term) ->
+    case readKnownConstant @_ @(PLiftedRepr p) @(MachineError PLC.DefaultFun) Nothing term of
+      Right r -> case pliftFromRepr @_ @h @p r of
+        Just h -> Right h
+        Nothing -> Left LiftError
+      Left _ -> Left LiftError
+  Left _ -> Left LiftError
```
srid, line 51: Why are we dropping error messages here (and elsewhere)?

L-as: It was simply because I wanted to get staging to compile again ASAP.

L-as: These will definitely be added back before merging into main, hence the FIXME.

PR #110, opened 2022-01-14
instance PIsData PBool (compatible with Plutus API)

t1lde:
Originally branched from #109, as I thought this might involve those changes. Merge after #109.

### DIFF
--- a/Plutarch/Builtin.hs
+++ b/Plutarch/Builtin.hs
@@ -199,29 +199,22 @@ instance PIsData PByteString where
 -}
 instance PIsData PBool where
   pfromData x =
-    (phoistAcyclic $ plam $ \d -> toBool #$ tag # d)
-      # pforgetData x
+    (phoistAcyclic $ plam toBool) # pforgetData x
     where
-      toBool :: Term s (PInteger :--> PBool)
-      toBool = phoistAcyclic $ plam (#== 1)
-
-      tag :: Term s (PData :--> PInteger)
-      tag = phoistAcyclic $ plam $ \d -> pfstBuiltin #$ pasConstr # d
+      toBool :: Term s PData -> Term s PBool
+      toBool d = pfstBuiltin # (pasConstr # d) #== 1
 
   pdata x =
-    (phoistAcyclic $ plam $ \b -> constr #$ toInt # b)
-      # x
+    (phoistAcyclic $ plam toData) # x
     where
-      toInt :: Term s (PBool :--> PInteger)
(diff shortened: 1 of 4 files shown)


Plutarch/Builtin.hs
```diff
@@ -207,102 +184,40 @@ instance PIsData PByteString where
...
   pdata x = punsafeBuiltin PLC.BData # x
 
+{- |
+  Instance for PBool following the Plutus IsData repr
+  given by @makeIsDataIndexed ''Bool [('False,0),('True,1)]@,
+  which is used in 'TxInfo' via 'Closure'.
+-}
+instance PIsData PBool where
+  pfromData x =
+    (phoistAcyclic $ plam $ \d -> toBool #$ tag # d)
```
TotallyNotChase, line 194: I'm not sure if @L-as already has optimizations for this, but if not - it may be beneficial to use a more direct application here-
```hs
phoistAcyclic $ plam $ \d -> pfstBuiltin # (pasConstr # d) #== 1
```
Same for the other method impl as well.

TotallyNotChase (changes requested):

L-as:
Any chance this can be rebased? GItHub is showing a weird diff.

Plutarch/Builtin.hs
L-as: This is indeed better.

examples/Main.hs
```diff
@@ -14,7 +14,7 @@ import Examples.Tracing (traceTests)
 import Plutarch (POpaque, pconstant, plift', popaque, printTerm, punsafeBuiltin)
 import Plutarch.Api.V1 (PScriptPurpose (PMinting))
 import Plutarch.Bool (PBool (PFalse, PTrue), pand, pif, pnot, por, (#&&), (#<), (#<=), (#==), (#||))
-import Plutarch.Builtin (PBuiltinList (..), PBuiltinPair, PData, pdata)
+import Plutarch.Builtin (PBuiltinList (..), PBuiltinPair, PData, PIsData (..), pdata)
```
L-as, line 17: Please no .. imports.

L-as (changes requested):

Plutarch/Builtin.hs
t1lde: Not 100% sure what you meant for the other method but should be better now

t1lde:
Added the better method, and some more tests for `PIsData` instances

L-as (approved):
LGTM

PR #115, opened 2022-01-14
`s :: S`

L-as:
To migrate your code, a good first step is:

```sh
find -type f -name '*.hs' ! -path '*/dist-newstyle/*' -exec sed -i -e 's/k -> Type/PType/g' -e 's/s :: k/s :: S/g' {} +
```

If you're using type applications, some functions will now need fewer type applications.

### DIFF
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -22,6 +22,8 @@ module Plutarch (
   PI.punsafeConstant,
   PI.Term,
   PI.TermCont (..),
+  PI.S,
+  PI.PType,
   PlutusType (..),
   printTerm,
   printScript,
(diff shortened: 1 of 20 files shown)


L-as:
I should have made this into 3 PRs.

Plutarch.hs
```diff
@@ -22,9 +22,6 @@ module Plutarch (
   PI.punsafeConstant,
   PI.Term,
   PI.TermCont (..),
-  PL.plift,
```
L-as, line 25: I did this because I don't see why 2 modules should export the same function. These should arguably be in the Prelude, but I saved that for another PR.

srid (approved):
@Geometer1729 would appreciate this.

PR #118, opened 2022-01-14
PLift bidirectionality

L-as:
This makes `pconstant True` be inferred to be of type `Term s PBool`.

There is however an issue: deriving via seems to work quite suboptimally.

We need two derives because there are two type classes now:
```haskell
data PBool (s :: S) = PTrue | PFalse
  deriving (PUnsafeLiftDecl) via (DerivePLiftViaCoercible Bool PBool Bool)

deriving via (DerivePConstant Bool) PBool instance (PConstant Bool)
```

The issue then is that the instance is:
```haskell
instance (PConstant h, PConstanted h ~ DerivePLiftViaCoercible h p r, Coercible h r, PLC.DefaultUni `PLC.Includes` r) => PUnsafeLiftDecl (DerivePLiftViaCoercible h p r) where
```
`PConstanted h ~ DerivePLiftViaCoercible h p r` fails to hold, because `PConstanted h` is set by another
derive. I'm not sure whether this is easy to fix.

The reason I switched away from functional dependencies is ironically that it seemed
easier to make newtype derives work.

Is there any solution that doesn't involve TH?

### DIFF
--- a/Plutarch/Api/V1.hs
+++ b/Plutarch/Api/V1.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Plutarch.Api.V1 (
   -- * V1 Specific types
@@ -60,7 +61,7 @@ module Plutarch.Api.V1 (
 --------------------------------------------------------------------------------
 
 import Plutarch (PMatch)
-import Plutarch.Bool (PBool, PEq, POrd)
+import Plutarch.Bool (PBool)
 import Plutarch.Builtin (PAsData, PBuiltinList, PData, PIsData, type PBuiltinMap)
 import Plutarch.ByteString (PByteString)
 import Plutarch.DataRepr (
@@ -73,12 +74,12 @@ import Plutarch.DataRepr (
   pmatchRepr,
  )
 import Plutarch.Integer (PInteger, PIntegral)
-import Plutarch.Lift (PLifted, PUnsafeLiftDecl)
+import Plutarch.Lift (DerivePConstantViaData (DerivePConstantViaData), DerivePConstantViaNewtype (DerivePConstantViaNewtype), PConstant, PLifted, PUnsafeLiftDecl)
 
 -- ctor in-scope for deriving
 import Plutarch.Prelude
 import qualified Plutus.V1.Ledger.Api as Plutus
-import qualified PlutusTx.Prelude as PlutusTx
+import qualified PlutusTx.Builtins.Internal as PT
 
 --------------------------------------------------------------------------------
 
@@ -105,8 +106,11 @@ newtype PTxInfo (s :: S)
           )
       )
   deriving
-    (PMatch, PIsData, PUnsafeLiftDecl Plutus.TxInfo)
-    via PIsDataReprInstances PTxInfo Plutus.TxInfo
+    (PMatch, PIsData)
+    via PIsDataReprInstances PTxInfo
+
+instance PUnsafeLiftDecl PTxInfo where type PLifted PTxInfo = Plutus.TxInfo
+deriving via (DerivePConstantViaData Plutus.TxInfo PTxInfo) instance (PConstant Plutus.TxInfo)
 
 instance PIsDataRepr PTxInfo where
   type
@@ -129,8 +133,11 @@ instance PIsDataRepr PTxInfo where
 newtype PScriptContext (s :: S)
   = PScriptContext (Term s (PDataList '[PTxInfo, PScriptPurpose]))
   deriving
-    (PMatch, PIsData, PUnsafeLiftDecl Plutus.ScriptContext)
-    via PIsDataReprInstances PScriptContext Plutus.ScriptContext
+    (PMatch, PIsData)
+    via PIsDataReprInstances PScriptContext
+
+instance PUnsafeLiftDecl PScriptContext where type PLifted PScriptContext = Plutus.ScriptContext
+deriving via (DerivePConstantViaData Plutus.ScriptContext PScriptContext) instance (PConstant Plutus.ScriptContext)
 
 instance PIsDataRepr PScriptContext where
   type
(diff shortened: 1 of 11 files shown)


L-as:
I preferably don't want to give up on this. If we do, we can just remove `PConstant`.

L-as:
```
Plutarch/Bool.hs:25:13: error:
    • Couldn't match type ‘PBool’
                     with ‘DerivePLiftViaCoercible Bool PBool Bool’
        arising from the 'deriving' clause of a data type declaration
    • When deriving the instance for (PUnsafeLiftDecl PBool)
   |
25 |   deriving (PUnsafeLiftDecl) via (DerivePLiftViaCoercible Bool PBool Bool)
   |             ^^^^^^^^^^^^^^^
```

Plutarch/Lift.hs
```diff
@@ -16,41 +16,46 @@ import PlutusCore.Constant (readKnownConstant)
...
+class (PLift (PConstanted h), PLifted (PConstanted h) ~ h) => PConstant (h :: Type) where
+  type PConstanted h :: PType
+
+type PLift = PUnsafeLiftDecl
 
 newtype DerivePLiftViaCoercible (h :: Type) (p :: PType) (r :: Type) (s :: S) = DerivePLiftViaCoercible (p s)
+newtype DerivePConstant (h :: Type) (p :: PType) = DerivePConstant h
 
-instance (Coercible h r, PLC.DefaultUni `PLC.Includes` r) => PUnsafeLiftDecl h (DerivePLiftViaCoercible h p r) where
+instance (PConstant h, PConstanted h ~ DerivePLiftViaCoercible h p r, Coercible h r, PLC.DefaultUni `PLC.Includes` r) => PUnsafeLiftDecl (DerivePLiftViaCoercible h p r) where
```
t1lde, line 33: hmm so the issue here comes from the `PConstanted (PLifted p) ~ p` constraint not working if the type instances are defined in the way we want for the DerivingVia instance...

If the constraint isn't necessary for another instance, maybe they could be moved to `PLift` ?

L-as: Yep, that's the solution I've thought about too. I don't think the constraint is necessary for `PConstant`, so I'll try that.

L-as: I fixed this and will push my finalised changes later.

L-as:
I noticed that GHC hadn't inferred a correct type role, added the annotation, and noticed we had some bogus derivations for many of the types in the V1 module. Some of the derivations were sound, but they couldn't be derived using that mechanism anymore, so we'll have to add that back somehow.

Plutarch/Api/V1.hs
```diff
@@ -421,8 +433,11 @@ instance PIsDataRepr PTxId where
 newtype PTxOutRef (s :: S)
   = PTxOutRef (Term s (PDataList '[PTxId, PInteger]))
   deriving
-    (PMatch, PIsData, PUnsafeLiftDecl Plutus.TxOutRef)
-    via PIsDataReprInstances PTxOutRef Plutus.TxOutRef
+    (PMatch, PIsData)
+    via PIsDataReprInstances PTxOutRef
+
+instance PUnsafeLiftDecl PStakeValidatorHash where type PLifted PStakeValidatorHash = Plutus.StakeValidatorHash
+deriving via (DerivePConstantViaData Plutus.StakeValidatorHash PStakeValidatorHash) instance (PConstant Plutus.StakeValidatorHash)
```
t1lde, line 440: Looks like some of these instance decls got pasted in the wrong part of the file :sweat_smile:

Plutarch/Builtin.hs
```diff
@@ -35,14 +36,17 @@ import PlutusTx (Data)
...
-instance (PUnsafeLiftDecl ah a, PUnsafeLiftDecl bh b) => PUnsafeLiftDecl (ah, bh) (PBuiltinPair a b) where
-  type PLiftedRepr (PBuiltinPair a b) = (PLiftedRepr a, PLiftedRepr b)
+instance (PLift a, PLift b) => PUnsafeLiftDecl (PBuiltinPair a b) where
   type PLifted (PBuiltinPair a b) = (PLifted a, PLifted b)
-  pliftToRepr (x, y) = (pliftToRepr @_ @a x, pliftToRepr @_ @b y)
-  pliftFromRepr (x, y) = do
-    x' <- pliftFromRepr @_ @a x
-    y' <- pliftFromRepr @_ @b y
+
+-- FIXME: figure out good way of deriving this
```
t1lde, line 42: it would be acceptable if we didn't tbh

Plutarch/Builtin.hs
```diff
@@ -154,11 +163,13 @@ data PAsData (a :: PType) (s :: S)
...
-  type PLiftedRepr (PAsData a) = Data
-  pliftToRepr = \case {}
-  pliftFromRepr _ = Nothing
+instance PConstant (PAsDataLifted a) where
+  type PConstantRepr (PAsDataLifted a) = Data
+  type PConstanted (PAsDataLifted a) = PAsData a
+  pconstantToRepr = \case {}
+  pconstantFromRepr _ = Nothing
+
+instance PUnsafeLiftDecl (PAsData a) where type PLifted (PAsData a) = PAsDataLifted a
```
t1lde, line 172: doesn't this mean we can't actually use `pconstant`? Or is that deliberate...

L-as:
I'm merging this as I need to work on top of this.

Plutarch/Lift.hs
```diff
@@ -3,67 +3,77 @@
...
+  pconstantToRepr = coerce
+  pconstantFromRepr = Just . coerce
+
+newtype DerivePConstantViaData (h :: Type) (p :: PType) = DerivePConstantViaData h
+
+instance (PLift p, Ledger.FromData h, Ledger.ToData h) => PConstant (DerivePConstantViaData h p) where
+  type PConstantRepr (DerivePConstantViaData h p) = Ledger.Data
+  type PConstanted (DerivePConstantViaData h p) = p
+  pconstantToRepr (DerivePConstantViaData x) = Ledger.toData x
+  pconstantFromRepr x = DerivePConstantViaData <$> Ledger.fromData x
```
t1lde, line 71: I'm not sure this leads to the correct handling under `PAsData`, I think it's only correct when `pfromData` and `pdata` are both id

Plutarch/Api/V1.hs
L-as: lol

L-as:
I undid the merge

Plutarch/Builtin.hs
L-as: No, you can, it's just not obvious. You can do e.g. `pconstant []` for a list of this, but you can't make a non-empty list. This is a slight but important difference, because it means you can make `pnil` work.

L-as: The instance should probably be overlappable.

Plutarch/Lift.hs
L-as: This isn't supposed to work with `PAsData` anyway, nor did it work with it.

t1lde: not sure if I've misunderstood the implications but I think it would mean a `Term s PFoo` from the result of `pfromData` vs from `pconstant` would be represented differently.

PR #119, opened 2022-01-15
do-syntax

L-as:
```haskell
import qualified Plutarch.Monadic as P

P.do
  PTxInfo y <- pmatch (x :: Term s PTxInfo)
  PTrue <- pmatch $ pcon PFalse -- calls `P.fail errormsg` if there's not a match, which is `ptrace errormsg perror`.
  z <- plet $ x + y
  ptrace "ahduawhdw"
  perror
```

### DIFF
--- a/Plutarch/Monadic.hs
+++ b/Plutarch/Monadic.hs
@@ -1,14 +1,16 @@
-module Plutarch.Monadic ((>>=), fail) where
+module Plutarch.Monadic ((>>=), (>>), fail) where
 
-import Prelude hiding (fail, (>>=))
+import Prelude hiding (fail, (>>), (>>=))
 
 import Data.String (fromString)
-import Plutarch (PMatch)
 import Plutarch.Prelude
 import Plutarch.Trace (ptrace)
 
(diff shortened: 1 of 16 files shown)


L-as:
I like this syntax.

Plutarch/Monadic.hs
```diff
@@ -0,0 +1,16 @@
...
+import Plutarch.Trace (ptrace)
+
+(>>=) :: a -> a
+(>>=) = id
+
+(>>) :: a -> a
+(>>) = id
+
+fail :: String -> Term s a
+fail msg = ptrace (fromString msg) perror
```
L-as, line 16: ptraceError

L-as (commented):
I need to change this.

t1lde (commented):
Is this back-compatible with ghc<9? I don't think we want to force users into ghc9 (yet)

L-as:
I've tried not to do anything that will cause the base library to fail to compile with GHC 8.10, albeit the tests won't work obviously.

t1lde:
should be fine, plutus itself is probably likely to start requiring ghc9 anyway

t1lde (approved):

PR #129, opened 2022-01-18
Plutarch Natural (basic sketch)

sergesku:
This is the starting point for implementing numerical stuff on top of Plutarch.

I'm not sure if we need to import the `plutus-numeric` modules to implement numeric classes such as `AdditiveHemigroup`, `IntegralDomain` etc.

### DIFF
--- a/Plutarch/Natural.hs
+++ b/Plutarch/Natural.hs
@@ -1,0 +1,119 @@
+module Plutarch.Natural (
+  PNatural,
+  pnatToInt,
+  pnatFromInt,
+  pmonusNat,
+) where
+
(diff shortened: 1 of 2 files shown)


L-as:
We definitely don't want to depend on `plutus-numeric`.

L-as:
Do we want a `plutarch-extra` now @kozross or what do you think?

kozross:
We could, but I'd rather fractionate into multiple separate libraries, rather than a megarepo with sublibraries.

L-as:
#139

sergesku:
This stuff will be combined with the new Numeric library

PR #133, opened 2022-01-18
Make depending on for downstream users easier

L-as:


TotallyNotChase:
Could you please also add some information about the dependency adding to the readme as well? or is that for another PR?

L-as:
I'll do that in another PR.

PR #137, opened 2022-01-18
Basic scattered cleanup

TotallyNotChase:
Sorry for jumbling a bunch of cleanup into a singular commit. I generally try to avoid that but oh well. This takes care of a bunch of the review comments from #99 - mostly just the basic stuff though.


L-as (approved):
LGTM

Plutarch/Internal.hs
```diff
@@ -105,7 +107,7 @@ mapTerm f (TermResult t d) = TermResult (f t) d
 mkTermRes :: RawTerm -> TermResult
 mkTermRes r = TermResult r []
 
--- | Type of `s`.
+-- | Type of `s` in `Term s a`. See: "What is the `s`?" section on the Plutarch guide.
```
blamario, line 110: on -> in
Do we have a stable link for the Plutarch guide?

TotallyNotChase: I suppose we should consider the one on `master` stable until we set up a website

PR #140, opened 2022-01-19
Update guide

### LINKED ISSUE #122: do-syntax in docs

TotallyNotChase:
* Add `do` syntax usage guide
* Update `PLift` section, added `PConstant`
* Remove `InDefaultUni` information, replace with `PLift`.

Closes #122 
Resolves a review comment in #99

### DIFF
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -35,7 +35,7 @@
       - [Implementing `PLift`](#implementing-plift)
     - [PlutusType, PCon, and PMatch](#plutustype-pcon-and-pmatch)
     - [PListLike](#plistlike)
-    - [PIsDataRepr & PDataList](#pisdatarepr--pdatalist)
+    - [PIsDataRepr](#pisdatarepr)
       - [Implementing PIsDataRepr](#implementing-pisdatarepr)
   - [Working with Types](#working-with-types)
     - [PInteger](#pinteger)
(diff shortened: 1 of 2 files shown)


docs/GUIDE.md
```diff
@@ -560,24 +589,86 @@ purp = pconstant $ Minting "be"
...
-data PBool s
-  deriving (PLift) via PBuiltinType PBool Bool
+deriving via (DerivePConstantViaCoercible Integer PInteger Integer) instance (PConstant Integer)
 ```
-In this case, `PBool` is represented as a builtin boolean in Plutus Core, as [`Bool`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Bool.html#t:Bool) corresponds to builtin boolean.
-This lets you do `pconstant True` to obtain a `Term s PBool`. And `plift $ pconstant True` to obtain a `Bool`.
+`DerivePConstantViaCoercible` takes in 3 type parameters-
+* The Haskell type itself, for which `PConstant` is being implemented for.
+* The **direct** Plutarch synonym to the Haskell type.
+* Another Haskell type that the first param can be coerced into. TODO: more info
```
TotallyNotChase, line 636: @L-as I put a `TODO` here for more information. We should put some more information here before merging. I'm not entirely sure about this 3rd param's significance yet - so feel free to tell me what to put in here.

L-as: Now that I think about it basically all uses of it would instead use `ViaNewtype`.

TotallyNotChase: Alright, updated.

TotallyNotChase: > Now that I think about it basically all uses of it would instead use `ViaNewtype`.
Yeah.....

PR #143, opened 2022-01-19
Add support for using tuples with `plam`

### LINKED ISSUE #2: Improve type inference of plam

L-as:
Fixes #2

`plam $ \(x, y) -> perror`

### DIFF
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -125,6 +125,21 @@ instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ T
 instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d, e' ~ Term s e, f' ~ Term s f, g' ~ Term s g) => PLamN (a' -> b' -> c' -> d' -> e' -> f' -> g') (Term s (a :--> b :--> c :--> d :--> e :--> f :--> g)) where
   plam f = plam' $ \x -> plam (f x)
 
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c) => PLamN ((a', b') -> c') (Term s (a :--> b :--> c)) where
+  plam f = plam $ \x y -> f (x, y)
+
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d) => PLamN ((a', b', c') -> d') (Term s (a :--> b :--> c :--> d)) where
+  plam f = plam $ \x y z -> f (x, y, z)
+
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d, e' ~ Term s e) => PLamN ((a', b', c', d') -> e') (Term s (a :--> b :--> c :--> d :--> e)) where
+  plam f = plam $ \x y z w -> f (x, y, z, w)
+
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d, e' ~ Term s e, f' ~ Term s f) => PLamN ((a', b', c', d', e') -> f') (Term s (a :--> b :--> c :--> d :--> e :--> f)) where
+  plam f = plam $ \x y z w w' -> f (x, y, z, w, w')
+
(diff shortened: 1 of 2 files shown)


L-as:
I'm not sure this is what we want to go for, but it's a possible solution.

blamario (commented):
I'm not a fan to be honest. This doesn't *remove* any of the existing instances, so users are still open to the same confusing type errors as before when they write  `plam $ \a b c-> ...`. They have to know or find out what the problem is and then rewrite to `plam $ \(a, b, c)->`. But if they recognize the problem, they can already rewrite to `plam' $ \a-> plam' $ \b-> plam' $ \c-> ...`.

I think the root cause of the inference failure today was that the second of the variables was unused in the body of the lambda, so GHC had no clue it was supposed to be a `Term`. In that case, again, it would have been easy to fix the problem once we realized the cause.

srid:
Pretty much agree with @blamario re: confusion. /cc @Geometer1729 for opinion.

Geometer1729:
I share the concern that without knowing to do this it doesn't help much. Could we get a linter to suggest this?

L-as:
The point is that we'd recommend this syntax as default.

L-as:
We'd deprecate the old syntax.

TotallyNotChase:
Maybe it'd be better to have a separate `plamCurry` for this. I don't think overloading the behavior of `plam` is a good idea either. We could remove the old syntax - but why? why not just have a separate function?

L-as:
Because plam would still be confusing it people.

L-as:
The alternative to this is someone fixing the inference for the original instances, but I'm not sure that's possible.

L-as:
To give some context:
Yesterday a mob programming session internally at MLabs with quite a lot of people were stuck for quite some time due to a simple inference issue stemming from the following piece of code:
```haskell
bad :: Term s (a :--> b :--> c)
bad = plam $ \x y -> plet x $ \_ -> perror

good :: Term s (a :--> b :--> c)
good = plam $ \(x, y) -> plet x $ \_ -> perror

good' :: Term s (a :--> b :--> c)
good' = plam $ \x (y :: Term _ _) -> plet x $ \_ -> perror
```

There is another instance of confusing error messages in Plutarch's code base itself:
```haskell
instance PlutusType (PMaybe a) where
  type PInner (PMaybe a) b = (a :--> b) :--> PDelayed b :--> b
  pcon' :: forall s. PMaybe a s -> forall b. Term s (PInner (PMaybe a) b)
  pcon' (PJust x) = plam $ \f (_ :: Term _ _) -> f # x
  pcon' PNothing = plam $ \_ g -> pforce g
  pmatch' x f = x # plam (f . PJust) # pdelay (f PNothing)
```

By replacing `f (_ :: Term _ _)` with `(f, _)`, type inference starts working again.

If there is a way to fix this problem without changing the syntax, I am all for it, but at the current point in time,
it is confusing to the point that ~15 people could not figure out what is happening until I came in and pointed
out the fact that it was likely because type inference failed on `plam`.

If we *do* switch, it would a period of deprecating the old syntax (and making sure everyone uses tuples, converting all existing code to tuples, etc.), then we'd remove the support in Plutarch 1.2.

It would be much better if we could just figure out how to make inference work in these cases, but it's not trivial.

The error message in question:
```
    • Couldn't match type: p0 -> Term s b0
                     with: Term s (b :--> c)
        arising from a use of ‘plam’
    • In the first argument of ‘($)’, namely ‘plam’
      In the expression: plam $ \ x y -> plet x $ \ _ -> perror
      In an equation for ‘_example’:
          _example = plam $ \ x y -> plet x $ \ _ -> perror
    • Relevant bindings include
        _example :: Term s (a :--> (b :--> c))
          (bound at Plutarch/Maybe.hs:20:1)
   |
20 | _example = plam $ \x y -> plet x $ \_ -> perror
```

The reason pertains to these two instances:
```haskell
instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b) => PLamN (a' -> b') (Term s (a :--> b)) where
  plam = plam'

instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c) => PLamN (a' -> b' -> c') (Term s (a :--> b :--> c)) where
  plam f = plam' $ \x -> plam (f x)
```

For some reason it's selecting the first instance rather than the second one. Because it chooses the first one, it does
`b' ~ Term s b`, which tries to unify the type of `\y -> plet x $ \_ -> perror` with `Term s (b :--> c)`, which obviously fails.

This can happen AFAICT because the instances are incoherent, but I'm not sure type inference can even work properly if they weren't incoherent, which was the problem I had before.

L-as:
Nevermind, I just realised I can fix this! Not sure why I went with incoherent instances originally. Closing this!

TotallyNotChase:
> Nevermind, I just realised I can fix this! Not sure why I went with incoherent instances originally. Closing this!

Holy shit - actually? I can't wait

blamario:
> it is confusing to the point that ~15 people could not figure out what is happening until I came in

To be fair, we ran into the issue only a minute or two before you came, and I was going to suggest two `plam'`s to try next. Anyway, all's well that ends well.

PR #144, opened 2022-01-19
PR Benchmark Diffing

MatthewCroughan:
This adds a new `effect` via https://github.com/hercules-ci/hercules-ci-effects which takes a pull request and performs a diff on the benchmark by using `nix run .#benchmark` on the PR `HEAD` and on the merge-base of that PR.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,6 +4,10 @@
   inputs.haskell-nix.url = "github:L-as/haskell.nix?ref=master";
   inputs.nixpkgs.follows = "haskell-nix/nixpkgs-unstable";
   inputs.flake-compat-ci.url = "github:hercules-ci/flake-compat-ci";
+  inputs.flake-compat = {
+    url = "github:edolstra/flake-compat";
+    flake = false;
+  };
 
   # https://github.com/input-output-hk/plutus/pull/4328
   inputs.plutus.url = "github:L-as/plutus?ref=master";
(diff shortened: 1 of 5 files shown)


L-as:
Please retarget to staging...

L-as:
Is there any issue on GH for the bug?

MatthewCroughan:
@L-as No, I'm in the process of making one.

MatthewCroughan:
Turns out there was no bug, it was just a missing `)` in the `effectScript` which was easily conflated with the Nix stdenv when debugging.

flake.nix
```diff
@@ -451,9 +456,43 @@
       );
       devShell = perSystem (system: self.flake.${system}.devShell);
 
-      nixCi = flake-compat-ci.lib.recurseIntoFlakeWith {
+      effects = { src }:
+        let
+          pkgs = nixpkgs.legacyPackages.x86_64-linux;
+          nix = (builtins.getFlake "github:matthewcroughan/nixcfg/f16df936ffc9850d42f68f4badef9b55d7e7fee8").nixosConfigurations.swordfish.pkgs.nixFlakes;
```
srid, line 462: Why do we need `github:matthewcroughan/nixcfg`? Isn't the nix executable available from the `pkgs` in line 461?

srid (approved):
Except for that one comment, we are good I think.

flake.nix
MatthewCroughan: Ah, I left that in by mistake, this was something I was debugging.

Issue #124, opened 2022-01-17
Add plutarch-benchmark library


srid:
We want `Benchmark.hs` exposed separately in a library called `plutarch-benchmark` (which the perf bench will depend on), so that it can be used by Plutarch smart contract projects for benchmarking their own functions.

---

As part of this, @L-as I suggest moving plutarch library (.hs, .cabal) to a sub directory `./plutarch` per convention of multi-package Cabal projects ([example](https://github.com/input-output-hk/plutus-apps)).

Issue #123, opened 2022-01-17
GitHub PR: benchmark diff


srid:
Produce reports like [this](https://github.com/input-output-hk/plutus/pull/3494#issuecomment-874588174) for PRs in this repo, so that we can assess the impact of the PRs on benchmarks (code size, cpu & men). Take `bench.csv` produced by Hercules Effects for both commit A (commit where the PR branch branched off) and commit B (PR commit), and diff them.

---

https://github.com/Plutonomicon/plutarch/pull/87 is an example of such PR which improved the `types:builtin:intlist:plistEquals` benchmark.

[Old](https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fi4gy7rp596lk0rq75yhwgc4rmk0a4zi3-benchmark.drv/log?via-job=494e885c-d1e9-4eb2-b76f-247059eaebf5)

```
types:builtin:intlist:plistEquals:==(n=3)            25843624(cpu)  59392(mem)  180(bytes)
types:builtin:intlist:plistEquals:/=(n=4)            12939651(cpu)  30928(mem)  181(bytes)
types:builtin:intlist:plistEquals:/=(empty;n=3)       7497147(cpu)  19064(mem)  177(bytes)
```

[New](https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Frnz3dlw59wiv6hl8i7864lb0r23grl8d-benchmark.drv/log?via-job=3fee9a77-90c8-49a3-9b85-b02eecf89cad)

```
types:builtin:intlist:plistEquals:==(n=3)             9398080(cpu)  20846(mem)  95(bytes)
types:builtin:intlist:plistEquals:/=(n=4)             9398080(cpu)  20846(mem)  96(bytes)
types:builtin:intlist:plistEquals:/=(empty;n=3)       1639885(cpu)   4664(mem)  92(bytes)
```

Diff'ing these two tables, we would add a comment to the PR showing the number differences for each row/column much like the Plutus report. In the Plutus repo, benchmark reporting is manually triggered by adding a PR comment containing `/benchmark`.

emiflake:
I was looking into implementing this. With Hercules, how would we create artifacts that we can use as means for comparison? It doesn't seem like Hercules publishes the bench.csv anywhere, or am I missing something?

srid:
@nixinator @MatthewCroughan Could one of you help Emily here?

[...]

PR #146, opened 2022-01-20
plutarch-benchmark with perf-diff binary

emiflake:
Draft because I want to clean up some of the diff code + it needs to be wired up to be used in the Hercules effect

Implements #123, #124


L-as:
@MatthewCroughan ?

MatthewCroughan:
Let's add this binary via a PR, and then let's make a separate PR to add it to the Hercules-CI Effect. But not in the same PR.

L-as (approved):
LGTM

cabal.project
```diff
@@ -1,3 +1,5 @@
-packages: ./.
+packages:
+  ./.
+  ./plutarch-benchmark/plutarch-benchmark.cabal
```
srid, line 3: I think `./plutarch-benchmark` is equivalent. Even `./*` IIRC.

PR #149, opened 2022-01-20
Fix instances for plam by making them overlapping and not incoherent

### LINKED ISSUE #2: Improve type inference of plam

L-as:
I think I originally made these instances incoherent because they looked slightly
different, which somehow made OVERLAPPING not work. Then I changed them, without noticing
that I could make them coherent.

### DIFF
--- a/examples/Examples/List.hs
+++ b/examples/Examples/List.hs
@@ -55,14 +55,15 @@ tests = do
           (pzipWith' (+) # integerList [1 .. 10] # integerList [1 .. 10])
             #== integerList (fmap (* 2) [1 .. 10])
     , testCase "pfoldl" $ do
+        let m :: Term _ (PInteger :--> PInteger :--> PInteger) = plam $ \x y -> x - y
         expect $
-          (pfoldl # plam (-) # 0 # integerList [1 .. 10])
+          (pfoldl # m # 0 # integerList [1 .. 10])
             #== pconstant (foldl (-) 0 [1 .. 10])
         expect $
           (pfoldl' (-) # 0 # integerList [1 .. 10])
             #== pconstant (foldl (-) 0 [1 .. 10])
         expect $
-          (pfoldl # plam (-) # 0 # integerList [])
+          (pfoldl # m # 0 # integerList [])
             #== pconstant 0
         expect $
           (pfoldl' (-) # 0 # integerList [])
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE IncoherentInstances #-}
 
 module Plutarch (
   (PI.:-->),
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE UndecidableInstances #-}
-{-# LANGUAGE IncoherentInstances #-}
 
 module Plutarch (
   (PI.:-->),
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -101,28 +101,22 @@ infixr 0 #$
 class PLamN a b | a -> b where
   plam :: a -> b
 
--- FIXME: This piece of code doesn't work unless you do (_ :: Term _ _)
-{-
-f :: Term s ((a :--> b) :--> b :--> b)
-f = plam $ \f _ -> f # perror
--}
-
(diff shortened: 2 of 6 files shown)


L-as:
And unfortunately this makes some other code break, but it seems worth it.

L-as:
This basically breaks the `id` function `plam $ \x -> x`...

L-as:
#150

L-as:
We probably want to merge this regardless of #150, since it makes error messages much more comprehendable when it does fail:
```
examples/Main.hs:199:37: error:
    • Overlapping instances for Plutarch.PLamN
                                  (p0 -> p0) (Term s1 (PInteger @{S} :--> a0))
        arising from a use of ‘plam’
      Matching instances:
        instance forall a' (s :: S) (a :: PType) b' (b :: PType).
                 ((a' :: Type) ~ (Term s a :: Type),
                  (b' :: Type) ~ (Term s b :: Type)) =>
                 Plutarch.PLamN (a' -> b') (Term s (a :--> b))
          -- Defined in ‘Plutarch’
        instance [overlapping] forall a' (s :: S) (a :: PType) b'
                                      (b :: PType) c' (c :: PType) d' (d :: PType) e' (e :: PType)
                                      f' (f :: PType) g' (g :: PType).
                               ((a' :: Type) ~ (Term s a :: Type),
                                (b' :: Type) ~ (Term s b :: Type),
                                (c' :: Type) ~ (Term s c :: Type),
                                (d' :: Type) ~ (Term s d :: Type),
                                (e' :: Type) ~ (Term s e :: Type),
                                (f' :: Type) ~ (Term s f :: Type),
                                (g' :: Type) ~ (Term s g :: Type)) =>
                               Plutarch.PLamN
                                 (a' -> b' -> c' -> d' -> e' -> f' -> g')
                                 (Term s (a :--> (b :--> (c :--> (d :--> (e :--> (f :--> g)))))))
          -- Defined in ‘Plutarch’
        instance [overlapping] forall a' (s :: S) (a :: PType) b'
                                      (b :: PType) c' (c :: PType) d' (d :: PType) e' (e :: PType)
                                      f' (f :: PType).
                               ((a' :: Type) ~ (Term s a :: Type),
                                (b' :: Type) ~ (Term s b :: Type),
                                (c' :: Type) ~ (Term s c :: Type),
                                (d' :: Type) ~ (Term s d :: Type),
                                (e' :: Type) ~ (Term s e :: Type),
                                (f' :: Type) ~ (Term s f :: Type)) =>
                               Plutarch.PLamN
                                 (a' -> b' -> c' -> d' -> e' -> f')
                                 (Term s (a :--> (b :--> (c :--> (d :--> (e :--> f))))))
          -- Defined in ‘Plutarch’
        ...plus three others
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of ‘p0, s1, a0’
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    • In the first argument of ‘($)’, namely ‘plam’
      In the second argument of ‘($)’, namely ‘plam $ \ x -> x’
      In the first argument of ‘(#)’, namely
        ‘(phoistAcyclic $ plam $ \ x -> x)’
    |
199 |         printTerm ((phoistAcyclic $ plam $ \x -> x) # (0 :: Term s PInteger)) @?= "(program 1.0.0 0)"
```

L-as:
I made instances incoherent again, but I had to add the pragma to the module rather than to the instances, because they need to be both overlapping and incoherent. Is there a way of doing this?

L-as:
It still fails when doing things like `plam (-)`, but it's good enough in 99% of cases AFAICT.

L-as:
This doesn't work well at all actually, but semiworks within GHCi when you add the language pragma after loading the project...

examples/Examples/List.hs
```diff
@@ -55,14 +55,15 @@ tests = do
           (pzipWith' (+) # integerList [1 .. 10] # integerList [1 .. 10])
             #== integerList (fmap (* 2) [1 .. 10])
     , testCase "pfoldl" $ do
+        let m :: Term _ (PInteger :--> PInteger :--> PInteger) = plam $ \x y -> x - y
         expect $
-          (pfoldl # plam (-) # 0 # integerList [1 .. 10])
+          (pfoldl # m # 0 # integerList [1 .. 10])
```
blamario, line 60: What's the reason for this change?

blamario (approved):

TotallyNotChase:
Is this contending against #150 ? As in, I assume you're planning to merge only one of these?

L-as:
They don't conflict, but merging #150 would entail eventually removing the curried syntax, probably in Plutarch 1.2, and deprecating them for 1.1.

examples/Examples/List.hs
L-as: Making them coherent breaks this. In general the curried syntax just doesn't work well.

L-as: The type of `m` also needs to be written explicitly unfortunately.

PR #150, opened 2022-01-20
Add support for using tuples with `plam`

### LINKED ISSUE #2: Improve type inference of plam

L-as:
Fixes #2

#143 reopened.

### DIFF
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -125,6 +125,21 @@ instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ T
 instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d, e' ~ Term s e, f' ~ Term s f, g' ~ Term s g) => PLamN (a' -> b' -> c' -> d' -> e' -> f' -> g') (Term s (a :--> b :--> c :--> d :--> e :--> f :--> g)) where
   plam f = plam' $ \x -> plam (f x)
 
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c) => PLamN ((a', b') -> c') (Term s (a :--> b :--> c)) where
+  plam f = plam $ \x y -> f (x, y)
+
+instance {-# INCOHERENT #-} (a' ~ Term s a, b' ~ Term s b, c' ~ Term s c, d' ~ Term s d) => PLamN ((a', b', c') -> d') (Term s (a :--> b :--> c :--> d)) where
+  plam f = plam $ \x y z -> f (x, y, z)
+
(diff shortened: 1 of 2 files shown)


L-as:
After asking internally and messing around with #149 , I've come to the conclusion that this is probably the best solution unfortunately.

TotallyNotChase:
To be transparent, I've been quite ok with the current behavior (and syntax) of `plam`. I think the guide addresses the only workaround you'll ever need for all `plam` related issues already. We could even put it in the `plam` haddock.

However, if we do go for this syntax and everyone's fine with it - I think we really should either remove the other syntax in the same PR, or make a separate function for tuple based syntax.

L-as:
That would be a huge backward incompatible change though.

srid:
> That would be a huge backward incompatible change though.

We can afford to break backwards compatibility for 1.1 can't we?

L-as:
That's true honestly, we could just do the change right now and break everything.

L-as:
Are people for or against changing the syntax to tuples in a breaking change, and removing the old syntax for Plutarch 1.1? If we stay with the current syntax, there is essentially no way of fixing #2 AFAICT. Please react with :+1: or :-1: .

blamario:
I'm not happy with the idea of dropping the existing `plam` interface because it fits so well with the Haskell function calling style when it works. I can't guess all the places where I'd be missing it. Scott encodings don't take tuples for example.

For the time being, can we add a new function, call it `tuplam` or something, and put it forefront it in the user guide? If, in time, it turns out that many of its uses require `uncurry` we'll know it can't be the only solution.

L-as:
> I can't guess all the places where I'd be missing it. Scott encodings don't take tuples for example.

What do you mean by this?

blamario:
Only that a Haskell function that's a Scott encoding of a data structure would not be taking a tuple. But more generally, if you're relying on the `base` library you're pretty likely to end up with a point-free multi-argument function at some point. The existing `plam` can work with them, on a good day anyway, the new `tuplam` would require naming and tupling all the arguments.

I realize this may read like FUD right now. I'll try to come up with concrete examples and if I can't I'll change my vote.

blamario:
I tried a few things and the inference is even worse than I imagined. I got so annoyed that I tried to implement something myself (branch `mario/plam` is you care to look) and failed. So you have a green light from me.

blamario:
Do have a look actually. I've improved the inference to the extent I was able to restore the original staging `Plutarch.Maybe` and `Plutarch.Either` modules, but `plam const` still fails with this ridiculous error of
```
    • Couldn't match type ‘b0’ with ‘Term s b'0’
        arising from a use of ‘plam’
        because type variable ‘s’ would escape its scope
      This (rigid, skolem) type variable is bound by
        a type expected by the context:
          Plutarch.Internal.ClosedTerm
            (a0 :--> (Plutarch.LambdaArg b0 :--> a0))
        at examples/Main.hs:221:23-34
    • In the first argument of ‘printTerm’, namely ‘(plam const)’
```

L-as:
```haskell
ghci> :t plam const
plam const
  :: forall {s :: PI.S} {b :: PType} {a :: PType}.
     Term s (b :--> (a :--> b))
```
with #168

L-as:
We don't need this anymore.

PR #151, opened 2022-01-20
Add `PlutusType` and `PIsData` deriver for newtypes + instances for ledger API types

TotallyNotChase:
This is the result of me hammering my head against deriving via for half a day 😅


Plutarch/Deriver/PNewtype.hs
```diff
@@ -0,0 +1,45 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
TotallyNotChase, line 1: I put this under `Plutarch.Deriver.PNewType`. If we go with this separate deriver organization - we should probably put the other derivers within `Deriver.*` as well.

If not, let me know where to put this file.

Plutarch/Deriver/PNewtype.hs
```diff
@@ -0,0 +1,45 @@
...
+  pmatch' x f = f . DerivePNewType $ ptypeOuter x
+
+instance (forall (s :: S). Coercible (a s) (Term s b), PIsData b) => PIsData (DerivePNewType a b) where
+  pfromData x = pcon . DerivePNewType $ ptypeOuter target
+    where
+      target :: Term _ b
+      target = pfromData $ pinnerData x
+  pdata x = pouterData . pdata $ pto x
+
+ptypeInner :: forall (x :: PType) y s. Coercible (x s) (Term s y) => x s -> Term s y
```
TotallyNotChase, line 35: should these utility types be exported from somewhere?

Plutarch/Api/V1.hs
```diff
@@ -533,6 +602,8 @@ instance PIsDataRepr PDCert where
 newtype PMap (a :: PType) (b :: PType) (s :: S)
   = PMap (Term s (PBuiltinMap a b))
 
+-- TODO: PMap needs instances for PConstant/PLift, PlutusType, and PIsData.
```
TotallyNotChase, line 605: Currently getting a bunch of type checking errors when I try doing this. Want to tackle it on a separate PR if possible.

t1lde: I was running into many headaches with this too...
Manual instances would be IMO acceptable if necessary.

Plutarch/Api/V1.hs
```diff
@@ -188,76 +190,143 @@ newtype PDatum (s :: S) = PDatum (Term s PData)
...
 deriving via (DerivePConstantViaNewtype Plutus.StakeValidatorHash PStakeValidatorHash PByteString) instance (PConstant Plutus.StakeValidatorHash)
+deriving via (DerivePNewType PStakeValidatorHash PByteString) instance (PlutusType PStakeValidatorHash)
+deriving via (DerivePNewType PStakeValidatorHash PByteString) instance (PIsData PStakeValidatorHash)
 
 newtype PRedeemerHash (s :: S)
   = PRedeemerHash (Term s PByteString)
 
 instance PUnsafeLiftDecl PRedeemerHash where type PLifted PRedeemerHash = Plutus.RedeemerHash
 deriving via (DerivePConstantViaNewtype Plutus.RedeemerHash PRedeemerHash PByteString) instance (PConstant Plutus.RedeemerHash)
+deriving via (DerivePNewType PRedeemerHash PByteString) instance (PlutusType PRedeemerHash)
```
L-as, line 224: I don't like the fact that we're using standalone derives here. If you flip around the parameters to `DerivePNewType`, it works fine without standalone derives.

Plutarch/Deriver/PNewtype.hs
L-as: I don't think so.

Plutarch/Deriver/PNewtype.hs
```diff
@@ -0,0 +1,45 @@
...
+
+For any newtype represented as-
+> newtype PFoo (s :: S) = PFoo (Term s PBar)
+
+where 'PBar' has a 'PIsData' instance, you can derive 'PlutusType' and 'PIsData' using-
+> deriving (PlutusType, PIsData) via (DerivePNewType PFoo PBar)
+
+This will make 'PFoo' simply be represnted as 'PBar' under the hood.
+-}
+newtype DerivePNewType a (b :: PType) s = DerivePNewType (a s)
```
L-as, line 21: Wrt. naming: `DerivePConstantViaNewtype` is `Newtype` and not `NewType`. We probably want to use the same convention for both. Feel free to choose either.

L-as (changes requested):
Looks mostly good to me

Plutarch/Deriver/PNewtype.hs
L-as: We probably want to export this in the prelude TBH.

Plutarch/Api/V1.hs
TotallyNotChase: Oh yeah what the heck. Silly me, will fix this! I also don't like standalone deriving unnecessarily.

Plutarch/Deriver/PNewtype.hs
L-as: Well the derivers should usually be next to the definition of the class itself. This one deriver is a special-case IMO, so it's probably better to put in `Plutarch.hs` and export from `Plutarch.Prelude` too.

TotallyNotChase: Can't put it directly in `Plutarch` due to the dependency on `.Builtin`, just put it in prelude maybe?

L-as: You'd define `DerivePNewtype` (or whatever name you choose) in `Plutarch`, then have an "orphan" instance in `Plutarch.Builtin`. It's not really an orphan instance because it's impossible to import the class without also importing the instance for `DerivePNewtype`, hence it's 100% safe.

L-as: Similarly, downstream users who make their own classes, e.g. `PMonad`, would also implement an instance for this type in the same module they define the class.

TotallyNotChase: That makes sense, done!

Plutarch/Api/V1.hs
TotallyNotChase: Fixed!

L-as (approved):
LGTM

PR #153, opened 2022-01-20
Update GUIDE.md

jhodgdev:
Fixed arithmetic mistake in guide.


TotallyNotChase (approved):
I'm so good at math. Thanks!

jhodgdev:
> I'm so good at math. Thanks!

As if learning Plutarch isn't enough of an endeavour without division being redefined :joy:

PR #158, opened 2022-01-20
Enable api tests (partially)

TotallyNotChase:
This enables a few of the existing API tests that didn't work due to missing instances.

The examples that depend on `PMap` still don't work due to missing instances.

I'm going to add a more involved test as well - should ensure some level of stability.


TotallyNotChase:
Added another validator test - a bit more involved this one.

L-as:
Might overlap with #159 ?

TotallyNotChase:
> Might overlap with #159 ?

Slightly it seems; would that be an issue?

L-as (approved):
LGTM

Issue #95, opened 2022-01-12
Convenient field accessors for use with API types


t1lde:
The API types are still pretty rough to use IMO, the dance with `pmatch` and such is pretty messy, and `PAsData` also adds some noise.

To address this, I want to extend `DataRepr` with named fields, so we can have something that looks like:

```haskell
pfield #fieldName :: (Term s (r :--> a))
```

that can also work over `PAsData`:

```haskell
pfield #fieldName :: (Term s (r :--> PAsData a))
```

This could be used like:

```haskell
x & pfield #fieldName
```
, or perhaps with an operator that applies the value:

```haskell
x #. #fieldName
```

Which can also support RDS:

```haskell
x.fieldName
```

it would also be nice to have numbered fields for `PIsDataRepr` types without field names, and for `PDataList`, which would be useful when matching sum types. 

Perhaps, if the type-level stuff works out:

```haskell
pfield #0 :: (Term s (r :--> a))
```

L-as:
Indexing the data list is always suboptimal because it will lead to excess code. There might still be a way to express field names however. An easy solution is to just use newtypes.

t1lde:
the cost of indexing field names should be relatively constant over most scripts, with hoisting.

I'm not sure what a smaller alternative would be, aside from maybe repeatedly getting the head of the DataList in some cases...

L-as:
Accessing all n elements of a record will use O(n^2) CPU if all elements are accessed using indexing. The accessor functions can be shared across all data structures, but it's still far from optimal.

L-as:
I think it's possible to make an HList-like type such that you can take the first N elements from the `PDataList` and be able to access the fields using the `HasField` type class. It would generate exactly N calls to `HeadList`.

L-as:
Another partial solution is to always take all N fields, which would simplify the types.

L-as:
FWIW since we have GHC 9 now we can use record dot syntax.

PR #159, opened 2022-01-20
Efficiently binding data fields to a HRec  #95

t1lde:
Addressing #95 

(will write this up & fix remaining things tomorrow)


Plutarch/Api/V1.hs
```diff
@@ -66,14 +66,20 @@ import Plutarch.Builtin (PAsData, PBuiltinList, PData, PIsData, type PBuiltinMap
 import Plutarch.ByteString (PByteString)
 import Plutarch.DataRepr (
   DataReprHandlers (DRHCons, DRHNil),
+<<<<<<< HEAD
+  PDataRecord,
+=======
```
L-as, line 71: lol

Plutarch/Api/V1.hs
```diff
@@ -401,17 +454,23 @@ instance PIsDataRepr (PExtended a) where
...
     via (PIsDataReprInstances PCredential)
+>>>>>>> origin/staging
 
 instance PIsDataRepr PCredential where
   type
     PIsDataReprRepr PCredential =
-      '[ '[PPubKeyHash]
-       , '[PValidatorHash]
+      '[ '["_0" ':= PPubKeyHash]
+       , '["_0" ':= PValidatorHash]
```
L-as, line 473: this doesn't look right

Plutarch/Api/V1.hs
t1lde: oops :woozy_face:

Plutarch/Api/V1.hs
t1lde: I think it's correct, though field naming for single item constructors could be different...

Plutarch/DataRepr.hs
```diff
@@ -35,60 +38,67 @@ import Plutarch.Builtin (
...
 import qualified PlutusCore as PLC
 
-data PDataList (as :: [PType]) (s :: S)
+data PDataRecord (as :: [PLabeled]) (s :: S)
 
-pdhead :: Term s (PDataList (a : as) :--> PAsData a)
+data PLabeled = Symbol := PType
+
+pdhead :: Term s (PDataRecord ((l ':= a) : as) :--> PAsData a)
 pdhead = phoistAcyclic $ pforce $ punsafeBuiltin PLC.HeadList
```
t1lde, line 54: I wonder if we want more operations for working with `PDataRecord`s, or if all of this is sufficient :thinking:

Plutarch/Field.hs
```diff
@@ -0,0 +1,190 @@
...
+  ) =>
+  PDataFields (DerivePDataFields a)
+  where
+  type
+    PFields (DerivePDataFields a) =
+      SingleItem (PIsDataReprRepr a)
+
+  ptoFields t =
+    (punsafeCoerce $ phoistAcyclic $ plam $ \d -> psndBuiltin #$ pasConstr # d)
+      # (punsafeCoerce t :: Term _ PData)
```
t1lde, line 95: not 100% on how sound this is, in theory if `PIsDataRepr` is being used properly then it should work this way but perhaps there's a safer way to impl this...

A `PlutusType` impl for `PIsDataReprInstances` would do the trick, I think.
Or maybe `(Coercible a (PDataRecord (PFields (DerivePDataFields a))))`

Plutarch/Field.hs
```diff
@@ -0,0 +1,190 @@
...
+instance {-# OVERLAPS #-} BindFields ((l ':= a) ': '[]) where
+  bindFields t =
+    pure $ HCons (pdhead # t) HNil
+
+instance {-# OVERLAPPABLE #-} (BindFields as) => BindFields ((l ':= a) ': as) where
+  bindFields t = do
+    t' <- TermCont $ plet t
+    -- tail <- TermCont $ plet $ pdtail # t
+    xs <- bindFields @as (pdtail # t')
+    pure $ HCons (pdhead # t') xs
```
t1lde, line 167: Could the bindings be more efficient here? This generates pretty decent code, but maybe there's some free wins missing

Plutarch/Field/HList.hs
```diff
@@ -0,0 +1,180 @@
...
+
+-- | Indexing type-level lists
+type family IndexList (n :: Nat) (l :: [k]) :: k where
+  IndexList 0 (x ': _) = x
+  IndexList n (x : xs) = IndexList (n - 1) xs
+
+-- | Get the Index of a type in a list
+type family IndexOf (x :: k) (xs :: [k]) :: Nat where
+  IndexOf x (x ': _) = 0
+  IndexOf x (y ': xs) = (IndexOf x xs + 1)
```
t1lde, line 101: Probably want to add some type errors

Plutarch/Api/V1.hs
```diff
@@ -154,24 +183,30 @@ instance PIsDataRepr PScriptContext where
 -- General types, used by V1 and V2
 
 data PScriptPurpose (s :: S)
-  = PMinting (Term s (PDataList '[PCurrencySymbol]))
-  | PSpending (Term s (PDataList '[PTxOutRef]))
-  | PRewarding (Term s (PDataList '[PStakingCredential]))
-  | PCertifying (Term s (PDataList '[PDCert]))
+  = PMinting (Term s (PDataRecord '["_0" ':= PCurrencySymbol]))
```
t1lde, line 186: There's an opportunity to revise the naming. 
I've defaulted to numeric naming like this for everything in plutus that doesn't itself provide field names, but these could be, for example, something like 'get', or be named after their constructors.

Plutarch/Api/V1.hs
L-as: So it's legal to use the same field name twice?

examples/Examples/Api.hs
```diff
@@ -1,27 +1,30 @@
 {-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE QualifiedDo #-}
```
L-as, line 2: We should just add this to the cabal file

Plutarch/Api/V1.hs
t1lde: yep, in the case of sum types you can't use `pletFields` directly, you have to `pmatch` first, so it's safe.

examples/Examples/Api.hs
t1lde: I'm still being careful about compat, but we can add just for the test-suite ig

Plutarch/Field.hs
L-as: What's the commented out code?

t1lde: Different version I tried, will remove

Plutarch/Api/V1.hs
L-as: I think this is fine TBH. It's sort of reminiscent of Rust's "newtypes".

L-as:
I'm worried that this PR breaks GHC 8.10.7 support. We should perhaps add CI for that.

L-as:
Specifically, OverloadedRecordDot is used in Plutarch.Field.

t1lde:
> Specifically, OverloadedRecordDot is used in Plutarch.Field.

Oh it's not actually used, removed.

examples/Examples/Api.hs
```diff
@@ -1,29 +1,25 @@
...
-import Data.Proxy (Proxy (..))
 import Plutarch
 import Plutarch.Api.V1 (
-  PAddress (PAddress),
-  PCredential (PScriptCredential),
+  PCredential (..),
   PCurrencySymbol,
   PPubKeyHash,
-  PScriptContext (PScriptContext),
+  PScriptContext (..),
```
L-as, line 8: Please don't import `..`. See styleguide.

Plutarch/Field/HList.hs
```diff
@@ -0,0 +1,193 @@
...
+-- | Usual GADT Heterogenous List encoding
+data HList (xs :: [Type]) where
+  HNil :: HList '[]
+  HCons :: x -> HList xs -> HList (x ': xs)
+
+{- |
+  Heterogenous Record, using a list of symbols as a
+  set of corresponding indices by position.
+-}
+newtype HRec (fs :: [Symbol]) (as :: [Type])
```
L-as, line 49: i think it would be simpler if you merged `fs` and `as` like done elsewhere. You can remove `HList` too.

examples/Examples/Field.hs
```diff
@@ -0,0 +1,293 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+module Examples.Field (
+  -- * Examples
+  Triplet (..),
+  mkTrip,
+  tripA,
+  tripB,
+  tripC,
```
L-as, line 9: Why do you export all of this?

examples/Examples/Field.hs
```diff
@@ -0,0 +1,293 @@
...
+
+--------------------------------------------------------------------------------
+
+import Plutarch
+import Plutarch.Builtin (PAsData, PBuiltinList, PIsData (..))
+import Plutarch.DataRepr (
+  DataReprHandlers (..),
+  PDataRecord,
+  PIsDataRepr (..),
+  PIsDataReprInstances (..),
```
L-as, line 34: Please no `..` imports.

Plutarch/Field.hs
```diff
@@ -0,0 +1,261 @@
...
+  DerivePDataFields (..),
+  pletFields,
+  pletNFields,
+  pletDropFields,
+  pletRangeFields,
+  pfield,
+  pfield',
+
+  -- * BindFields class mechanism
+  BindFields (..),
```
L-as, line 16: Please no `..` imports.

L-as (changes requested):
Looks great! Some small nitpicks.

Plutarch/DataRepr.hs
```diff
@@ -35,60 +40,85 @@ import Plutarch.Builtin (
...
-punDataRepr :: Term s (PDataRepr '[def] :--> PDataList def)
+punDataRepr :: Term s (PDataRepr '[def] :--> PDataRecord def)
 punDataRepr = phoistAcyclic $
   plam $ \t ->
     plet (pasConstr #$ pasData t) $ \d ->
-      (punsafeCoerce $ psndBuiltin # d :: Term _ (PDataList def))
+      (punsafeCoerce $ psndBuiltin # d :: Term _ (PDataRecord def))
 
-pindexDataRepr :: (KnownNat n) => Proxy n -> Term s (PDataRepr (def : defs) :--> PDataList (IndexList n (def : defs)))
+pindexDataRepr :: (KnownNat n) => Proxy n -> Term s (PDataRepr (def : defs) :--> PDataRecord (IndexList n (def : defs)))
```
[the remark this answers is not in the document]
t1lde: I like `PDataSum`

Plutarch/DataRepr.hs
```diff
@@ -35,60 +40,85 @@ import Plutarch.Builtin (
...
+  PTypes ((l ':= t) ': as) = t ': (PTypes as)
+
+type family PNames (as :: [PLabeled]) :: [Symbol] where
+  PNames '[] = '[]
+  PNames ((l ':= t) ': as) = l ': (PNames as)
+
+type PDataRepr :: [[PLabeled]] -> PType
+data PDataRepr (defs :: [[PLabeled]]) (s :: S)
 
 pasData :: Term s (PDataRepr _) -> Term s PData
```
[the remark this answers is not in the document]
t1lde: :shrug: , could be convenient to keep this though

Plutarch/Api/V1.hs
```diff
@@ -1,5 +1,5 @@
 {-# LANGUAGE UndecidableInstances #-}
-{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wwarn=orphans #-}
```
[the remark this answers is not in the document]
t1lde: Makes the orphans more visible

Plutarch/DataRepr.hs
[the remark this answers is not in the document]
L-as: It's essentially `pforgetData . pdata`.

examples/Examples/Field.hs
t1lde: visibility in docs

Plutarch/Api/V1.hs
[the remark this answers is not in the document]
L-as: That's true, however, I don't think anyone would look at the warnings anyway, since they're 99% bogus.

Plutarch/Api/V1.hs
```diff
@@ -95,68 +97,81 @@ import qualified PlutusTx.AssocMap as PlutusMap
...
+               , "signatories" ':= PBuiltinList (PAsData PPubKeyHash)
+               , "data" ':= PBuiltinList (PAsData (PTuple PDatumHash PDatum))
+               , "id" ':= PTxId
                ]
           )
       )
   deriving
     (PMatch, PIsData)
     via PIsDataReprInstances PTxInfo
+  deriving (PDataFields) via (DerivePDataFields PTxInfo)
```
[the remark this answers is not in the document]
t1lde: Yeah I agree, initially I wanted them to be separate to provide field names separately via the tag, but went back on that...

Plutarch/Field/HList.hs
```diff
@@ -0,0 +1,193 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Plutarch.Field.HList (
```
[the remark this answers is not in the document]
t1lde: makes sense

Plutarch/Field/HList.hs
```diff
@@ -0,0 +1,193 @@
...
+  --    Example:
+  --
+  --    >>> natElem @_ @0
+  --    Here
+  --
+  --    >>> natElem @_ @3
+  --    There (There (There Here))
+  natElem :: Elem x xs
+
+instance {-# OVERLAPS #-} NatElem 0 x (x ': xs) where
```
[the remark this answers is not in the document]
t1lde: I forgot the difference

Plutarch/Field.hs
```diff
@@ -0,0 +1,261 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Plutarch.Field (
```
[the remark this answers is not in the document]
t1lde: Makes sense

Plutarch/Field/HList.hs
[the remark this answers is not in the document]
L-as: https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/instances.html#overlapping-instances

You probably don't want to use `OVERLAPS`.

Plutarch/List.hs
```diff
@@ -170,6 +174,26 @@ punsafeIndex = phoistAcyclic $
         (phead # xs)
         (self # (n - 1) #$ ptail # xs)
 
+{- |
+  Version of 'punsafeIndex' using repeated application of 'ptail', with
+  'pdrop'.
+-}
+punsafeIndex' :: (PIsListLike list a) => Natural -> Term s (list a) -> Term s a
+punsafeIndex' n xs = phead # (pdrop n xs)
```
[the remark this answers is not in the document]
t1lde: a different convention for partials is a good idea, sadly p is already taken :thinking:

Plutarch/Field/HList.hs
t1lde: Went back on forth on how relevant the internal HList would be... it would take a little rewrite to have the more usual encoding... probably worthwhile though

t1lde:
my work week has come to an end, will have to get to addressing these Monday

L-as:
@t1lde if you use a branch from here it's easier for us to collaborate

Plutarch/DataRepr.hs
```diff
@@ -35,60 +40,85 @@ import Plutarch.Builtin (
...
+import Plutarch.List (pdrop, punsafeIndex, punsafeIndex')
 import Plutarch.Prelude
 import qualified Plutus.V1.Ledger.Api as Ledger
 import qualified PlutusCore as PLC
 
-data PDataList (as :: [PType]) (s :: S)
+data PDataRecord (as :: [PLabeled]) (s :: S)
 
-pdhead :: Term s (PDataList (a : as) :--> PAsData a)
+data PLabeled = Symbol := PType
```
srid, line 53: `PLabeledType` for clarity?

Plutarch/DataRepr.hs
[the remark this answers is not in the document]
srid: @t1lde Wouldn't it be `PDataProduct`, because you are using this in products - such as `PTxInfo` fields, and the arguments of a sum constructor like `PMinting`, rather than sum types per se (for which we use Haskell ADT syntax).

It is the "P" in "SOP" (sum of products) of `generics-sop`.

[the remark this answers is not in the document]
L-as: `PDataSum` is a replacement for `PDataRepr`.

srid:
I opened my PR directly on top of this branch to obviate having to deal with avoidable merge conflicts later: https://github.com/t1lde/plutarch/pull/1

Plutarch/Api/V1.hs
```diff
@@ -431,19 +470,32 @@ instance PIsDataRepr PCredential where
...
-  | PStakingPtr (Term s (PDataList '[PInteger, PInteger, PInteger]))
+  = PStakingHash (Term s (PDataRecord '["_0" ':= PCredential]))
+  | PStakingPtr
+      ( Term
+          s
+          ( PDataRecord
+              '[ "_0" ':= PInteger
+               , "_1" ':= PInteger
+               , "_2" ':= PInteger
+               ]
```
srid, line 481: There is an opportunity to generalize this using a `PDataList` that is computed from `PDataRecord`. You can even redefine `PTuple` in terms of it. Something like:

```haskell
type PDataList (as :: [PType]) s =
  PDataRecord (Indexed as 'Z)

type family Indexed (as :: [PType]) (start :: Nat) :: [PLabeled] where
  Indexed '[] _ = '[]
  -- TODO: convert `Nat` to `Symbol`
  Indexed (a ': as) n = "_?" ':= a ': Indexed as ('S n)

type PTuple' a b s =
  PDataList '[a, b] 'Z s
```

L-as: I don't think this would be worth it.

srid: Now that I think about it, we do _want_ to be explicit here in regards to indices, so as to not inadvertently break decoding of Data. So ignore my suggestion.

This explicitness is enabled in products only, however. Not sums, unfortunately.

t1lde: I tried out something like this, it was more of a headache than is worth it

flake.nix
```diff
@@ -450,7 +450,7 @@
         // {
           benchmark = {
             type = "app";
-            program = "${self.flake.${system}.packages."plutarch:bench:perf"}/bin/perf";
+            program = "${self.flake.${system}.packages."plutarch-benchmark:bench:perf"}/bin/perf";
```
L-as, line 453: What?

t1lde:
probably didn't want to merge that, will revert and open a new PR on a proper branch

Plutarch/Api/V1.hs
```diff
@@ -95,68 +97,81 @@ import qualified PlutusTx.AssocMap as PlutusMap
...
+          ( PDataRecord
+              '[ "txInfo" ':= PTxInfo
+               , "purpose" ':= PScriptPurpose
+               ]
+          )
+      )
   deriving
     (PMatch, PIsData)
     via PIsDataReprInstances PScriptContext
+  deriving (PDataFields) via (DerivePDataFields PScriptContext)
```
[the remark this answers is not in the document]
t1lde: not sure what you mean

Plutarch/DataRepr.hs
t1lde: done :)

Plutarch/Api/V1.hs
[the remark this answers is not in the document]
L-as: Indeed, this isn't right. I thought this was `ScriptPurpose`.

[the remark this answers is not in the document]
L-as: In fact, I added the comment in the wrong place! See a few lines below the instance for `PScriptPurpose`, it is bogus, because the type family doesn't reduce.

PR #160, opened 2022-01-21
Plutarch Numeric hierarchy

sergesku:


Plutarch/Api/V1.hs
```diff
@@ -290,8 +290,6 @@ deriving via
 
 newtype PPOSIXTime (s :: S)
   = PPOSIXTime (Term s PInteger)
-  deriving (PIntegral) via (PInteger)
```
L-as, line 293: ?

Plutarch/Builtin.hs
```diff
@@ -228,15 +228,15 @@ instance PIsData PBool where
     (phoistAcyclic $ plam toBool) # pforgetData x
     where
       toBool :: Term s PData -> Term s PBool
-      toBool d = pfstBuiltin # (pasConstr # d) #== 1
+      toBool d = pfstBuiltin # (pasConstr # d) #== pconstant 1
```
L-as, line 231: Why?

Plutarch/DataRepr.hs
```diff
@@ -84,7 +95,7 @@ pindexDataList n =
       punsafeIndex @PBuiltinList @PData # ind
   where
     ind :: Term s PInteger
-    ind = fromInteger $ natVal n
+    ind = pconstant $ natVal n
```
L-as, line 98: Why?

Plutarch/Integer.hs
```diff
@@ -21,37 +42,89 @@ data PInteger s
...
-  prem = punsafeBuiltin PLC.RemainderInteger
-
 instance PEq PInteger where
   x #== y = punsafeBuiltin PLC.EqualsInteger # x # y
 
 instance POrd PInteger where
   x #<= y = punsafeBuiltin PLC.LessThanEqualsInteger # x # y
   x #< y = punsafeBuiltin PLC.LessThanInteger # x # y
 
-instance Num (Term s PInteger) where
```
L-as, line 43: Why remove this instance?

Plutarch/Integer.hs
```diff
@@ -21,37 +42,89 @@ data PInteger s
...
+                  (punsafeBuiltin PLC.QuotientInteger # x # y)
+                  (punsafeBuiltin PLC.RemainderInteger # x # y)
+            )
+      )
+
+-- | Raise by an 'Integer' power.
+ppowInteger :: PMultiplicativeGroup a => Term s a -> Term s PInteger -> Term s a
+ppowInteger a int =
+  phoistAcyclic
+    ( plam $ \x i ->
```
L-as, line 85: do-syntax please

Plutarch/Integer.hs
```diff
@@ -21,37 +42,89 @@ data PInteger s
...
+premainderInteger = punsafeBuiltin PLC.RemainderInteger
+
+-- Helpers
+
+-- We secretly know that i is always positive.
+pexpBySquaring ::
+  forall s a.
+  (PMultiplicativeMonoid a) =>
+  Term s (a :--> PInteger :--> a)
+pexpBySquaring = pfix #$ plam f
```
L-as, line 124: This isn't hoisted.

Plutarch/List.hs
```diff
@@ -154,8 +160,8 @@ plength :: PIsListLike list a => Term s (list a :--> PInteger)
 plength = phoistAcyclic $
   plam $ \xs ->
     let go :: PIsListLike list a => Term s (list a :--> PInteger :--> PInteger)
-        go = (pfix #$ plam $ \self ls n -> pelimList (\_ xs -> self # xs # n + 1) n ls)
-     in go # xs # 0
+        go = (pfix #$ plam $ \self ls n -> pelimList (\_ xs -> self # xs # n #+ pone) n ls)
+     in go # xs # pzero
```
L-as, line 164: Why?

Plutarch/Natural.hs
```diff
@@ -0,0 +1,216 @@
...
+        pif
+          (i #< pconstant 0)
+          (pcon PNothing)
+          (pcon . PJust . pcon . PNatural $ i)
+    )
+    # x
+
+-- We made sure that PInteger >= zero as it goes from PNatural.
+pnatFromIntUnsafe :: Term s PInteger -> Term s PNatural
+pnatFromIntUnsafe = pcon . PNatural
```
L-as, line 203: In general it should be `punsafe*`.

Plutarch/Natural.hs
```diff
@@ -0,0 +1,216 @@
...
+  pnegate,
+ )
+import Plutarch.Pair (PPair (PPair))
+import Plutarch.Prelude
+
+newtype PNatural (s :: S) = PNatural (Term s PInteger)
+
+instance PIsData PNatural where
+  pfromData x =
+    phoistAcyclic
```
L-as, line 47: You can just use `pfromData` for integers here without all this complex logic. The argument passed to this function is `PAsData PNatural`, so we already know it's correct. No need to check anything.

Plutarch/Natural.hs
```diff
@@ -0,0 +1,216 @@
...
+      ( plam $ \x' ->
+          pmatch (pnatFromInt $ pasInt # pforgetData x') $ \case
+            PJust n -> n
+            PNothing -> perror
+      )
+      # x
+  pdata x =
```
L-as, line 54: No need for the unnecessary lambda here either.

L-as (changes requested):
In general looks good, very few missed hoists, etc., although I'm not really sure why you're removing the `Num` instances?

sergesku:
I`m not sure about `Num` and `PIntegral` instances.
I get the task from @kozross to replace `Num` and `Fractional` instances with the new `Numeric` hierarchy. 

And I must admit, using `pconstant` on numeric literals all the time is a pain.

L-as:
I think it's good to have the new classes, but I don't see why `Num` should be removed. It's quite convenient.

sergesku:
OK, I'll put them back

Plutarch/Integer.hs
L-as: We can't actually use do-syntax, because it breaks GHC 8.10.7...

TotallyNotChase:
I'm trying to get this ready for Plutarch 1.2, as part of `plutarch-extra`.

L-as:
@TotallyNotChase is this ready?

TotallyNotChase:
@L-as Not yet, I'm going to do some finishing touches post #435 and then we should get a review from Koz and fill in any missing pieces.

PR #161, opened 2022-01-21
Add `PIsData` and `PlutusType` instances for `PMap` and `PValue`

TotallyNotChase:
Here it is, it works!

We're still missing `PConstant`/`PLift` instance for `PMap` and `PValue` - I can't figure out the error but I have documented it. With that said, I'd argue a `PConstant`/`PLift` instance isn't as important right now.

### DIFF
--- a/Plutarch/Api/V1.hs
+++ b/Plutarch/Api/V1.hs
@@ -245,12 +245,11 @@ deriving via
 newtype PValue (s :: S) = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName PInteger)))
   deriving
     ( PlutusType
-    -- FIXME: This requires `PIsData` instance for `PMap`.
-    -- , PIsData
+    , PIsData
     )
     via (DerivePNewtype PValue (PMap PCurrencySymbol (PMap PTokenName PInteger)))
 
--- FIXME: This fails typecheck.
+-- FIXME: This fails typecheck. Representations do not match.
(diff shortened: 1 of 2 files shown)


TotallyNotChase:
actually I think I can implement the remaining instances manually. Don't merge this yet

L-as:
`PLift a` is necessary for `PBuiltinList a`.

TotallyNotChase:
Never mind, the `PConstant` and `PLift` instances must follow later. Faced some issues. This PR can be merged on its own.

> If further compromise is needed just to get lifting for Value working, then it itself could just be monomorphically special-cased.

So, the instance for `Value` comes naturally and automatically from the deriver if `PMap` has the correct instances. I have tested this. It's just the implementation for `PMap` that I'm stuck on right now. Getting endless missing constraints.....

Plutarch/Api/V1.hs
```diff
@@ -572,10 +571,15 @@ instance PIsDataRepr PDCert where
 
 ---------- AssocMap
 
-newtype PMap (a :: PType) (b :: PType) (s :: S)
-  = PMap (Term s (PBuiltinMap a b))
+newtype PMap (k :: PType) (v :: PType) (s :: S) = PMap (Term s (PBuiltinMap k v))
```
L-as, line 574: @t1lde why did you make a newtype for this BTW?

L-as (approved):
LGTM

PR #162, opened 2022-01-21
plam: Make functional dependency bidirectional

L-as:
This doesn't fix `plam $ \x -> x`, but it probably does
fix some other corner cases.


L-as:
#2

PR #163, opened 2022-01-21
PLift for PMap

L-as:


Plutarch/Api/V1.hs
```diff
@@ -75,12 +75,13 @@ import Plutarch.DataRepr (
   pmatchRepr,
  )
 import Plutarch.Integer (PInteger, PIntegral)
-import Plutarch.Lift (DerivePConstantViaNewtype (DerivePConstantViaNewtype), PConstant, PLifted, PUnsafeLiftDecl)
+import Plutarch.Lift (DerivePConstantViaNewtype (DerivePConstantViaNewtype), PConstant, PConstantRepr, PConstanted, PLift, PLifted, PUnsafeLiftDecl, pconstantFromRepr, pconstantToRepr)
```
TotallyNotChase, line 78: Formatting!

Plutarch/Api/V1.hs
```diff
@@ -75,12 +75,13 @@ import Plutarch.DataRepr (
...
 
 -- ctor in-scope for deriving
 import Plutarch.Prelude
 import qualified Plutus.V1.Ledger.Api as Plutus
 import qualified Plutus.V1.Ledger.Crypto as PlutusCrpyto
+import qualified PlutusTx.AssocMap as Map
```
TotallyNotChase, line 84: nitpick: should be named `PlutusMap` to be consistent with the previous two IMO

Plutarch/Api/V1.hs
```diff
@@ -574,12 +574,41 @@ instance PIsDataRepr PDCert where
...
+  , Plutus.FromData v
+  , Plutus.ToData v
+  , PConstant v
+  ) =>
+  PConstant (Map.Map k v)
+  where
+  type PConstantRepr (Map.Map k v) = [(Plutus.Data, Plutus.Data)]
+  type PConstanted (Map.Map k v) = PMap (PConstanted k) (PConstanted v)
+  pconstantToRepr m = (\(x, y) -> (Plutus.toData x, Plutus.toData y)) <$> Map.toList m
+  pconstantFromRepr :: [(Plutus.Data, Plutus.Data)] -> Maybe (Map.Map k v)
```
TotallyNotChase, line 606: is this type annotation intentional or an artifact of debugging :P

Plutarch/Api/V1.hs
```diff
@@ -574,12 +574,41 @@ instance PIsDataRepr PDCert where
...
+  pconstantFromRepr m = (Map.fromList <$>) $
```
TotallyNotChase, line 607: nitpick: I would argue `fmap Map.fromList $` is cleaner than `(Map.fromList <$>) $` but whatever

Plutarch/Builtin.hs
```diff
@@ -201,12 +201,7 @@ instance PIsData a => PIsData (PBuiltinList (PAsData a)) where
   pfromData x = punsafeCoerce $ pasList # pforgetData x
   pdata x = punsafeBuiltin PLC.ListData # x
 
-instance
-  ( PIsData k
```
TotallyNotChase, line 205: Not entirely sure if removing these constraints is a good idea. They are certainly redundant in this case. But it didn't make any sense to do `pdata` on a `PBuiltinMap k v` where `k` and `v` don't have `PIsData` impls.

Then again, if they don't have `PIsData` impls, you wouldn't be able to build the `PAsData k` and `PAsData v` required anyway. Which reminds me, what's the recommended way to build `PBuiltinMap`?

TotallyNotChase (changes requested):

Plutarch/Api/V1.hs
```diff
@@ -574,12 +574,41 @@ instance PIsDataRepr PDCert where
...
+  , Plutus.FromData (PLifted k)
+  , PLift k
+  , PLift v
+  ) =>
+  PUnsafeLiftDecl (PMap k v)
+  where
+  type PLifted (PMap k v) = Map.Map (PLifted k) (PLifted v)
+
+instance
+  ( PLifted (PConstanted k) ~ k
```
TotallyNotChase, line 590: Aha! I was so close. This equality constraint was what I needed to stop the endless constraint requirement issue I was facing.

Plutarch/Builtin.hs
L-as: A constant? MkPairData and MkCons? Something like that. There are situations where `PIsData k` is not fulfillable, even though it's in a `PAsData`, e.g. `PDataList`. In addition, reducing the constraints here also means reducing the number of constraints downstream.

L-as:
I think that's it.

TotallyNotChase (approved):
Awesome!

PR #164, opened 2022-01-21
Fix flake benchmark, use perf-diff instead of diff

emiflake:
This should use perf-diff in the effect code now. 
cc: @MatthewCroughan

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -450,11 +450,11 @@
         // {
           benchmark = {
             type = "app";
-            program = "${self.flake.${system}.packages."plutarch-benchmark:bench:perf"}/bin/perf";
+            program = "${self.flake.${system}.packages."plutarch-benchmark:bench:benchmark"}/bin/benchmark";
           };
-          perf-diff = {
+          benchmark-diff = {
             type = "app";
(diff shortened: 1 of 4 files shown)


L-as (dismissed):
LGTM

flake.nix
```diff
@@ -450,7 +450,11 @@
         // {
           benchmark = {
             type = "app";
-            program = "${self.flake.${system}.packages."plutarch:bench:perf"}/bin/perf";
+            program = "${self.flake.${system}.packages."plutarch-benchmark:bench:perf"}/bin/perf";
+          };
+          perf-diff = {
```
srid, line 455: Can we call this `benchmark-diff` for consistency?

emiflake: Yeah, should the `perf` and `perf-diff` binaries be renamed too?

MatthewCroughan: It would be good for `perf` to become `benchmark` and for `perf-diff` to become `benchmark-diff` in all cases. Once you've pushed this I'll test the branch and then merge it.

L-as (approved):
CI passes!

PR #165, opened 2022-01-21
Fix `nix run .#benchmark`

srid:
Just a quick fix to get CI building on `staging`.


srid:
Discard; I didn't see #164

PR #167, opened 2022-01-22
Document plutarch-benchmark library & diffing workflow

emiflake:
This also fixes the lack of `README.md` under `plutarch-benchmark` which would cause issues for projects depending on plutarch with haskell.nix :safety_pin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [Herc badge]: https://img.shields.io/badge/ci--by--hercules-green.svg
 [Herc link]: https://hercules-ci.com/github/Plutonomicon/plutarch
 [Cachix badge]: https://img.shields.io/badge/cachix-public--plutonomicon-blue.svg
-[Cachix link]: https://public-plutonomicon.cachix.org 
+[Cachix link]: https://public-plutonomicon.cachix.org
 
 Plutarch is a typed eDSL in Haskell for writing efficient Plutus Core validators.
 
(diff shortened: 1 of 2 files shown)


emiflake:
Ah looks like this doesn't cause any of the effects to run on the PR, which is the workflow I think we're wanting to facilitate (i.e. running them _before_ they get merged)
@matthewcroughan is there a reason we can't ease the `runIf` restriction? Maybe run if `staging` is strictly a subset of `HEAD` or something the like.

MatthewCroughan:
@emiflake Because Hercules/Nix doesn't give us any information about the merge-base. Which means we have to **for now** assume it's `staging`. Hercules also does not currently allow running CI against forked branches, only those that are part of the repo. That makes sense, because why would I want to allow you to run arbitrary code on the CI machine. If I allow you to do it, it means any anonymous user on Github can also run arbitrary code on the CI machine. It's a simple security measure.

emiflake:
> If it allowed you to do it, it means any anonymous user on Github can also run arbitrary code on the CI machine, not just members of this organization. It's a simple security measure.

Yeah this is not really what I'm asking about in the comment above. I'm okay with that limitation already, I'm more so asking as to why the effect doesn't run on a PR (or on _all_ non-fork commits, that would be fine too).

In the future, once hercules adds the button that starts it on fork PRs, then those with that access can execute it to see if there were no major regressions in the benchmarks.

MatthewCroughan:
@emiflake Yes. Although if any maintainer that has push access to this github organization wants to test any PR, they can simply run this and they'll be able to test the PR in the same way you described. It's just not a button on the GitHub interface.

```
# get $rev from tip of branch from fork
git fetch origin $rev # This works even if it's on a fork
git push origin "$rev:refs/heads/tmpbranch"
git push origin :tmpbranch
```

emiflake:
> ```
> # get $rev from tip of branch from fork
> git fetch origin $rev # This works even if it's on a fork
> git push origin "$rev:refs/heads/tmpbranch"
> git push origin :tmpbranch
> ```

Oh neat, good to know, for sure. But this still wouldn't run the `benchmark-diff` effect, since it's not on staging. _That's_ what I'm asking about.

MatthewCroughan:
@emiflake When making effects, I make it without the if statement to test that it works. Then, when I'm sure it works as expected, I add the if statement.

L-as (approved):
LGTM

PR #168, opened 2022-01-24
plam: Make inference work for `id` function

### LINKED ISSUE #2: Improve type inference of plam

L-as:
Fixes #2

### DIFF
--- a/Plutarch.hs
+++ b/Plutarch.hs
@@ -101,25 +101,28 @@ infixr 0 #$
  > const = plam (\x y -> x)
 -}
 
-class PLamN a b | a -> b, b -> a where
-  plam :: a -> b
+class PLamN a b s | a -> b where
+  plam :: forall c. (Term s c -> a) -> Term s (c :--> b)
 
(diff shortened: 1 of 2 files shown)


TotallyNotChase:
This probably broke long ago and I missed it but the following doesn't work, even though it sure as hell worked before-
```hs
pf :: Term s (b :--> a :--> a)
pf = plam $ \_ y -> y
```
It gives
```
Overlapping instances for Plutarch.PLamN
  (b'2 -> b'2) (a :--> a) s
arising from a use of ‘plam’
```

L-as:
This broke when we switched away from incoherent instances unfortunately.

TotallyNotChase:
I see, somewhat concerning that neither of these work though-
```hs
pf :: Term s (b :--> a :--> a)
pf = plam $ \_ y -> y

pf' :: Term s (POpaque :--> a :--> a)
pf' = plam $ \(_ :: Term _ _) y -> y

pf'' :: Term s (b :--> a :--> a)
pf'' = plam $ flip const
```

TotallyNotChase:
On the bright side, I did actually notice this improving type inference in quite a few places.

L-as:
OK, so it's a step forward, even if it doesn't fix #2

PR #169, opened 2022-01-24
Efficiently binding data fields to a HRec #95

t1lde:
Re-opening #195

I didn't merge in @srid's stuff yet (ought to be a separate PR I think?)


Plutarch/DataRepr/Internal.hs
```diff
@@ -0,0 +1,176 @@
...
+  PLabel '[] = '[]
+  PLabel ((l ':= t) ': as) = l ': (PLabel as)
+
+type PDataSum :: [[PLabeledType]] -> PType
+data PDataSum (defs :: [[PLabeledType]]) (s :: S)
+
+pasData :: Term s (PDataSum _) -> Term s PData
+pasData = punsafeCoerce
+
+punDataRepr :: Term s (PDataSum '[def] :--> PDataRecord def)
```
L-as, line 73: Naming?

Plutarch/DataRepr/Internal.hs
```diff
@@ -0,0 +1,176 @@
...
+punDataRepr = phoistAcyclic $
+  plam $ \t ->
+    plet (pasConstr #$ pasData t) $ \d ->
+      (punsafeCoerce $ psndBuiltin # d :: Term _ (PDataRecord def))
+
+pindexDataRepr :: (KnownNat n) => Proxy n -> Term s (PDataSum (def : defs) :--> PDataRecord (IndexList n (def : defs)))
```
L-as, line 79: Dead code?

Plutarch/Api/V1.hs
```diff
@@ -165,12 +178,12 @@ instance PIsDataRepr PScriptContext where
...
-  | PSpending (Term s (PDataList '[PTxOutRef]))
-  | PRewarding (Term s (PDataList '[PStakingCredential]))
-  | PCertifying (Term s (PDataList '[PDCert]))
+  = PMinting (Term s (PDataRecord '["_0" ':= PCurrencySymbol]))
+  | PSpending (Term s (PDataRecord '["_0" ':= PTxOutRef]))
+  | PRewarding (Term s (PDataRecord '["_0" ':= PStakingCredential]))
+  | PCertifying (Term s (PDataRecord '["_0" ':= PDCert]))
   deriving
-    (PMatch, PIsData)
+    (PMatch, PIsData, PDataFields)
```
L-as, line 186: This instance is bogus I think.

Plutarch/DataRepr/Field.hs
```diff
@@ -0,0 +1,231 @@
...
+{- |
+  Bind a HRec of named fields from a compatible type.
+-}
+pletFields ::
+  forall a b s.
+  ( PDataFields a
+  , BindFields (PFields a)
+  ) =>
+  Term s a ->
+  (HRec (PLabel (PFields a)) (TermsOf s (PUnLabel (PFields a))) -> Term s b) ->
```
L-as, line 111: The reason why I recommended merging the type parameters is that this type is a bit scary for users.

L-as (approved):
LGTM

L-as:
I'm merging this so others can work on top of it easilier.

Plutarch/DataRepr/Field.hs
t1lde: Yeah, and I think it would make it easier to add good type-errors

Plutarch/DataRepr/Internal.hs
t1lde: I think it should be exported, incase there's a use for it, though most of the time pmatch would work instead

Plutarch/Api/V1.hs
t1lde: Ah! yeah that's correct this time

t1lde: I did the same for all the sum types without thinking... :facepalm:

Plutarch/DataRepr/Internal.hs
L-as: It has the same function as `pindexDataRecord` though?

PR #170, opened 2022-01-24
Really fix #2

### LINKED ISSUE #2: Improve type inference of plam

L-as:
Removing the extra instances made type inference work better...

This should also allow many more arguments with `plam`.


L-as:
The previous commit that added `   plam :: forall c. (Term s c -> a) -> Term s (c :--> b)`  seems to have been what allowed this simplification.

L-as:
The thing this changes specifically, is that all arguments are easily inferred to have type `Term _ _`.

Issue #135, opened 2022-01-18
Generic deriving


srid:
This is a meta issue for generic deriving of various Plutarch classes.

## Typeclasses to derive 

- [x] `PIsDataRepr` (Data for non-Scott encoded types) #173 
- [x] `PlutusType` for Scott encoded types: https://github.com/Plutonomicon/plutarch/pull/189
- [x] `PlutusType` for Data: https://github.com/Plutonomicon/plutarch/issues/187
- [ ] ... what else?

L-as:
We will probably still need something separate for newtypes.

TotallyNotChase:
are we missing anything here?

L-as:
Nope! Thanks to both @srid and @TotallyNotChase !

srid:
I have an idea in mind (Data-encoded types using normal sum/records), but that can wait I suppose. What we have right now should be sufficient.

L-as:
That can by definition not be efficient currently without punsafeLetPure and such

PR #173, opened 2022-01-24
Add generic deriving for `PIsDataRepr`

srid:
#135


srid:
Should `Plutarch.Generic` and `Plutarch.DataRepr.Internal.HList.hs` be moved under `./Plutarch/Util` - because they are independent of the Plutarch itself?

Plutarch/DataRepr/Internal.hs
```diff
@@ -158,7 +180,48 @@ newtype PIsDataReprInstances (a :: PType) (s :: S) = PIsDataReprInstances (a s)
...
+    , PDataRecordFields2 code ~ PIsDataReprRepr a
+    , MkDataReprHandler s a 0 code
+    ) =>
+    (a s -> Term s out) ->
+    DataReprHandlers out (PIsDataReprRepr a) s
+  pmatchDataReprHandlers =
+    mkDataReprHandler @s @a @0 @code
+
+pmatchRepr :: forall s a b. PIsDataRepr a => Term s (PDataSum (PIsDataReprRepr a)) -> (a s -> Term s b) -> Term s b
+pmatchRepr dat = pmatchDataRepr dat . pmatchDataReprHandlers @a @s @b
```
srid, line 198: I should undo this.

Plutarch/Generic.hs
```diff
@@ -0,0 +1,36 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Common generics-sop utilities for use in Plutarch.
+module Plutarch.Generic (
```
L-as, line 5: i don't see why this code should be exported. Can you move this into the module where you use it or name it something with `Internal` in it?

L-as (changes requested):
Looks great, thanks!

Plutarch/Generic.hs
srid: It may come in handy for further generics deriving work (PlutusType, etc.), so I figured it is worth having a common module. Okay, I've moved it to `Plutarch.DataRepr.Internal.Generic`.

L-as (approved):
LGTM

PR #174, opened 2022-01-24
Re-add `AllowAmbiguousTypes` to `Plutarch.hs`

TotallyNotChase:
This removal was causing errors on projects using ghc 8.10.7.


L-as (approved):
LGTM

PR #175, opened 2022-01-24
Mario/more records

blamario:
Fixed a couple of bugs, exported more utility functions, documentation, many more examples.

### DIFF
--- a/examples/Examples/LetRec.hs
+++ b/examples/Examples/LetRec.hs
@@ -2,14 +2,14 @@
 
 module Examples.LetRec (tests) where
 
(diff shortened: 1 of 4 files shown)


examples/Examples/LetRec.hs
```diff
@@ -1,32 +1,52 @@
+{-# LANGUAGE RecordWildCards #-}
```
L-as, line 1: Please no wild cards.

blamario: Removed.

L-as (approved):
LGTM

PR #176, opened 2022-01-24
Simplification and cleanup of HRec

t1lde:
I tried to get nicer type-errors to work the oddities of type-families were giving me trouble, so no luck.


L-as (approved):
LGTM

PR #178, opened 2022-01-24
flake: always run benchmark-diff effect

emiflake:
Summary: Run benchmark-diff effect on _all commits_ instead of only those on staging.
Essentially, this change is necessary to make the workflow work as intended by https://github.com/Plutonomicon/plutarch/issues/123#issuecomment-1020583755.

Now all future PRs (like this one) will actually run the effects and show them in the checks.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -516,6 +516,11 @@
           hci-effects = hercules-ci-effects.lib.withPkgs pkgs;
         in
         {
+          # Hercules 0.9 will allow us to calculate the merge-base so we can test all PRs.
+          # Right now we just hardcode this effect to test every commit against
+          # origin/staging. We set != "refs/head/master" so that merges into master don't
+          # cause a lot of unnecessary bogus benchmarks to appear in CI for the time
+          # being.
           benchmark-diff = hci-effects.runIf (src.ref != "refs/heads/master") (
             hci-effects.mkEffect {
               src = self;
--- a/flake.nix
+++ b/flake.nix
@@ -516,7 +516,7 @@
           hci-effects = hercules-ci-effects.lib.withPkgs pkgs;
         in
         {
-          benchmark-diff = hci-effects.runIf (src.ref == "refs/heads/staging") (
+          benchmark-diff = hci-effects.runIf (src.ref != "refs/heads/master") (
             hci-effects.mkEffect {
               src = self;
               buildInputs = with pkgs; [ git nixFlakes ];


srid (dismissed):

MatthewCroughan (changes requested):
This compares everything against staging. This means that if I make a PR to Master, it will actually lie and mislead me, and make me think it's comparing the performance of my PR, which it would not be. I do not think this should be merged.

emiflake:
> This compares everything against staging. This means that if I make a PR to Master, it will actually lie and mislead me, and make me think it's comparing the performance of my PR, which it would not be. I do not think this should be merged.

Fair point, I understand the concern. How's this, then. We'll do it when not on master. And assume all PRs follow the staging workflow and are supersets of staging.

MatthewCroughan:
I don't see the point of doing that. Simply make a PR targeting staging, and you'll get a benchmark. There is no other purpose for this effect. All other data is not relevant. We do not have access to the merge-base via Nix until Hercules-CI 0.9 is released. Once that is released, we can do this for any branch, until then we cannot.

srid:
> Simply make a PR targeting staging, and you'll get a benchmark

You will not get a benchmark *diff* (against staging). See #173 for an example. And that's what this PR fixes.

emiflake:
> Simply make a PR targeting staging, and you'll get a benchmark.

Except, sadly that's not true. See https://github.com/Plutonomicon/plutarch/pull/173/commits/bfd237b5f74ea471446738d3bedfab7bb8b71067. This is on a PR for #173.

image
image

MatthewCroughan:
Ah, okay. That makes sense. Adding the if statement was a mistake from me. It is the case that it will not run until it is merged, which is not correct.

emiflake:
> Ah, okay. That makes sense. Adding the if statement was a mistake from me. It is the case that it will not run until it is merged, which is not correct.

Yeah, although your concern is still valid. And this still will sort of be misleading in cases where it runs on branches off of master. Though sadly I don't think we can check if "HEAD came after staging" in a `runIf`, can we?

flake.nix
```diff
@@ -516,7 +516,7 @@
           hci-effects = hercules-ci-effects.lib.withPkgs pkgs;
         in
         {
-          benchmark-diff = hci-effects.runIf (src.ref == "refs/heads/staging") (
+          benchmark-diff = hci-effects.runIf (src.ref != "refs/heads/master") (
```
MatthewCroughan, line 519:
```suggestion
          # Hercules 0.9 will allow us to calculate the merge-base so we can test all PRs.
          # Right now we just hardcode this effect to test every commit against
          # origin/staging. We set != "refs/head/master" so that merges into master don't
          # cause a lot of unnecessary bogus benchmarks to appear in CI for the time
          # being.
          benchmark-diff = hci-effects.runIf (src.ref != "refs/heads/master") (
```

MatthewCroughan (changes requested):

srid:
Oh I hadn't noticed that the PR diff changed.

flake.nix
MatthewCroughan: Merge and flatten this, and I think this is a fine approach and we can merge it.

srid: I was wondering about that check, and this comment makes it clear to me. 👍🏿

emiflake: Perfect!

srid (approved):

MatthewCroughan (approved):

PR #179, opened 2022-01-24
Remove redundant constraint

srid:

PR #180, opened 2022-01-25
PRational: add PConstant,etc. instances

### LINKED ISSUE #166: Derive Lift for `Plutarch.Rational`
I noticed that this module has no instance for `PConstant`, etc., but it probably should - to be able to build it from Haskell values using `pconstant` (it is useful to do during testing).

/cc @Geometer1729 who tried doing this and faced some issues.

srid:
Resolves #166 

Includes #179

### DIFF
--- a/Plutarch/DataRepr/Internal.hs
+++ b/Plutarch/DataRepr/Internal.hs
@@ -241,7 +241,7 @@ instance PIsDataRepr a => PMatch (PIsDataReprInstances a) where
 
 newtype DerivePConstantViaData (h :: Type) (p :: PType) = DerivePConstantViaData h
 
-instance (PIsDataRepr p, PLift p, Ledger.FromData h, Ledger.ToData h) => PConstant (DerivePConstantViaData h p) where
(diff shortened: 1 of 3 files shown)


L-as:
I'm pretty sure your instance is bonus. The constraint isn't redundant either, it's there to ensure soundness. A rational is encoded as a function, yet `pconstant` will give you Data and say it's a rational. I am 99% sure this is wrong.

L-as:
s/bonus/bogus

L-as:
It is not possible to implement PLift for PRational. It can not be represented as a constant.

srid:
Okay - I’ll take a look at this tomorrow.

There definitely is some confusion around the types and instances, which I should clarify first.

srid:
Confirmed that you are right. As we use `punsafeConstantInternal` (note: unsafe), the internal representation is actually a Data, which mismatches with the Scott-encoded representation given by the PlutusType instance. 

As an aside, it sucks that Haskell cannot enforce it. The whole 'redundant _but_ necessary' constraint thing (ie., the `PLift p` in `PConstant (DerivePConstantViaNewtype h p p')`) feels like a hack, that should in theory be enforced by the compiler a la. make-illegal-states-unrepresentable.

Issue #177, opened 2022-01-24
More tests of compiled output


L-as:
There are too few tests for the compiled output. We need **a lot** more.

L-as:
`printTerm x @?= "something"`

srid:
With golden testing in place, this is now straight forward to do -- just add the Plutarch programs with `pgoldenSpec` and `@|`:

https://github.com/Plutonomicon/plutarch/blob/9c163c80b3df7eac3b6d2e5ab936b36584ce5316/plutarch-test/src/Plutarch/ByteStringSpec.hs#L11-L17

PR #181, opened 2022-01-25
More exports to prelude + more instances for `DerivePNewtype` + minor cleanup

### LINKED ISSUE #28: Stabilize module exports (more exports from prelude, remove unsafe exports from `Plutarch` etc)

TotallyNotChase:
* Let `PEq` and `POrd` be derived via `DerivePNewtype`
* Add `PEq` and `POrd` instances to relevant ledger api types
* Rename `DerivePConstantViaCoercible` -> `DerivePConstantDirect` and simplify it
* Add more exports to `Plutarch.Prelude` - fixes #28 
* Separate out `punsafe*` functions to `Plutarch.Unsafe`


Plutarch/Builtin.hs
```diff
@@ -245,7 +265,7 @@ instance PEq (PAsData a) where
   x #== y = punsafeBuiltin PLC.EqualsData # x # y
 
 instance (forall (s :: S). Coercible (a s) (Term s b), PIsData b) => PIsData (DerivePNewtype a b) where
-  pfromData x = pcon . DerivePNewtype $ ptypeOuter target
+  pfromData x = punsafeFrom target
```
TotallyNotChase, line 268: I changed this to use `punsafeFrom`, I assume this should be a safe usage here?

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
+  pfoldrLazy,
+  pfoldl,
+  pall,
+  pany,
+
+  -- * Scott encoded list type
+  PList (..),
+
+  -- * Scott encoded maybe type and utilities
+  PMaybe (..),
```
TotallyNotChase, line 88: These should probably be renamed before being added to prelude. Otherwise importing the ledger api becomes a bit painful.

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
+  psndBuiltin,
+  PBuiltinPair,
+  PBuiltinList (..),
+  PIsData (pfromData, pdata),
+  PAsData,
+  ppairDataBuiltin,
+  PBuiltinMap,
+
+  -- * DataRepr and related functions
+  PDataFields,
```
TotallyNotChase, line 108: I was unsure whether or not the DataRepr stuff should be added to prelude. I think they should be - but I don't know if these exports are all that's needed or if there are more commonly used functions that should also be exported here.

Plutarch/Api/V1.hs
```diff
@@ -173,46 +162,55 @@ deriving via (DerivePConstantViaData Plutus.ScriptPurpose PScriptPurpose) instan
...
-deriving via (DerivePConstantViaNewtype Plutus.ValidatorHash PValidatorHash PByteString) instance (PConstant Plutus.ValidatorHash)
+deriving via
+  (DerivePConstantViaNewtype Plutus.ValidatorHash PValidatorHash PByteString)
+  instance
+    (PConstant Plutus.ValidatorHash)
 
 ---------- Value
 
 newtype PTokenName (s :: S) = PTokenName (Term s PByteString)
   deriving newtype (Semigroup, Monoid)
```
TotallyNotChase, line 212: Why does this have a `Semigroup` and `Monoid` instance but the others don't btw?

Plutarch/Builtin.hs
L-as: why replace something safe with something unsafe?

Plutarch/Prelude.hs
L-as: The issue is you'd have to add all the other stuff, like generics, too. Maybe it's worth it.

Plutarch/Api/V1.hs
L-as: This doesn't even make sense. Token names have a limit of 32 bytes. I propose removing the instance because it's almost bogus.

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
+  -- * Cryptographic hashes and signatures
+  psha2_256,
+  psha3_256,
+  pverifySignature,
+
+  -- * Converstion between Plutarch terms and Haskell types
+  pconstant,
+  plift,
+  plift',
+  LiftError,
```
L-as, line 130: We don't need this.

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
+  ppairDataBuiltin,
+  PBuiltinMap,
+
+  -- * DataRepr and related functions
+  PDataFields,
+  PDataRecord,
+  PDataSum,
+  PIsDataRepr,
+  pfield,
+  pletFields,
```
L-as, line 113: We should add the other ones too.

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
   plet,
   pinl,
-  pcon,
-  pmatch,
   pto,
   pfix,
   Type,
   S,
   PType,
+  PlutusType,
```
L-as, line 22: `pcon'` and `pmatch'` will be necessary for implementing `PlutusType`. Same for `PInner`.

Plutarch/Builtin.hs
TotallyNotChase: eh, mostly to remove the boilerplate a bit. I *assumed* they both have the same meaning in this case - but of course, I'm not entirely sure. Should this be changed back?

Plutarch/Prelude.hs
TotallyNotChase: Well, maybe. But is that the compromise we're settling on regarding this scott-encoding/data (/builtin) encoding ambiguity issue?

Plutarch/Prelude.hs
TotallyNotChase: Which imports, in particular, are needed in association with this? You don't need `GHC.Generics.Generic` to *use* data repr stuff right? Just to derive it for your own type.

Plutarch/Api/V1.hs
TotallyNotChase: I agree with that. Removed.

I noticed that the plutus ledger api also doesn't implement semigroup/monoid for most of these (if any) types.

Plutarch/Api/V1.hs
```diff
@@ -254,30 +251,30 @@ deriving via
...
   instance
-    (PConstant PlutusCrpyto.Signature)
+    (PConstant PlutusCrypto.Signature)
 
 ---------- Time
 
 newtype PPOSIXTime (s :: S)
   = PPOSIXTime (Term s PInteger)
   deriving (PIntegral) via (PInteger)
   deriving newtype (Num)
```
TotallyNotChase, line 276: Is this a valid instance? It's for `Num (PPOSIXTime s)` rather than `Num (Term s PPOSIXTime)`.

Plutarch/Prelude.hs
L-as: We don't need to export `PDataFields` if you do not intend to allow defining data types with just the prelude.

Plutarch/Prelude.hs
```diff
@@ -13,17 +14,143 @@ module Plutarch.Prelude (
...
+
+  -- * Builtin types and utilities
+  PData (..),
+  pfstBuiltin,
+  psndBuiltin,
+  PBuiltinPair,
+  PBuiltinList (..),
+  PIsData (pfromData, pdata),
+  PAsData,
+  ppairDataBuiltin,
```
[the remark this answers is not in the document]
TotallyNotChase: I haven't found a use for this in my experience so far. But this *is* the only way to build a `BuiltinPair (PAsData a) (PAsData b)`. That would probably only be useful when building `PMap`s though - so maybe it shouldn't exist in prelude.

Plutarch/Prelude.hs
TotallyNotChase: I assume the `LiftError` export should be removed *alongside* `plift'` then.

Plutarch/Prelude.hs
TotallyNotChase: Can you give me a quick rundown on the field accessor methods we want to encourage in ideal situations?

Plutarch/Api/V1.hs
L-as: It's valid, but it's certainly not useful..

Plutarch/Prelude.hs
L-as: Yeah.

Plutarch/Prelude.hs
L-as: We want to use the most specific function essentially. If we don't want all the fields, we shouldn't use `pletFields`, and it is very very rare that we want to use all the fields. We probably want a more fine `pletFields` too.

TotallyNotChase: Like `pletNFields`? That could be useful more often yeah?

I was under the impression `pletFields` *lazily* binds fields. I suppose that's not the case, though I haven't looked into it much yet.

In any case, we want to encourage `pfield` for sure, correct?

Plutarch/Prelude.hs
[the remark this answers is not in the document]
L-as: I don't think this needs to be in the prelude.

Plutarch/Prelude.hs
L-as: I haven't inspected the generated code. It's something we need to ascertain.

L-as: #177

Plutarch/Api/V1.hs
TotallyNotChase: I wonder if `DerivePNewtype` should be able to derive `Num`, `Semigroup`, and `Monoid` then (over actual terms, of course). Would that be useful?

L-as: It would definitely be.

L-as: We should also add PlutusType instances.

Plutarch/Prelude.hs
t1lde: the doc comments in `Example.Field` give a decent rundown

- `pletFields` if (and only if) you want to use all fields
- `pletNFields @n` if you want the 0th field, the nth field and any amount of other fields between
- `pletRangeFields @n @m` if you want the nth field, the mth field and any amount of other fields between
- and `pfield @n` (or a dot accessor if you already have a `PDataRecord` from `pmatch`) for a single field

L-as: I don't think it's possible for it to be lazy. `plet` is strict after all, and it's not possible for it to know that it can eliminate the variable if it's not used, not to mention that currently it doesn't even *know* when the variable is not used. I think the best solution is to create a variant of `pletFields` that takes a type-level list of fields we want to access.

Plutarch/Prelude.hs
TotallyNotChase: Removed.

Plutarch/Prelude.hs
TotallyNotChase: Yes, for sure - added those.

Plutarch/Prelude.hs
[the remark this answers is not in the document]
TotallyNotChase: Agree, removed.

Plutarch/Prelude.hs
L-as: #181

Plutarch/Prelude.hs
TotallyNotChase: Removed `PDataFields` export.

Plutarch/Prelude.hs
L-as: I will make a PR after this to change the interface, perhaps without making it efficient. We should have just this in the prelude.

TotallyNotChase: I see. For the time being, these are the `DataRepr` related exports I'm thinking of-
```hs
PDataRecord,
PDataSum,
PIsDataRepr,
pfield,
pletRangeFields,
pindexDataRecord,
```
Let me know if any modifications should be made there.

t1lde: >  it is very very rare that we want to use all the fields.

I'd say it's decently common. ScriptContext, for example.

TotallyNotChase: Heh, race conditions - I didn't see you guys' comments! I'll wait for Las or Tilde's PR for a streamlined interface and add that to prelude.

Plutarch/Api/V1.hs
TotallyNotChase: We have `PlutusType` and `PIsData` already, as well as `PEq` and `POrd`.

Plutarch/Prelude.hs
L-as: I don't think you should wait! We'll just change `pletFields` in another PR.

t1lde: The fields will be bound *somewhat* lazily - each list tail is `plet` bound, but the fields (the heads) themselves are not. So the `phead x` terms will not be present in the compiled output unless they are used.

Plutarch/Builtin.hs
L-as: Yeah they do, this is fine.

L-as (approved):
LGTM

L-as:
@TotallyNotChase this is ready for merging I assume?

TotallyNotChase:
> @TotallyNotChase this is ready for merging I assume?
Almost, going to add the `Semigroup`/`Monoid`/`Num` instances to `DerivePNewtype` - I have the code laying around anyway.

TotallyNotChase:
I added the instances, worth reviewing https://github.com/Plutonomicon/plutarch/pull/181/commits/82a25a7c54605b72c2dde1f18ac61f7732b3bc4e independently.

L-as (approved):
LGTM

Issue #182, opened 2022-01-25
`pletFields` should take type-level list of fields to bind


L-as:
https://github.com/Plutonomicon/plutarch/pull/181#discussion_r791637577

This would make it the interface efficient as it can be AFAICT.

t1lde:
I think I have an idea of how to do this, let me give it a go

L-as:
@t1lde One (perhaps unfortunate) thing to take into account is that exposing the fields with `HasField` doesn't give warnings for when you don't use the fields. It's perhaps best to give the user the `HRec` and have them pattern match on it completely as `Labeled @"field" field`, which albeit unfortunate, AFAICT is the only way to give proper warnings for unused fields.

t1lde:
That doesn't seem worth it to me, As long as the docs make it clear how to use it efficiently IMO convenience overrules

L-as:
Sure.

L-as:
It would be amazing if GHC could somehow infer the type level list from how we use it, but I don't think that's possible.

PR #185, opened 2022-01-25
More efficient pletFields with list of fields

t1lde:
Addresses #182

Still need to make sure this is back-compatible, there's some polykind weirdness I'm unsure of


t1lde:
(waiting on local build so I can fix after the merge)

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -186,6 +226,21 @@ type family TermsOf (s :: S) (as :: [PLabeledType]) :: [Type] where
...
+  LabelIndex name ((Labeled name _) ': _) = 0
+  LabelIndex name (_ ': as) = (LabelIndex name as) + 1
+
+type AllIndices :: [Symbol] -> [k] -> [Nat]
+type family AllIndices (ns :: [Symbol]) (as :: [k]) :: [Nat] where
+  AllIndices '[] _ = '[]
+  AllIndices (n ': ns) as = (LabelIndex n as) ': (AllIndices ns as)
+
+type family FieldsRange (ns :: [Symbol]) (as :: [k]) :: (Nat, Nat) where
+  FieldsRange ns as = FindMinMax (AllIndices ns as)
```
t1lde, line 242: There's now *a lot* of type-family evaluation going on... with the nesting, and list-recursion the type errors are pretty crazy

Plutarch/DataRepr/Internal/HList.hs
```diff
@@ -100,6 +106,19 @@ type family Drop (n :: Nat) (as :: [k]) :: [k] where
 type family Range (from :: Nat) (to :: Nat) (as :: [k]) :: [k] where
   Range from to xs = Take (to - from + 1) (Drop from xs)
 
+type family Min (a :: Nat) (b :: Nat) :: Nat where
+  Min a b = If (a <=? b) a b
+
+type family Max (a :: Nat) (b :: Nat) :: Nat where
+  Max a b = If (a <=? b) b a
```
t1lde, line 113: latest version of `base` has polykinded `Min` and `Max` that could be used if not for backcompat

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -103,89 +107,55 @@ instance
...
+  fields.
+  This will bind the hull of the subset of fields provided,
 
-{- | Bind a HRec, dropping the first N fields.
+  for example, suppose that:
+  @PFields x ~ '["x" ':= x, "y" ':= y, "z" ':= z, "w" ':= w]@,
 
-  Usually more efficient than binding all fields.
+  then:
+  @pletFields @["y", "w"]@ will bind the fields @'["y", "z", "w"]@.
```
L-as, line 118: This doesn't make sense. Why is `z` bound?

L-as (changes requested):

Plutarch/DataRepr/Internal/Field.hs
t1lde: currently the field binding doesn't do any skipping, you're right that we can do better

t1lde: I'll try and get this working

L-as: BTW it's OK if the implementation doesn't make use of the fact that you only need some of the fields, since we can just fix that later in a non-breaking change.

examples/Examples/Field.hs
```diff
@@ -304,4 +304,4 @@ tripYZComp :: String
 tripYZComp = "(program 1.0.0 (\\i0 -> (\\i0 -> addInteger (unIData (force headList i1)) (unIData (force headList (force tailList i1)))) (force tailList ((\\i0 -> force (force sndPair) (unConstrData i1)) i1))))"
 
 letSomeFieldsComp :: String
-letSomeFieldsComp = "(program 1.0.0 (\\i0 -> (\\i0 -> (\\i0 -> (\\i0 -> (\\i0 -> addInteger (addInteger (unIData (force headList i4)) (unIData (force headList i3))) (unIData (force headList (force tailList i1)))) (force tailList i1)) (force tailList i1)) (force tailList i1)) ((\\i0 -> force tailList (force tailList (force tailList i1))) i1)))"
+letSomeFieldsComp = "(program 1.0.0 (\\i0 -> (\\i0 -> (\\i0 -> addInteger (addInteger (unIData (force headList i2)) (unIData (force headList i1))) (unIData (force headList (force tailList (force tailList (force tailList i1)))))) (force tailList i1)) (force tailList (force tailList (force tailList i1)))))"
```
L-as, line 307: I just noticed that there is an inefficiency that can't be solved easily, but we can make it better. The `force headList x` part should be plet'ed.

L-as: It's still not perfect, because if you use it only once it should be inlined. There's not really a good way of doing that, but it's better that it be a plet IMO. It's very unintuitive for the field access to be expensive IMO. Maybe there's some other way we can make users `plet` it themselves.

L-as (approved):
Looks great, do you think this is solvable? Will merge since it doesn't affect the API.

L-as:
>  Could you please add hrecField to Plutarch.Prelude now that pletFields is encouraged?

it's encouraged now!

examples/Examples/Field.hs
TotallyNotChase: This could be somewhat problematic since, very often - people use a field only once. In which case, this would be a rather unnecessary `plet`. But I absolutely understand the concern.

TotallyNotChase: On a related note: what's up with so much builtin duplication in that compiled output? Is that missing hoisting?

t1lde: It would be a simple change to `plet` bind all the `force headList x` terms by default. The question is whether we want that.

I think it's acceptable enough for users to have to know to plet bind the terms if they're going to use them more than once, aside from changes to hoisting I'm not sure if we can have it work automagically

L-as: We are in fact missing some `phoistAcyclic` here! Good catch. #177

L-as: This is very odd, I can't find what's not hoisted.

L-as: #195

TotallyNotChase: It's ok - I'll be testing these parts pretty extensively in the coming days, hopefully we'll be able to catch everything one by one.

PR #189, opened 2022-01-25
Generic deriving for PlutusType (Scott encoding)

srid:
- [x] type
- [x] pcon
    - [x] `PDelayed` for constructors with no args
      - [x] #193 
- [x] pmatch
- [x] Handle single-constructor case ([ctx](https://github.com/Plutonomicon/plutarch/pull/189#issuecomment-1022064194))
- [x] Refactor
    - [x] `GPCon` undefined
    - [x] Reduce type families


L-as:
One thing to note is that the handler for a constructor with no arguments must be translated into `PDelayed a`.

L-as:
AFAICT you're not doing that here.

L-as:
Another thing to note is that we can eta-reduce it when there is only one constructor with one field, making it a newtype essentially as done with `DerivePNewtype`. Users *can* manually derive it, albeit it would be nice if it could either err in that case, telling the user to use `DerivePNewtype`, or if it could do it all automatically.

TotallyNotChase:
May I propose separating out `PlutusType` to a separate module (and re-exporting from `Plutarch.hs` - of course). That is a lot of machinery to put into `Plutarch.hs`.

srid:
I've added comments on current type-families. 

I opened it as a draft upon Las' request in Discord, but note that PR is still in flux, and nothing is finalized - but they should be after I implement generic `pmatch`.

srid:
> May I propose separating out `PlutusType` to a separate module (and re-exporting from `Plutarch.hs` - of course). That is a lot of machinery to put into `Plutarch.hs`.

In my opinion, yes. We would also have to create a `PLam.hs` to avoid cyclic imports.

srid:
> @L-as Another thing to note is that we can eta-reduce it when there is only one constructor with one field, making it a newtype essentially as done with `DerivePNewtype`. Users *can* manually derive it, albeit it would be nice if it could either err in that case, telling the user to use `DerivePNewtype`, or if it could do it all automatically.

What about one constructor with >1 products? Such as `PPair`, defined as:

```haskell
data PPair (a :: PType) (b :: PType) (s :: S) = PPair (Term s a) (Term s b)
```

L-as:
> > @L-as Another thing to note is that we can eta-reduce it when there is only one constructor with one field, making it a newtype essentially as done with `DerivePNewtype`. Users _can_ manually derive it, albeit it would be nice if it could either err in that case, telling the user to use `DerivePNewtype`, or if it could do it all automatically.
> 
> What about one constructor with >1 products? Such as `PPair`, defined as:
> 
> ```haskell
> data PPair (a :: PType) (b :: PType) (s :: S) = PPair (Term s a) (Term s b)
> ```
> 
> EDIT: Also,
> 
> ```haskell
> data PProxy a s = PProxy
> ```

The first one becomes `forall c. (a -> b -> c) -> c`.

The second one should optimally be `forall a. a -> a`, but with the current logic becomes `forall a. PDelayed a -> a`. You could disallow it TBH.

L-as:
This is really great BTW.

Plutarch/PLam.hs
```diff
@@ -0,0 +1,61 @@
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Plutarch.PLam (
```
L-as, line 4: Do you intend for this module to be exposed to end-users?

L-as (approved):
LGTM! Really nice.

Plutarch/PlutusType.hs
```diff
@@ -0,0 +1,335 @@
...
+
+  The 'PlutusType' class allows encoding Haskell data-types as plutus terms
+  via constructors and destructors.
+
+  A simple example, encoding a Sum type as an Enum via PInteger:
+
+  > data AB (s :: S) = A | B
+  >
+  > instance PlutusType AB where
+  >   type PInner AB _ = PInteger
```
L-as, line 46: We probably want to fix this documentation.

Plutarch/PLam.hs
srid: Nope; should it be put under `Plutarch.Internal.PLam`? 

(We also have *another* Internal, at `Plutarch.DataRepr.Internal`, which can be a bit confusing).

srid:
Ready for review now.

Plutarch/PLam.hs
L-as: We have a lot of internals! Maybe it should all be under `Plutarch/Internal`

Plutarch/PlutusType.hs
srid: I updated it to mention deriving for Scott encoding, assuming I didn't misinterpret what you meant here.

Plutarch/PLam.hs
srid: Moved to `Plutarch.Internal.PLam`. I didn't touch the DataRepr stuff because it is outside the scope of this PR.

srid: I think `PlutusType` should be made internal too? It is not meant to be exported directly either. We split these modules only because Plutarch.hs was growing to be big.

srid: > I think `PlutusType` should be made internal too? It is not meant to be exported directly either. We split these modules only because Plutarch.hs was growing to be big.

Done now.

L-as (approved):
LGTM, is there anything we're missing?

srid:
Can't think of any; I think it is good for merge.

PR #192, opened 2022-01-26
Fix missing imports during development mode

TotallyNotChase:
We need to have `cabal build -f development` in CI

### DIFF
--- a/Plutarch/Trace.hs
+++ b/Plutarch/Trace.hs
@@ -7,7 +7,7 @@ module Plutarch.Trace (ptrace, ptraceIfTrue, ptraceIfFalse, ptraceError) where
 
 import Plutarch (Term, perror)
 #ifdef Development
-import Plutarch (punsafeBuiltin, type (:-->))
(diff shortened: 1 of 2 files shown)


TotallyNotChase:
Hold on - there's a failing test. I'll fix that.

TotallyNotChase:
Here's something subtle: Any function that uses tracing in its dependency chain (`ptraceError`, `ptrace` etc.), needs finer attention in the test suite if comparing script outputs directly. Of course, the `trace` builtin function call is only present in the script in development mode.

So the compiled script will be different in the two modes. Hence, when testing against compiled output - be careful of the difference.

TotallyNotChase:
And having to use CPP in test files is a pain..... I wish backpack worked.

I'm having a hard time getting fourmolu to not die on a file with CPP

TotallyNotChase:
Alright, that should work.

Sorry for butchering up your test file @blamario :(

L-as (approved):
LGTM

PR #204, opened 2022-01-27
Remove pcon' and pmatch' exports from prelude

TotallyNotChase:
These are unnecessary now thanks to #189

### DIFF
--- a/examples/Examples/LetRec.hs
+++ b/examples/Examples/LetRec.hs
@@ -5,6 +5,7 @@ module Examples.LetRec (tests) where
 
 import Plutarch (printTerm)
 import Plutarch.Builtin (pasConstr, pforgetData)
+import Plutarch.Internal.PlutusType (pcon', pmatch')
 import Plutarch.Prelude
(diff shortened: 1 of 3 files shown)


examples/Examples/LetRec.hs
```diff
@@ -5,6 +5,7 @@ module Examples.LetRec (tests) where
 
 import Plutarch (printTerm)
 import Plutarch.Builtin (pasConstr, pforgetData)
+import Plutarch.Internal.PlutusType (pcon', pmatch')
```
L-as, line 8: Why not import it from `Plutarch`?

L-as (changes requested):

examples/Examples/LetRec.hs
TotallyNotChase: Primarily since I meant for the import to be "more granular". But it's probably better to import `Plutarch` than an internal module - especially since it's a public facing example.

L-as (approved):
LGTM

PR #206, opened 2022-01-27
Fix duplicate work in pmatch' impl of builtin lists

TotallyNotChase:
This also tries to set up a test for this kind of sanity tests.

### DIFF
--- a/examples/Examples/PlutusType.hs
+++ b/examples/Examples/PlutusType.hs
@@ -56,30 +56,16 @@ tests =
         "PlutusType instances sanity"
         -- TODO: Add more sanity tests here!
         [ testCase "PBuiltinList" $ do
-            pmatchTargetEvalBList $ pconstant [1 :: Integer, 2, 3, 4]
+            pmatchTargetEval $ pconstant [1 :: Integer, 2, 3, 4]
         ]
     ]
 
 -- CPP support isn't great in fourmolu.
 {- ORMOLU_DISABLE -}
 
--- | Make sure the target builtin list of 'pmatch' is not evaluated multiple times.
-pmatchTargetEvalBList :: (HasTester, PLift a) => ClosedTerm (PBuiltinList a) -> Assertion
-pmatchTargetEvalBList target = pmatch (ptrace (pconstant tag) target) (\case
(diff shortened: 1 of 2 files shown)


examples/Examples/PlutusType.hs
```diff
@@ -39,13 +40,50 @@ swap x = pmatch x $ \case
...
+  `traces` replicate 1 tag
+#else
+  `traces` []
+#endif
+  where
+    tag = "evaluating"
+
+-- FIXME: This needs to fully "evaluate" (e.g evaluate in UPLC) the `x` within the `pmatch` handler.
+-- | Make sure the target of 'pmatch' is only evaluated once.
+_pmatchTargetEval :: HasTester => PlutusType p => ClosedTerm p -> Assertion
```
TotallyNotChase, line 81: This is what I was trying. It's missing one key piece- the argument to the `pmatch` handler needs to have its inner terms "evaluated" with `plet` (see: `pmatchTargetEvalBList` impl). Only then this test is valid and can work generically on any plutus type.

The thing is, I don't think we can do that without another typeclass.

TotallyNotChase: Actually never mind, I think I fixed it.

L-as (approved):
LGTM

examples/Examples/PlutusType.hs
```diff
@@ -39,13 +40,36 @@ swap x = pmatch x $ \case
...
+-- CPP support isn't great in fourmolu.
+{- ORMOLU_DISABLE -}
+
+-- | Make sure the target of 'pmatch' is only evaluated once.
+pmatchTargetEval :: HasTester => PlutusType p => ClosedTerm p -> Assertion
+pmatchTargetEval target = pmatch (ptrace (pconstant tag) target) (\x -> plet (pcon x) $ \_ -> pconstant ())
+#ifdef Development
+  `traces` replicate 1 tag
+#else
+  `traces` []
```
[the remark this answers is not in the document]
TotallyNotChase: Good idea. I'll make a note to make that change the next time I need to touch the test suite - unless it's already taken care of.

PR #209, opened 2022-01-27
Added an example using the proper monadic do with Cont

blamario:
An example of proper monadic do, where the monad is Cont.


L-as (approved):
LGTM

PR #212, opened 2022-01-28
Make pmatchDataSum use punDataSum when possible

L-as:


examples/Examples/Api.hs
```diff
@@ -190,7 +199,14 @@ tests =
...
+        printTerm getFields @?= getFields_compiled
+    , testCase "getFields'" $
+        printTerm getFields' @?= getFields_compiled
     ]
 
 ctx_compiled :: String
 ctx_compiled = "(program 1.0.0 #d8799fd8799f9fd8799fd8799fd8799f41a0ff00ffd8799fd8799fd87a9f41a1ffd87a80ffa0d8799f41d0ffffffff80a0a141c0a149736f6d65746f6b656e018080d8799fd8799fd87980d87a80ffd8799fd87b80d87a80ffff9f45ab01fe235c4312301443abcdefff80d8799f41b0ffffd87a9fd8799fd8799f41a0ff00ffffff)"
+
+getFields_compiled :: String
+getFields_compiled = "(program 1.0.0 (\\i0 -> force (force sndPair) (unConstrData i1)))"
```
TotallyNotChase, line 212: this output will be different when tracing is on right? you should probably put the switch in here.

TotallyNotChase (approved):

examples/Examples/Api.hs
L-as: Why would it?

TotallyNotChase: Oh right that's not a partial match since it's a newtype. All good!

Issue #214, opened 2022-01-28
Improve `DataRepr` based machinery (e.g `pmatch`) efficiency


TotallyNotChase:
We should try to have as little deviation as possible, between the performance of typed abstractions (e.g `pmatch`, `pletFields`, `pfield`) and raw `pasConstr`. Currently `pmatch` in particular can generate inefficient code in certain cases.

We should also have benchmarks to track the progress of this.

L-as:
This has already been fixed AFAICT.

TotallyNotChase:
> This has already been fixed AFAICT.

Not quite, we still have some pretty significant inefficiencies in place. These are being tracked by the following benches-
* `types:data:deconstruction:matching:*:sumtype(ignore-fields)`
* `types:data:deconstruction:matching:*:sumtype(exhaustive)(ignore-fields)`
* `types:data:deconstruction:combined:*:toValidatorHash`

I meant this issue to track our progress on *optimizing* the stuff.

PR #215, opened 2022-01-28
Add tests for data deconstruction

TotallyNotChase:
This should help us track our progress on matching the performance of `PDataRepr` based machinery to raw untyped builtin functions.

Relevant: https://github.com/Plutonomicon/plutarch/issues/214

### DIFF
--- a/plutarch-benchmark/bench/Main.hs
+++ b/plutarch-benchmark/bench/Main.hs
@@ -1,11 +1,19 @@
+{-# LANGUAGE QualifiedDo #-}
+
 module Main (main) where
 
 import Data.ByteString (ByteString)
+import Plutarch.Api.V1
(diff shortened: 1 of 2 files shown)


plutarch-benchmark/bench/Main.hs
```diff
@@ -130,3 +139,226 @@ intListBench =
...
+                      ( \x ->
+                          plet (pasConstr # x) $ \d ->
+                            pif
+                              (pfstBuiltin # d #== 0)
+                              (psndBuiltin # d)
+                              perror
+                      )
+                      #$ pconstant
+                      $ toData minting
+                , bench "sumtype(exhaustive)" $
```
TotallyNotChase, line 271: I *think* there's some inefficiency in this one and the "sumtype(exhaustive)(ignore-fields)" bench. Can't find it though.

L-as (approved):
Thanks!

Issue #202, opened 2022-01-27
Linear types to track use count?


L-as:
#201 doesn't allow inlining the term, since we don't know how many times it will be used. With linear types, we can precisely track how many times something will be used. Not sure how feasible this is. Do note that this wouldn't be "real" linearity, e.g. even though the HOF passed to `pmap` is only used "once", it should probably not be linear, since morally it is used many times.

L-as:
It would, that's part of why we can't do it right now.

L-as:
I've thought more and more about this, and this would be *extremely* useful, and would truly take the eDSL to the next level. Perhaps this should be Plutarch 2.0.

L-as:
Perhaps we also want dependent types for Plutarch 2.0, let's hope they land in GHC this year!

L-as:
I'm less convinced we need linear types.

L-as:
![image](https://user-images.githubusercontent.com/22075344/164013986-36d0c005-b351-4b26-b2d7-66ae156fde01.png)

L-as:
We don't need this.

Issue #201, opened 2022-01-27
`punsafeLetPure`


L-as:
Should be like `plet`, but should not bind anything if the bound variable isn't used. It's unsafe because the semantics of the program can change unexpectedly depending on whether you use the variable or not. You should only use this if you're 100% sure that the binding will never fail. This is e.g. true for indexing `PDataRecord`.

L-as:
If the bound variable is only used once, it should be inlined, but it's not possible to know how many times it will be used, because it could be one appearance in e.g. a `pmap`.

PR #217, opened 2022-01-28
`plet` the head and tail in `pmatch'` impl for builtin lists

TotallyNotChase:
This evaluates the head and tail before handing it off to the user provided matching function.

Previously
```hs
test :: Term s PInteger
test = pmatch (pconstant @(PBuiltinList PInteger) [1, 2, 3]) $ \case
   PCons x _ -> x + x
   PNil -> 0
```
would call `HeadList` twice.

This doesn't happen with this fix - but there's a cost..... the head and tail are evaluated whether or not you use them in your function.

I'm unhappy with this compromise - but we ought to make our API consistent. In most other places, the bindings introduced by such continuation functions (e.g `plet`, `pmatch`, `pletFields`) are already evaluated before being fed to the function. We either make all of them take in unevaluated stuff to allow for laziness - or we feed all of them evaluated terms.

This is somewhat related to #214. There's a benchmark for testing the efficiency of `pmatch` vs `pasConstr` when ignoring the fields altogether (just matching on constructor). `pmatch` *will always lose* in this case since the fields are evaluated before being fed to the user provided function.


L-as:
Linear types would solve this. #202

TotallyNotChase:
> Linear types would solve this. #202

That doesn't seem neither here nor there, I'm afraid. It'd be a rather big undertaking - though that's not the issue. The GHC 9 requirement is the issue.

L-as:
#201 would also fix this.

L-as:
After thinking about this more, this is impossible to solve completely without linear types. See second comment in #201. This PR + `punsafeLetPure` would still be better than nothing.

TotallyNotChase:
@L-as I think we should get this fix merged for the time being. Although I do have it in mind to tackle `punsafeLetPure` - it's not gonna be done anytime soon. Until then, this behavior is a big trap we ought to get rid of.

L-as:
What other `pmatch` instances do we have where this is done where it's not a detriment in some cases?

TotallyNotChase:
`pmatch` for `PList` feeds in pre-evaluated `x` and `xs` to the case handler. `pmatchRepr` also does the extra `SndPair` to obtain the fields so that it's readily available in the handler function. As far as I can tell, every `pmatch` has this same behavior - except the `PBuiltinLIst` one. All of them share the same performance detriment when the binding isn't actually used. We've discussed that already. I don't think it's trivially possible to get rid of this performance detriment for scott encoded types though - strict application is rather problematic, but that's a topic for another day.

We should either make it so *all* `pmatch` handlers have unevaluated bindings in the Plutarch type constructors, or so *all* the `pmatch` handlers have evaluated bindings in the constructor fields.

L-as:
For `PList`, that's the most optimal way to do it. Not precomputing them makes no sense. Hence, it's not a detriment in some cases, it's always the most optimal way.

For `pmatchDataSum`, we don't actually do this in the single handler case AFAICT! In the multiple handler case, we do do this, and that's because we have multiple branches. There's not another way to go about it.  This is the most optimal way.

As for `PDataRecord`, we don't `plet` the heads either, because those aren't always shared.

The general guidance is to **never** `plet` unless you are absolutely sure it's optimal or as good as not doing it.
I don't think we have a single case for `pmatch` specifically where this isn't followed.

The reason being that if you do force a `plet`, then the user can't `unplet` it, but if you don't do a `plet`, the user can choose.
In addition, there's an optimisation such that `plet`ing the same thing twice is reduced into one `plet` as you'd expect.

L-as:
I think this is a documentation problem. We should just note this in the guide until we have linear types.

TotallyNotChase:
> I think this is a documentation problem. We should just note this in the guide until we have linear types.

Yeah, was just about to say. If we do indeed have valid reasons for different behavior - we ought to acknowledge it in the guide. I'll get on that.

L-as:
Thanks!

L-as:
Closing this since this isn't really a fix.

PR #224, opened 2022-01-31
checkSignatorCont: match on PSpending

srid:
Make `checkSignatoryCont` behave similar to `checkSignatory`. 

It looks like with `P.do`, match failures invoke `perror` implicitly (not clear with what string exactly) via MonadFail. But the cont monad does not implement MonadFail, so we have to case over and `perror` manually. The PR make this obvious.


L-as:
We're not even using do-syntax anymore FYI.

PR #225, opened 2022-01-31
TermCont: `MonadFail` instance; example; benchmark

srid:
- Add `MonadFail` instance to TermCont
    - Create separate `TermCont.hs` module, to avoid cyclic imports when using `ptraceError`
- Add an example in `Api.hs` (builds on top of #224 )
- Add new benchmark group to evaluate the four different syntaxes


srid:
cc @Geometer1729

srid:
> Why are you removing the exports from Plutarch.hs, that's a breaking change.

I had to remove them to avoid cyclic imports. In this PR, `Plutarch.hs` cannot import `Plutarch.Internal.TermCont`:

```
Module imports form a cycle:
        module ‘Plutarch.Unsafe’ (./Plutarch/Unsafe.hs)
        imports module ‘Plutarch’ (Plutarch.hs)
  which imports module ‘Plutarch.Internal.TermCont’ 
(./Plutarch/Internal/TermCont.hs)
  which imports module ‘Plutarch.Trace’ (./Plutarch/Trace.hs)
  which imports module ‘Plutarch.Bool’ (./Plutarch/Bool.hs)
  which imports module ‘Plutarch.Unsafe’ (./Plutarch/Unsafe.hs)
```

So I exported TermCont from `Plutarch.Prelude` instead, thinking that this ought to be sufficient.

---

What is the different between `import Plutarch` and `import Plutarch.Prelude`, btw, from user standpoint?

srid:
Then we make TermCont non-internal (ie. `Plutarch.TermCont`) because it is importing **non**-internal modules (Trace, then Bool, etc.), wouldn't you agree?

L-as:
You are right, albeit it's still a backward incompatible change, not to mention that hashOpenTerm uses it.

srid:
So should I make this against `staging`?

L-as:
Can't you reexport it from Plutarch to avoid that? You can do some reorganising to remove the cyclicity.

PR #228, opened 2022-02-01
Explicitly set the locale encoding to utf8

blamario:
Fixed the issue with

    examples: <stdout>: commitBuffer: invalid argument (invalid character)

with `setLocaleEncoding utf8` as suggested by Maciej Bendkowski.


srid:
FTR, I use `with-utf8` for this: https://serokell.io/blog/haskell-with-utf8

PR #230, opened 2022-02-01
`TermCont` in guide

TotallyNotChase:
* Add a section for `TermCont` do syntax
* Reference it in the "Alternatives to `do` syntax" section.
* Re-organize some of the points in the "Alternatives to `do` syntax" section. No changes were made to the points themselves.

### DIFF
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -429,8 +429,7 @@ import Plutarch.Prelude
 f :: Term s (PScriptPurpose :--> PUnit)
 f = plam $ \x -> unTermCont $ do
   PSpending _ <- tcont $ pmatch x
-  ptrace "matched spending script purpose"
-  pconstant ()
+  pure $ ptrace "matched spending script purpose" $ pconstant ()
 ```
 
 The best part is that this doesn't require `QualifiedDo`! So you don't need GHC 9.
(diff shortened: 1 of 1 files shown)


docs/GUIDE.md
```diff
@@ -400,17 +397,60 @@ There are three ways to do this-
   ```
 
   Note that you have to translate the pattern matching manually, as `Cont` doesn't have a `MonadFail` instance.
-* Don't use do syntax at all. You can easily translate the `do` syntax to regular continuation chains.
+* Use [`RebindableSyntax`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/rebindable_syntax.html). You can replace the `>>=`, `>>`, and `fail` functions in your scope with the ones from `Plutarch.Monadic` using `RebindableSyntax`. This is arguably a bad practice but the choice is there. This will let you use the `do` syntax word for word. Although you wouldn't be qualifying your `do` keyword (like `P.do`), you'd just be using `do`.
```
srid, line 400: Why is this considered bad practice?

I do know that it means you can't use regular do-syntax, but that's rarely used in smart contracts (except for conveniences like Maybe monad).

srid (approved):

docs/GUIDE.md
TotallyNotChase: Yeah that's pretty much the reason. It might not be as prominent in smart contracts, but it's still "arguably a bad practice" in general.

PR #231, opened 2022-02-02
Fix constructor ordering for `PMaybeData`

TotallyNotChase:
We should really have tests for ensuring the right constructor id has been assigned. Don't really know a good way to test it though.


L-as (approved):
LGTM

srid:
> Don't really know a good way to test it though.

Here's an idea: https://github.com/Plutonomicon/plutarch/issues/233

PR #234, opened 2022-02-02
Deconstr bench enhance

TotallyNotChase:
Well, that diff rendering is terrible - so here's all the changes-
* Move the `let` bindings being used in each `benchGroup` out to a `where` block within `deconstrBench`. Less nesting, better looking. Doesn't change anything anyway.
* Make the `sumtype(exhaustive)*` bench apply *all* `ScriptPurpose` constructors one by one to get a significantly better idea of performance.

  The reason the raw untyped one was performing *worse* than the typed one previously was because a `Spending` script purpose just happened to be the happy path for the typed one - but not the untyped one.


L-as (approved):
LGTM

PR #240, opened 2022-02-03
Bool instance

anton-k:
There is cool package `Boolean` which provides generic Booleans useful for EDSLs.
I thought maybe it can be useful for Plutarch.

### DIFF
--- a/Plutarch/Bool.hs
+++ b/Plutarch/Bool.hs
@@ -15,6 +15,18 @@ module Plutarch.Bool (
   pand,
   pand',
   por',
+  -- * Reexport for Booleans
+  Boolean(..),
+  BooleanOf,
+  IfB(..),
+  EqB(..),
+  OrdB(..),
+  boolean,
+  minB,
+  maxB,
+  sort2B,
+  guardedB,
+  caseB,
 ) where
 
 import Plutarch.Internal.Other (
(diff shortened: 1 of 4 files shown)


MatthewCroughan (approved):

TotallyNotChase:
I don't think this needs to be in the main Plutarch repo though. Perhaps it makes more sense to have it be a separate repository that people can add as a dependency if they need to.

anton-k:
@TotallyNotChase  With those instances we get some nice functions that emulate guards and case-expressions. I think it can also be useful for plutarch.

TotallyNotChase:
Ah, `guardB` and `caseB` look pretty cool. But I think those should just be regular, monomorphized Plutarch term utilities. Won't need the `boolean` dep or instances then.

L-as:
`guardB` looks nice, but I don't think it's good to depend on an unmaintained dependency. Last update is almost 5 years ago.

anton-k:
That's ok. I just saw that you use Num instances  and thought that you might be interested to rely on classes for booleans too.
Just to say couple of words in defense of that package. If something does not change every half a year does not mean that it's not useful. Num has not changed for about a 30 years. How often would you like the meaning of boolean operators or comparison operators to be changed?

nixinator:
https://xkcd.com/2347/

L-as:
I completely agree with that sentiment, however, GHC and base often have breaking changes.

L-as:
If we are to have this in Plutarch we want to do this without adding the unnecessary dependency.

PR #242, opened 2022-02-04
`PlutusType` generic derivation for data encoded types + more Data stuff

### LINKED ISSUE #187: `instance PIsDataRepr a => PlutusType (PIsDataReprInstances a)`
The missing piece is `pcon`. We need to use generics to implement it.

TotallyNotChase:
There is much to be done. I'll put a changelog with the related stuff in here soon.

On the bright side, the `PlutusType` derivation is in place.

Resolves #187 
Progresses #135

### DIFF
--- a/Plutarch/DataRepr/Internal.hs
+++ b/Plutarch/DataRepr/Internal.hs
@@ -23,7 +23,6 @@ module Plutarch.DataRepr.Internal (
 
 import Data.Kind (Type)
 import Data.List (groupBy, maximumBy, sortOn)
-import Data.Proxy (Proxy)
 import GHC.TypeLits (
   ErrorMessage (ShowType, Text, (:<>:)),
   KnownNat,
@@ -33,18 +32,21 @@ import GHC.TypeLits (
   natVal,
   type (+),
  )
-import Generics.SOP (Code, Generic, I (I), NP (Nil, (:*)), SOP (SOP), to)
+import Generics.SOP
 import Plutarch (
   Dig,
+  PInner,
   PMatch,
   PType,
+  PlutusType,
   S,
   Term,
+  pcon',
   perror,
   phoistAcyclic,
   plam,
   plet,
-  pmatch,
+  pmatch',
   (#),
   (#$),
   type (:-->),
@@ -53,6 +55,7 @@ import Plutarch.Bool (pif, (#==))
 import Plutarch.Builtin (
   PAsData,
   PBuiltinList,
+  PBuiltinPair,
   PData,
   PIsData,
   pasConstr,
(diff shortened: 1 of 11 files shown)


Plutarch/DataRepr/Internal.hs
```diff
@@ -33,18 +32,21 @@ import GHC.TypeLits (
   natVal,
   type (+),
  )
-import Generics.SOP (Code, Generic, I (I), NP (Nil, (:*)), SOP (SOP), to)
+import Generics.SOP
```
TotallyNotChase, line 35: Note to self: I forgot to put an explicit export list here.

TotallyNotChase:
> Doesn't the `pconstantData` stuff etc. fit better in another PR?

Potentially yes, but the feature itself is tiny and I added a bunch of tests for it. This PR does a bunch of stuff related to data and datarepr anyway so I figured I may as well.

Plutarch/DataRepr.hs
```diff
@@ -22,5 +24,27 @@ module Plutarch.DataRepr (
...
+You can specify the label to associate with the field using type applications-
+
+@
+
+foo :: Term s (PDataRecord '[ "fooField" ':= PByteString ])
+foo = pdcons @"fooField" # pdata (phexByteStr "ab") # pdnil
+
+@
+-}
+pdcons :: forall label a l s. Term s (PAsData a :--> I.PDataRecord l :--> I.PDataRecord ((label 'I.:= a) ': l))
```
TotallyNotChase, line 45: I wasn't sure about the type variable ordering here. Should `label` come before or after `a` - which one would users be more likely to use type application for?

TotallyNotChase:
That's weird, why did the CI suddenly fail with a compilation error? I only change between [`cb3668b`](https://github.com/Plutonomicon/plutarch/pull/242/commits/cb3668b8635b9faec843d52a809674e9b35aa8ac) and [`ef476b4`](https://github.com/Plutonomicon/plutarch/pull/242/commits/ef476b4394ceaa806e30f343f2be2505cf4b1f31) were documentation changes....?

My local test succeeds as well.

Plutarch/Builtin.hs
```diff
@@ -275,3 +298,13 @@ pinnerData = punsafeCoerce
...
 pouterData = punsafeCoerce
+
+pconstrBuiltin :: Term s (PInteger :--> PBuiltinList PData :--> PAsData (PBuiltinPair PInteger (PBuiltinList PData)))
+pconstrBuiltin = punsafeBuiltin $ PLC.ConstrData
+
+{- | Create a Plutarch-level 'PAsData' constant, from a Haskell value.
+Example:
+> pconstantData @PInteger 42
+-}
+pconstantData :: forall p h s. (ToData h, PLifted p ~ h, PConstanted h ~ p) => h -> Term s (PAsData p)
```
TotallyNotChase, line 309: Apparently this can't build bytestring constants (but it can build newtypes to bytestrings such as `PCurrencySymbol`). This is because `ByteString` doesn't have a `ToData` instance - only `BuiltinByteString` does.....

L-as:
This PR is huge. Please split it up.

PR #248, opened 2022-02-07
Mario/ffi without tests

blamario:


Plutarch/FFI.hs
```diff
@@ -0,0 +1,106 @@
...
+  PlutusTxInner ForallPhantom _ = ForallPhantom
+  PlutusTxInner (a -> b) x = PlutusTxInner a b -> PlutusTxInner b x   -- hack to support Scott encodings
+  PlutusTxInner a x = PlutusTxInner (ScottFn (ScottList (SOP.Code a) x) x) x
+
+{- |
+  List of scott-encoded constructors of a Haskell type (represented by 'SOP.Code')
+
+  ScottList (Code (Either a b)) c = '[a -> c, b -> c]
+-}
+type ScottList :: [[Type]] -> Type -> [Type]
```
L-as, line 76: Don't we already have this elsewhere?

srid: https://github.com/Plutonomicon/plutarch/blob/master/Plutarch/Internal/PlutusType.hs#L298

That one's for Plutarch types, however. This one seems to be for Haskell types.

blamario: Yup. The existing one is not reusable here.

Plutarch/FFI.hs
```diff
@@ -0,0 +1,106 @@
...
+  PlutarchInner (PDelayed a) x = PlutarchInner a x
+  PlutarchInner p x = PlutarchInner (PInner p x) x
+
+type family PlutusTxInner (t :: Type) (any :: Type) :: Type where
+  PlutusTxInner Bool _ = Bool
+  PlutusTxInner Integer _ = Integer
+  PlutusTxInner BuiltinString _ = Text
+  PlutusTxInner ForallPhantom _ = ForallPhantom
+  PlutusTxInner (a -> b) x = PlutusTxInner a b -> PlutusTxInner b x   -- hack to support Scott encodings
+  PlutusTxInner a x = PlutusTxInner (ScottFn (ScottList (SOP.Code a) x) x) x
```
srid, line 69: Doesn't `ScottList (ToPType2 (SOP.Code a))` allow us to reuse the existing type families from `PlutusType.hs`?

L-as:
I discussed this with Mario off GH, and I personally think that we should give up on translating arbitrary types between PlutusTx and Plutarch code. I think the most realistic thing to do is to cover the built-in types only. That would unfortunately exclude booleans, but booleans are Scott-encoded by PlutusTx. We shouldn't rely on the encoding either. The encoding might very well change.

PR #250, opened 2022-02-07
Add `PlutusType` derivation support for data encoded types

### LINKED ISSUE #187: `instance PIsDataRepr a => PlutusType (PIsDataReprInstances a)`

TotallyNotChase:
Subset of #242 

Resolves #187


Plutarch/DataRepr.hs
```diff
@@ -22,5 +24,27 @@ module Plutarch.DataRepr (
...
+You can specify the label to associate with the field using type applications-
+
+@
+
+foo :: Term s (PDataRecord '[ "fooField" ':= PByteString ])
+foo = pdcons @"fooField" # pdata (phexByteStr "ab") # pdnil
+
+@
+-}
+pdcons :: forall label a l s. Term s (PAsData a :--> I.PDataRecord l :--> I.PDataRecord ((label 'I.:= a) ': l))
```
TotallyNotChase, line 45: I wasn't sure about the type variable ordering here. Should label come before or after a - which one would users be more likely to use type application for?

L-as:
`pmatch` should be optimally efficient AFAICT, because you already know whether it's a cons or not.

TotallyNotChase:
> `pmatch` should be optimally efficient AFAICT, because you already know whether it's a cons or not.

That's fine - we don't have an issue with efficiency here but I'm more concerned about what this would achieve other than add in more concepts to `PDataRecord` that users *shouldn't* need. It gives the impression that `PDataRecord` is just another general purpose plutarch type that can be used everywhere. But I thought it was more domain specific, particularly for encoding product types and accesssing them. We *have* `PBuiltinList` and its plutus type instance already after all.

There'd be 2 ways to match on data records. There'd probably also be 2 ways to construct them because `pcon $ PDCons x xs`, or `pcon $ PDCons @"label" x xs` isn't as convenient as `pdcons # x # xs` and `pdcons @"label" # x # xs`.

FWIW, I did implement `PlutusType` locally just to see what it looks like. I just don't see a point in doing it, and I'm concerned about the implications such an instance gives.

L-as:
`PDataRecord` is a general purpose type though.

TotallyNotChase:
Alright, I added a `PlutusType` instance for `PDataRecord`

CHANGELOG.md
```diff
@@ -8,6 +8,28 @@
...
+
+  All existing ledger api types now have `PlutusType` instances - not just `PMatch`.
+
+  Started by: [#250](https://github.com/Plutonomicon/plutarch/pull/250)
+
+- Add `PDataRecord` construction utilities, necessary for full usage of data encoded `PlutusType` instances.
+
+  In particular, you can build `PDataRecord`s with `pdcon` and `pdnil` - refer to the guide for more info.
+
+  `pdcon` and `pdnil` are also exported from `Plutarch.Prelude`.
```
L-as, line 21: `pdcon`?

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
     "haskell.formattingProvider": "fourmolu"
-}
+}
```
L-as, line 3: Why did you remove the newline here?

Plutarch/DataRepr/Internal.hs
```diff
@@ -65,17 +91,63 @@ import Plutarch.Builtin (
...
  `PBuiltinList PData`.
 -}
-data PDataRecord (as :: [PLabeledType]) (s :: S)
+data PDataRecord (as :: [PLabeledType]) (s :: S) where
+  PDCons ::
+    forall name x xs s.
+    Term s (PAsData x) ->
+    (Term s (PDataRecord xs)) ->
+    PDataRecord ((name ':= x) ': xs) s
+  PDNil :: PDataRecord '[] s
```
L-as, line 112: Looks great

Plutarch/DataRepr/Internal.hs
```diff
@@ -65,17 +91,63 @@ import Plutarch.Builtin (
...
+
+instance PlutusType (PDataRecord ((name ':= x) ': xs)) where
+  type PInner (PDataRecord ((name ':= x) ': xs)) _ = PBuiltinList PData
+  pcon' (PDCons x xs) = punsafeCoerce result
```
L-as, line 116: There's no need to use `punsafeCoerce` here. Just do `pto`.

Plutarch/DataRepr/Internal.hs
```diff
@@ -65,17 +91,63 @@ import Plutarch.Builtin (
...
+      result :: Term _ (PDataRecord ((name ':= x) ': xs))
+      result = pdcons # x # xs
+  pmatch' x f = pmatch x $ \case
+    PCons x' xs' ->
+      let x :: Term _ (PAsData x)
+          x = punsafeCoerce x'
+          xs :: Term _ (PDataRecord xs)
+          xs = punsafeCoerce xs'
+       in f $ PDCons x xs
+    PNil -> error "PDataRecord:pmatch':PNil:absurd"
```
L-as, line 127: This will cause any code that uses `pmatch` on `PDataRecord` to err at compile-time unconditionally.

Plutarch/DataRepr/Internal.hs
```diff
@@ -65,17 +91,63 @@ import Plutarch.Builtin (
...
+  PDNil :: PDataRecord '[] s
+
+instance PlutusType (PDataRecord ((name ':= x) ': xs)) where
+  type PInner (PDataRecord ((name ':= x) ': xs)) _ = PBuiltinList PData
+  pcon' (PDCons x xs) = punsafeCoerce result
+    where
+      result :: Term _ (PDataRecord ((name ':= x) ': xs))
+      result = pdcons # x # xs
+  pmatch' x f = pmatch x $ \case
+    PCons x' xs' ->
```
L-as, line 121: We already know the list is non-empty. The reason for this is that on line 114, we declare the instance for only non-empty data records. You can just use `phead`.

Plutarch/DataRepr/Internal.hs
```diff
@@ -65,17 +91,63 @@ import Plutarch.Builtin (
...
+      let x :: Term _ (PAsData x)
+          x = punsafeCoerce x'
+          xs :: Term _ (PDataRecord xs)
+          xs = punsafeCoerce xs'
+       in f $ PDCons x xs
+    PNil -> error "PDataRecord:pmatch':PNil:absurd"
+
+instance PlutusType (PDataRecord '[]) where
```
L-as, line 129: Great job BTW splitting up the instance into two.

Plutarch/DataRepr/Internal.hs
```diff
@@ -203,10 +275,29 @@ pmatchDataSum d handlers =
 -}
 newtype PIsDataReprInstances (a :: PType) (s :: S) = PIsDataReprInstances (a s)
 
+-- Q: Why are these superclass constraints here?
 class (PMatch a, PIsData a) => PIsDataRepr (a :: PType) where
```
L-as, line 279: `PMatch a` should be replaced with `PlutusType a`. This is to make sure people don't forget to derive all necessary instances. `PIsData a` helps avoid accidentally unsound code.

examples/Examples/ConstrData.hs
```diff
@@ -0,0 +1,145 @@
...
+    (PlutusType, PIsData, PDataFields)
+    via (PIsDataReprInstances (Triplet a))
+
+data PVehicle (s :: S)
+  = PFourWheeler (Term s (PDataRecord '["_0" ':= PInteger, "_1" ':= PInteger, "_2" ':= PInteger, "_3" ':= PInteger]))
+  | PTwoWheeler (Term s (PDataRecord '["_0" ':= PInteger, "_1" ':= PInteger]))
+  | PImmovableBox (Term s (PDataRecord '[]))
+  deriving stock (GHC.Generic)
+  deriving anyclass (Generic)
+  deriving anyclass (PIsDataRepr)
```
L-as, line 53: Why have separate lines for the anyclass derives?

L-as (changes requested):
Looks mostly great!

Plutarch/DataRepr/Internal.hs
TotallyNotChase: oh I meant to write `perror` - whoops.

Plutarch/DataRepr/Internal.hs
TotallyNotChase: Yes, but we need the head and tail. `ChooseList` (what `pmatch` will expand to here) is going to more efficient than head + tail. At the cost of also having to handle the absurd nil case.

examples/Examples/ConstrData.hs
TotallyNotChase: Yeah we've kinda been following this style throughout the entire repo.

Plutarch/DataRepr/Internal.hs
TotallyNotChase: Sounds good. Updated!

Plutarch/DataRepr/Internal.hs
TotallyNotChase: Aha! Good idea! Updated.

.vscode/settings.json
TotallyNotChase: I have no idea - it was probably done by vscode and I didn't notice.

This is a bit of an issue since I have my own local settings in this file and I have to remove them everytime I check out branches.... oh well.

CHANGELOG.md
TotallyNotChase: Good catch! Fixed.

.vscode/settings.json
L-as: If it's done by VSCode, then don't care about it. Just VSCode being stupid I suppose.

L-as: What are your own local settings? Would it be best if we remove the entire file entirely again? You might want to discuss this with @srid , I don't use VSCode.

Plutarch/DataRepr/Internal.hs
L-as: I'm not entirely sure what you mean. `pmatch` also just gets the head and tail separately, it only uses `ChooseList` to figure out whether it's a head or tail IIRC. Look at your own PR! https://github.com/Plutonomicon/plutarch/pull/217/files

examples/Examples/ConstrData.hs
L-as: It's been on accident, it's not a good style.

TotallyNotChase: We probably want to fix that as a cleanup task - do you want it done here in this example for the time being?

Plutarch/DataRepr/Internal.hs
TotallyNotChase: Oh lol - I was under the impression choose list takes in lambdas. Never mind! Updated.

examples/Examples/ConstrData.hs
L-as: Yeah let's make sure we don't add more bad code.

Plutarch/DataRepr/Internal.hs
```diff
@@ -276,8 +275,7 @@ pmatchDataSum d handlers =
 -}
 newtype PIsDataReprInstances (a :: PType) (s :: S) = PIsDataReprInstances (a s)
 
--- Q: Why are these superclass constraints here?
-class (PMatch a, PIsData a) => PIsDataRepr (a :: PType) where
+class (PlutusType a, PIsData a) => PIsDataRepr (a :: PType) where
```
L-as, line 275: I just realised this is a breaking change... sorry. Can we revert this commit? lol

TotallyNotChase: Hmm, breaking because users may not be implementing `PlutusType` yet? Don't we *want* them to do just that though, from now on?

But yes, we can revert it if you want.

L-as: Existing code would fail to compile because `PlutusType` isn't derived.

TotallyNotChase: Reverted - made a note to change this in the future.

examples/Examples/ConstrData.hs
TotallyNotChase: done!

L-as (approved):
LGTM

PR #255, opened 2022-02-08
Add remaining `PIsData` instances + `PTuple` utilities

TotallyNotChase:
The remaining changes from #242


Plutarch/Api/V1/Value.hs
```diff
@@ -5,44 +5,63 @@ module Plutarch.Api.V1.Value (
...
-instance PUnsafeLiftDecl PValue where type PLifted PValue = Plutus.Value
+instance PUnsafeLiftDecl PValue where type PLifted PValue = PlutusValue.Value
 deriving via
-  (DerivePConstantViaNewtype Plutus.Value PValue (PMap PCurrencySymbol (PMap PTokenName PInteger)))
+  (DerivePConstantViaNewtype PlutusValue.Value PValue (PMap PCurrencySymbol (PMap PTokenName PInteger)))
   instance
-    (PConstant Plutus.Value)
+    (PConstant PlutusValue.Value)
+
+newtype PAssetClass (s :: S) = PAssetClass (Term s (PBuiltinPair (PAsData PCurrencySymbol) (PAsData PTokenName)))
```
TotallyNotChase, line 56: I'm still debating if this should be a newtype to a `PBuiltinPair` or a `PTuple`

L-as:
What's the motivation for all this? They aren't actually used anywhere AFAICT?

TotallyNotChase:
> What's the motivation for all this? They aren't actually used anywhere AFAICT?

API completion. I assume some of those types would be useful in some usecases. Minswap, for example, needs `PAssetClass` and I've just had to define it in the repo itself. We may as well have all (or at least, almost all) the V1 ledger api types.

L-as:
I honestly don't think they belong here. They're not part of the `V1` interface. These are all just utilities.

TotallyNotChase:
> I honestly don't think they belong here. They're not part of the `V1` interface. These are all just utilities.

Oh, really? Didn't know that. It's confusing how they jam all this into `V1.Ledger.*` modules.

I'll strip the utility types off of the PR then.

examples/Examples/PIsData.hs
```diff
@@ -93,3 +98,45 @@ pInteger =
...
+    [ testCase "pfromData (pdata ()) ≡ ()" $
+        testPToFrom @PUnit ()
+    , testCase "pfromData (PlutusTx.toData ()) ≡ ()" $
+        testPFromDataCompat @PUnit ()
+    , testCase "PlutusTx.fromData (pdata ()) ≡ Just ()" $
+        testPDataCompat @PUnit ()
+    ]
+
+pPair :: HasTester => TestTree
+pPair =
```
L-as, line 115: Naming

Plutarch/Builtin.hs
```diff
@@ -258,10 +260,28 @@ instance PIsData PBool where
...
+  pdata x = punsafeCoerce target
+    where
+      target :: Term _ (PAsData (PBuiltinPair PInteger (PBuiltinList PData)))
+      target = f # punsafeCoerce x
+      f = phoistAcyclic $
+        plam $ \pair -> pconstrBuiltin # 0 #$ pcons # (pfstBuiltin # pair) #$ pcons # (psndBuiltin # pair) # pnil
+
+instance PIsData PUnit where
+  pfromData _ = pconstant ()
+  pdata _ = punsafeCoerce $ pconstant (Constr 0 [])
```
L-as, line 279: Why not an empty list as the internal representation?

Plutarch/Api/V1/Tuple.hs
```diff
@@ -0,0 +1,25 @@
+module Plutarch.Api.V1.Tuple (PTuple, ptuple, ptupleFromBuiltin, pbuiltinPairFromTuple) where
+
+import Plutarch.Builtin (pconstrBuiltin, pforgetData)
+import Plutarch.Prelude
+import Plutarch.Unsafe (punsafeCoerce)
+
+type PTuple a b =
+  PDataSum
```
L-as, line 8: Why not `PDataRecord`? Seems unnecessary.

L-as (changes requested):
questions

examples/Examples/PIsData.hs
TotallyNotChase: Chosen to be consistent with rest of this module. Perhaps they should all be renamed to something like `ppairTests` and similar?

Plutarch/Builtin.hs
TotallyNotChase: Have to follow the rules set before us. `toData ()` from Plutus creates a `Constr 0 []`.

Plutarch/Api/V1/Tuple.hs
TotallyNotChase: I'm also curious why `PTuple` was originally defined as a `PDataSum`. It could just be a `newtype` to a `PDataRecord`. It can't be `PDataRecord` alone though. That's just a builtin list, `PTuple` is a `Constr` data.

Plutarch/Builtin.hs
L-as: `pfromData . pconstantData`, but this is valid in either case in this case! In any case I agree yeah.

Plutarch/Api/V1/Tuple.hs
L-as: No, I mean, why represent it as a `Constr`?

TotallyNotChase: Well we have to - that's how the ledger api does it. Tilde discovered this originally.

examples/Examples/PIsData.hs
L-as: I do think so. This was a mistake that we didn't notice.

Plutarch/Api/V1/Tuple.hs
L-as: Ah I see, I didn't realise we were using it there.

examples/Examples/PIsData.hs
TotallyNotChase: Updated.

L-as (approved):
LGTM

PR #264, opened 2022-02-09
Mario/ffi without tests

blamario:


Plutarch/FFI.hs
```diff
@@ -0,0 +1,114 @@
...
+unsafeForeignImport :: CompiledCode t -> ClosedTerm p
+unsafeForeignImport c = Term $ const $ TermResult (RCompiled $ UPLC.toTerm $ unScript $ fromCompiledCode c) []
+
+type family a ~~ b :: Constraint where
+  ForallPhantom ~~ _ = ()
+  _ ~~ ForallPhantom = ()
+  a ~~ b = a ~ b
+
+type family PlutarchInner (p :: PType) (any :: PType) :: Type where
+  PlutarchInner PBool _ = Bool
```
L-as, line 63: `BuiltinBool` etc. instead of `Bool` https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx-Builtins-Internal.html#t:BuiltinBool

Plutarch/FFI.hs
```diff
@@ -0,0 +1,114 @@
...
+  PlutarchInner PInteger _ = Integer
+  PlutarchInner PString _ = Text
+  PlutarchInner PhorallPhantom _ = ForallPhantom
+  PlutarchInner (a :--> b) x = PlutarchInner a b -> PlutarchInner b x -- hack to support Scott encodings
+  PlutarchInner (PDelayed a) x = PlutarchInner a x
```
L-as, line 68: This isn't legal

Plutarch/FFI.hs
```diff
@@ -0,0 +1,114 @@
...
+import Plutus.V1.Ledger.Scripts (Script (unScript), fromCompiledCode)
+import PlutusTx.Code (CompiledCode, CompiledCodeIn (DeserializedCode))
+import PlutusTx.Prelude (BuiltinString)
+import UntypedPlutusCore (fakeNameDeBruijn)
+import qualified UntypedPlutusCore as UPLC
+
+data ForallPhantom :: Type
+data PhorallPhantom :: PType
+
+foreignExport :: PlutarchInner p PhorallPhantom ~~ PlutusTxInner t ForallPhantom => ClosedTerm p -> CompiledCode t
```
L-as, line 39: I'm honestly not sure why this type is this complex. Why not just do `ClosedTerm p -> ClosedTerm (PlutarchToPlutusTx p)`? You could make type family injective to make it reversible.

Plutarch/FFI.hs
```diff
@@ -0,0 +1,114 @@
...
+unsafeForeignExport t = DeserializedCode program Nothing mempty
+  where
+    program =
+      UPLC.Program () (UPLC.Version () 1 0 0) $
+        UPLC.termMapNames fakeNameDeBruijn $
+          compile' $
+            asClosedRawTerm t
+
+unsafeForeignImport :: CompiledCode t -> ClosedTerm p
+unsafeForeignImport c = Term $ const $ TermResult (RCompiled $ UPLC.toTerm $ unScript $ fromCompiledCode c) []
```
L-as, line 55: Rather than returnining `p`, you should return `POpaque`, and it wouldn't be unsafe either.

Plutarch/FFI.hs
```diff
@@ -0,0 +1,114 @@
...
+data ForallPhantom :: Type
+data PhorallPhantom :: PType
+
+foreignExport :: PlutarchInner p PhorallPhantom ~~ PlutusTxInner t ForallPhantom => ClosedTerm p -> CompiledCode t
+foreignExport = unsafeForeignExport
+
+foreignImport :: PlutarchInner p PhorallPhantom ~~ PlutusTxInner t ForallPhantom => CompiledCode t -> ClosedTerm p
+foreignImport = unsafeForeignImport
+
+unsafeForeignExport :: ClosedTerm p -> CompiledCode t
```
L-as, line 45: I think we can possibly use `Any` instead of `t` here to make it safe.

L-as: If you're using this from PlutusTx code you'd use `unsafeCoerce` to turn it into the real thing.

Plutarch/Internal.hs
```diff
@@ -75,6 +77,7 @@ data RawTerm
   | RDelay RawTerm
   | RConstant (Some (ValueOf PLC.DefaultUni))
   | RBuiltin PLC.DefaultFun
+  | RCompiled (UPLC.Term DeBruijn UPLC.DefaultUni UPLC.DefaultFun ())
```
L-as, line 80: This is completely fine.

L-as (changes requested):
.

Plutarch/FFI.hs
blamario: `PlutarchToPlutusTx PScriptContext` would need to reduce to `ScriptContext`, and so on for all ADTs including user-defined ones.

Plutarch/FFI.hs
blamario: A typical `PlutusTx` user has no clue about data type representation of various types, and even less clue how they match between PlutusTx and Plutarch. Even I was surprised to find the `Bool` representations were mismatched. What's more, a mismatch would not be caught at compile time and not necessarily even at the first run-time: it may be just lurking in a rarely taken execution path. After it's submitted on chain, in a validator script guarding millions of dollars. We can't leave a bomb like this in users' hands.

L-as: Isn't this an argument against having `unsafeForeignExport`? The same arguments applies here as with `unsafeForeignImport`.

L-as: `foreignExport :: ClosedTerm p -> CompiledCode Any` wouldn't be unsafe though.

Plutarch/FFI.hs
L-as: You can not export `ClosedTerm PScriptContext` as `CompiledCode ScriptContext`.

blamario: That's true, because `PScriptContext` is currently defined via `PDataRecord`. I think we should have two variants of `PScriptContext` and the rest of API: one Data-encoded and the other decoded, i.e. plain.

Plutarch/FFI.hs
L-as: I think there's an unsafe function`CompiledCode a -> CompiledCode b`. You could always do `unsafeCoerce` inside a PlutusTx script.

blamario:
There's a test suite at https://github.com/blamario/plutarch-with-ghc-8.10.7/blob/master/test/Spec.hs

Plutarch/FFI.hs
L-as: Can you explain your motivation? `POpaque` was literally made for this.

blamario: I've added `PUnit` which can provide the same functionality.

Plutarch/FFI.hs
```diff
@@ -0,0 +1,145 @@
...
+-- | Equality of inner types - Plutarch on the left and Haskell on the right.
+type p >~< t = PlutarchInner p PhorallPhantom ~~ PlutusTxInner t ForallPhantom
+
+type family a ~~ b :: Constraint where
+  ForallPhantom ~~ _ = ()
+  _ ~~ ForallPhantom = ()
+  Delayed a ~~ Delayed b = a ~~ b
+  a ~~ b = a ~ b
+
+type family PlutarchInner (p :: PType) (any :: PType) :: Type where
```
L-as, line 79: I still don't understand why this type is so complex. What is wrong with something like
```haskell
type family PlutarchToPlutusTx (p :: PType) = t | t -> p where
  PlutarchToPlutusTx PBool = BuiltinBool
  PlutarchToPlutusTx PInteger = Integer
  PlutarchToPlutusTx PByteString = BuiltinByteString
  PlutarchToPlutusTx (a :--> b) = PlutarchToPlutusTx a -> PlutarchToPlutusTx b
```
etc.

Right now AFAICT you have two type families that convert to some intermediate types *that are never actually used* but only used for checking equality.

L-as: In addition, what does `any` do?

Plutarch/FFI.hs
```diff
@@ -0,0 +1,145 @@
...
+  PlutarchInner PInteger _ = Integer
+  PlutarchInner PString _ = Text
+  PlutarchInner PByteString _ = ByteString
+  PlutarchInner PData _ = BuiltinData
+  PlutarchInner PUnit _ = ()
+  PlutarchInner PhorallPhantom _ = ForallPhantom
+  PlutarchInner (PAsData a :--> PAsData b) x = PlutarchInner (PData :--> PData) x
+  PlutarchInner (PAsData a :--> b) x = PlutarchInner (PData :--> b) x
+  PlutarchInner (a :--> b) x = PlutarchInner a x -> PlutarchInner b x
+  PlutarchInner (PDelayed a) x = Delayed (PlutarchInner a x)
```
L-as, line 90: I'm honestly not sure why you support this given that the resulting term can't be used from `PlutusTx`, nor is the inverse ever possible.

Plutarch/FFI.hs
```diff
@@ -0,0 +1,145 @@
...
+  PlutusTxInner BuiltinUnit _ = ()
+  PlutusTxInner ForallPhantom _ = ForallPhantom
+  PlutusTxInner (a -> b) x = PlutusTxInner a x -> PlutusTxInner b x
+  PlutusTxInner (Delayed a) x = Delayed (PlutusTxInner a x)
+  PlutusTxInner a x = TypeEncoding a (Rep a) x
+
+type TypeEncoding :: Type -> (Type -> Type) -> Type -> Type
+type family TypeEncoding a rep x where
+  TypeEncoding a (M1 D ( 'MetaData _ _ _ 'True) (M1 C _ (M1 S _ (K1 _ b)))) x = PlutusTxInner b x -- newtype
+  TypeEncoding a _ x = Delayed (PlutusTxInner (ScottFn (ScottList (SOP.Code a) x) x) x)
```
L-as, line 108: Do you need this?

L-as: I'm not sure why you need `ScottList`  and so on.

L-as (changes requested):

Plutarch/FFI.hs
L-as: I can't see how `PUnit` covers this? `PUnit` has only one legal term, `λx.x`. Any term is a legal `POpaque`.

Plutarch/FFI.hs
blamario: There's also
```
PlutarchToPlutusTx (PPair a b) = (PlutarchToPlutusTx a, PlutarchToPlutusTx b)
```
But `PPair a b >~< (a, b)` is just one of many possible ADT pairings between Plutarch and Haskell. We can't know them all in advance, because some may be user-defined. We do know however that they will be Scott-encoded, so the two type families that map them to Scott encodings are sufficient to cover them all.

The `any` serves the same purpose as the second type parameter of `PInner`, it acts as a universal type variable. This solution only covers types with a single forall-bound type variable, but it seems sufficient for most uses.

Plutarch/FFI.hs
blamario: That's there to bridge a mismatch between the Scott encodings of product types. PlutusTx always delays them for some reason:
```
testCase "import a pair" $
            printEvaluatedTerm
              ( foreignImport (PlutusTx.liftCode ("foo" :: BuiltinString, 4 :: Integer)) ::
                  Term _ (PDelayed (PPair PString PInteger))
              )
              @?= Right "(program 1.0.0 (delay (\\i0 -> i1 \"foo\" 4)))"
```

Plutarch/FFI.hs
L-as: So what's the purpose of having the intermediate types?

Plutarch/FFI.hs
blamario: I meant it covers the case _when you don't care about the type and only the fact that it's evaluated_.

L-as: That isn't true from a performance perspective. You need to change the term to make it of type `()`.

Plutarch/FFI.hs
blamario: The intermediate types are used only for checking equality of the two compiled types. Say a user has existing Haskell code dealing with
```
data MyRedeemer = MyRedeemer PubKeyHash Value
```
They want to be able to convert between `MyRedeemer` on the Haskell/PlutusTx side and
```
data PMyRedeemer s = PMyRedeemer (Term s PPubKeyHash) (Term s PValue)
```
on the Plutarch side. We could export an *open* type family `PlutarchToPlutusTx` as you suggest so that the user can add a new equivalence between the two redeemer types, but we wouldn't be able to actually test its correctness. The only proof of correctness is the equivalence of their Scott encodings. Thus the two type families.

Plutarch/FFI.hs
blamario: I tried adding `data Opaque = forall a. Opaque a` to mirror `POpaque` on the Haskell side, but `plutus-tx-plugin` can't handle the constructor:
```
GHC Core to PLC plugin: E043:Error: Reference to a name which is not a local, a builtin, or an external INLINABLE function: Type variable: a
Context: Compiling type: a
Context: Compiling data constructor type: Plutarch.FFI.Opaque
```

L-as:
You probably want to rebase and target staging

Plutarch/FFI.hs
blamario: Because I don't see how `CompiledCode a -> ClosedTerm POpaque` improves things. I already have working bijections for most common types, including user-defined ones.

I can add `CompiledCode Any` <-> `ClosedTerm POpaque`. What that be enough?

blamario: I think I finally understand what you meant. Do I?

blamario: See the last commit.

L-as:
Yep! That's it. FWIW I think `Any` is what you want to use rather than `Void`, since otherwise it's unsound. You can define your own `Any` if the one from base doesn't work probably. If that doesn't work either, it's possible that you can use (), but it depends on how pattern matches on it are compiled. Otherwise going from Plutarch to PlutusTx in this opaque way is unsound unfortunately.

blamario:
Plutus's `Any` is defined in `plutus-ledger` which is not currently a Plutarch dependency, can I add it?

L-as:
It won't compile

L-as:
`data Any = forall a. Any a` probably

L-as:
Nevermind, this won't work. Hmm.

L-as:
What happens if you use https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-Exts.html#t:Any with plutus-tx?

blamario:
It compiles.

blamario:
But that seems pretty dangerous, as this also compiles:
```
test :: CompiledCode Any -> CompiledCode a
test = coerce
```

L-as:
It won't compile with GHC 9.2

L-as:
In any case it's isomorphic to `Void` and isn't like the `Any` from base

L-as:
That's a bug in Plutus! I'll file an issue upstream

blamario:
So what's unsound about `CompiledCode Void`?

blamario:
The Plutus bug is the wrong role of the `CompiledCode` parameter I suppose? Perhaps I just leave out the `opaqueExport` until they fix it.

L-as:
> So what's unsound about CompiledCode Void?

Given `compile $$(\case{})` and `CompiledCode Void`, you can get `CompiledCode Void -> CompiledCode a` again AFAICT.

blamario:
Well yeah, but that's just the same Plutus bug. I mean what makes `CompiledCode Void` any worse than `CompiledCode Any`?

L-as:
It's slightly different, in that fixing the roles doesn't fix that. Using `Void` is worse than using `Any` from `base`, because you can't cause the unsoundness with `Any` from `base`.

L-as:
Can you rebase this and retarget to staging?

PR #266, opened 2022-02-09
Add `packages.${system}.haddock` attribute for combined haddock

L-as:
Unfortunately, it seems that we can't add all packages to this,
it errs with `Couldn't get the project and version from the html`,
and I'm not sure why.

Do `nix build .#haddock` to test. This is self-contained and can be
hosted on GitHub Pages.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -602,6 +602,32 @@
                 nix --extra-experimental-features 'nix-command flakes' run .#benchmark-diff -- before.csv after.csv
               '';
             }
+          )
+          gh-pages = hci-effects.runIf (src.ref == "refs/heads/master") (
+            hci-effects.mkEffect {
+              src = self;
+              buildInputs = with pkgs; [ openssh git ];
+              secretsMap = {
+                "ssh" = "ssh";
+              };
+              effectScript =
+              let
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -479,16 +479,61 @@
...
+                * "Plutarch.Prelude"
+                * "Plutarch"
+            '';
+          };
+        };
     in
     {
       inherit extraSources cabalProjectLocal haskellModule tools;
 
+      nixpkgsDebug = perSystem nixpkgsFor;
```
srid, line 527: ?

L-as: Thank you for catching this!

L-as:
When this is merged, the documentation should be available on https://plutonomicon.github.io/plutarch/

TotallyNotChase:
Is this just for haddock or is it possible to get hoogle going as well?

L-as:
It's just for Haddock. Hoogle would require a ton of effort to get running I think, and I'm not even sure it works without a server.

srid:
Running Hoogle is pretty straightforward in Nix, but we need a place to run it, and deploy it via CI/CD.

CD can be obviated with https://docs.cachix.org/deploy/ - and should work on local machines as well. Cachix Deploy however was in beta, last I checked. So YMMV.

Issue #245, opened 2022-02-07
Add example for hashing script and serialising script to disk


L-as:
This is quite important

L-as:
Maybe we can just use sha256 on this.

PR #267, opened 2022-02-10
APIs for constructing, compiling, serialising & hashing real scripts, with examples

t1lde:
For Issue #245 

Added:

- `Validator`, `StakeValidator`, `MintingPolicy` types in `Plutarch.V1.Api`, analogous to Plutus API types, but also including the Plutarch `ClosedTerm`s.
- Utils for constructing script types, wrapping the arguments (ala `wrapValidator` & co )
- API for serializing scripts via `Codec.Serialise`
- API for hashing scripts with `Blake2b_224`, using the correct Plutus V1 prefix.
  - I have confirmed this gives the same results as `Ledger.Scripts` 
 - Test example constructing a Validator, MintingPolicy & StakeValidator, serializing, hashing & writing to files. In `Examples.Scripts`
 - `punsafeFromData` & `punsafeAsData` in `Plutarch.Builtins` - convenient for coercing & converting script arguments

Not Included: 

- Obtaining `CIP-0019` addresses (addr1 and etc formats)
- Exposing more useful crypto functions - Blake2b_256 for DatumHash, Blake2b_224 for PubKeyHash, Sha3_256, Sha2_256 from PlutusTx Builtins


Plutarch/Api/Internal/Scripts.hs
```diff
@@ -0,0 +1,29 @@
...
+import qualified Plutus.V1.Ledger.Scripts as Plutus
+import qualified PlutusTx.Builtins as PlutusTx
+
+-- | Serialise a Script to (Lazy) ByteString
+serialiseScript :: Plutus.Script -> Lazy.ByteString
+serialiseScript = serialise
+
+-- | Hash a Script with the given version prefix
+hashScriptWithPrefix :: ByteString -> Plutus.Script -> Plutus.ScriptHash
+hashScriptWithPrefix prefix scr =
```
t1lde, line 24: pre-empting V2 with "\x02" prefix

Plutarch/Api/V1/Scripts.hs
```diff
@@ -64,3 +113,230 @@ deriving via
...
+pwrapValidatorAsData = pwrapValidatorWith punsafeAsData punsafeAsData punsafeAsData
+
+{- |
+  Construct a `PValidator`, assuming the `PData` arguments are
+  of the correct form, wrapping them with `pfromData`.
+
+  Shorthand for:
+  @pwrapValidatorWith punsafeFromData punsafeFromData punsafeFromData@
+-}
+pwrapValidatorFromData ::
```
t1lde, line 214: These are *technically* unsafe, though it's perfectly valid to use them anyway in many cases...
Should they have `unsafe` in their names somehow? `punsafeWrapValidatorFromData` is pretty long...

Plutarch/Api/V1/Scripts.hs
```diff
@@ -64,3 +113,230 @@ deriving via
...
+  Shorthand for:
+  @pwrapStakeValidatorWith punsafeAsData punsafeAsData@
+-}
+pwrapStakeValidatorAsData ::
+  ( forall s.
+    Term s (PAsData redeemer) ->
+    Term s (PAsData ctx) ->
+    Term s p
+  ) ->
+  ClosedTerm PStakeValidator
```
t1lde, line 307: Equivalent to pwrapMintingPolicy... we could avoid the redundancy but it's nice to have separately named versions IMO

examples/Examples/Scripts.hs
```diff
@@ -0,0 +1,248 @@
...
+  if the parameter PubKeyHash signs the transaction.
+-}
+authorizedStakeValidator ::
+  forall s.
+  ClosedTerm (PAsData PPubKeyHash) ->
+  Term s PData ->
+  Term s PScriptContext ->
+  Term s PUnit
+authorizedStakeValidator authHash _redeemer ctx =
+  let sigs :: Term s (PBuiltinList (PAsData PPubKeyHash))
```
t1lde, line 97: equivalent to, and hashes the same as authorizedPolicy, interestingly enough.

Plutarch/Api/Internal/Scripts.hs
```diff
@@ -0,0 +1,29 @@
...
+import Data.ByteArray (convert)
+import Data.ByteString (ByteString)
+import qualified Data.ByteString.Lazy as Lazy
+
+import qualified Plutus.V1.Ledger.Scripts as Plutus
+import qualified PlutusTx.Builtins as PlutusTx
+
+-- | Serialise a Script to (Lazy) ByteString
+serialiseScript :: Plutus.Script -> Lazy.ByteString
+serialiseScript = serialise
```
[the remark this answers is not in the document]
t1lde: Good point, yes it probably should. I can get that in this PR

L-as (changes requested):
Looks great, thanks!

Plutarch/Api/V1/Scripts.hs
```diff
@@ -64,3 +113,230 @@ deriving via
...
+-}
+pwrapValidatorAsData ::
+  ( forall s.
+    Term s (PAsData datum) ->
+    Term s (PAsData redeemer) ->
+    Term s (PAsData ctx) ->
+    Term s p
+  ) ->
+  ClosedTerm PValidator
+pwrapValidatorAsData = pwrapValidatorWith punsafeAsData punsafeAsData punsafeAsData
```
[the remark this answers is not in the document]
t1lde: there are a good amount of cases where you _would_ want to just assume the shape of Data, without checking it. 
Though I agree doing so should be more explicit, and not encouraged as a default...
I don't know if I like `PAsDataMaybe` as a wrapper - wouldn't we then need to (transitively!) wrap everything 'downstream' in a 'maybe', similarly?

t1lde:
I removed the extra types & unsafe wrappers. 

How to handle `PData` with unknown shape is a can of worms to be opened elsewhere - for now I'm happy to have users think careful about how to safely use the `PData` args with existing APIs.

Plutarch/Api/V1/Scripts.hs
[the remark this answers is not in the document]
L-as: Probably, but it's usually not going to be a deeply nested type. A possible vulnerability that might not be obvious to users, is setting a global UTXO that contains some funds, the datum of which you can change by supplying a redeemer that is a list of integers. The new datum is the old datum, with the list of integers appended to the old list of integers. The issue is that with this, users would do `punsafeFromData` or use whatever abstraction you have on top of it, and assume as with PlutusTx that it in fact validates the data, when it does not. This would result in a UTXO that contains the global funds, with an invalid datum. This might make it unspendable, and thus be a vector for an attack.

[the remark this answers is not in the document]
t1lde: Yeah, I agree on the worry there. I removed the unsafe functions.
It's probably best left up to users to write explicit checks as necessary for their Datum/Redeemer types.

t1lde:
merged srid/testing & ported the tests. should be ready after #276 , module further merge conflicts

L-as:
Changelog says `serialiseValidator`, but I can't see any such function?

t1lde:
> Changelog says `serialiseValidator`, but I can't see any such function?

ah yeah, that ended up being removed along with the extra Validator types. will  update the changelog

L-as:
Shouldn't we have a `serialiseScript` function?

t1lde:
> Shouldn't we have a `serialiseScript` function?

It would just be a less-polymorphic synonym of `serialise`, but using it is non-obvious so I can see why it'd be nice to have.

Should we then also have `deserialiseScript`?

t1lde:
> It would just be a less-polymorphic synonym of `serialise`, but using it is non-obvious so I can see why it'd be nice to have.

Or should `serialiseScript` use the hex-encoding by default?

plutarch-test/src/Plutarch/ScriptsSpec.hs
```diff
@@ -0,0 +1,225 @@
...
+
+spec :: Spec
+spec = do
+  describe "scripts" $ do
+    describe "auth_validator" $ do
+      golden PrintTerm authValidatorTerm
+      it "serialization" $
+        pureGoldenTextFile
+          ("goldens" </> "scripts.auth_validator.plutus.golden")
+          validatorEncoded
```
srid, line 55: You can use `getTestDescriptionPath` to obviate hardcoding the path here, which is error-prone. It would be even better if we just used `goldens` or `golden` from `Plutarch.Test` here.  Maybe expand `PlutarchGolden` to include `ScriptEncoding`?

nixinator (commented):
that's the biggest PR i even seen on github, you win a cookie.

t1lde:
> @t1lde I propose removing [Examples/Scripts.hs](https://github.com/Plutonomicon/plutarch/compare/srid/testing..tilde/bench-unapplied?expand=1#diff-abf2949b9161269233c029bfb7305464532ba3998461914b9c22a7f8da7bdfc9) since there already is a [plutarch-test/src/Plutarch/ScriptsSpec.hs](https://github.com/Plutonomicon/plutarch/compare/srid/testing..tilde/bench-unapplied?expand=1#diff-5dc392de12f21e5c0fc0f4b7e027a9298c1b2683aa953526f28792737a5748d0R1).

meant to remove that already, oops

plutarch-test/src/Plutarch/ScriptsSpec.hs
t1lde: I went ahead and added `getGoldenFilePrefix` & `goldenFilePath` to make this a bit easier for tests.

t1lde: Don't think it's necessary to expand the existing golden tests with encoding, though

srid:
image

Are these used? Shouldn't they exist in `plutarch-test/goldens` instead of `examples/goldens`? `./examples` will eventually go away.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,224 @@
...
+-- | Get a golden filename prefix from the test description path
+getGoldenFilePrefix ::
+  forall (outers :: [Type]) (inner :: Type).
+  TestDefM outers inner String
+getGoldenFilePrefix =
+  T.unpack . T.intercalate "." . drop 1 . reverse <$> getTestDescriptionPath
+
+-- | Get the golden file name given the basepath, an optional suffix and a name
+goldenFilePath :: FilePath -> String -> Maybe String -> FilePath
+goldenFilePath base name suffix =
```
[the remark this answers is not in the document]
t1lde: I just propagated the argument from `goldens'` :shrug:

t1lde:
> image
> 
> Are these used? Shouldn't they exist in `plutarch-test/goldens` instead of `examples/goldens`? `./examples` will eventually go away.

meant to delete these

L-as:
You should merge master in

t1lde:
should be ready to merge now

Plutarch/Field.hs
```diff
@@ -0,0 +1,82 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Plutarch.Field (PField (..), FieldName (..), pfield) where
```
L-as, line 5: ?

L-as: Did your merge go bad?

plutarch.cabal
```diff
@@ -174,7 +177,9 @@ test-suite examples
     , plutus-ledger-api
     , plutus-tx
     , rank2classes       >=1.4.4
+    , serialise
     , tasty
+    , tasty-golden
```
L-as, line 182: We're using sydtest. Why are we adding this?

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -170,26 +172,38 @@ golden pg p =
...
 goldens pg ps = do
-  testAncestors <- fmap (drop 1 . reverse) $ getTestDescriptionPath
-  let name = T.unpack $ T.intercalate "." testAncestors
+  name <- getGoldenFilePrefix
   describe "golden" $ do
     -- Golden test for UPLC
     when (hasPrintTermGolden pg) $ do
       it "uplc" $
-        pureGoldenTextFile ("goldens" </> name <> ".uplc.golden") $
+        pureGoldenTextFile (goldenFilePath "goldens" name "uplc") $
```
L-as, line 180: Is this just a general clean-up?

CHANGELOG.md
```diff
@@ -38,6 +38,21 @@
...
 
+- Added APIs for constructing, compiling, serialising & hashing Plutarch scripts.
+
+  Type synonyms for Plutarch-typed scripts `PValidator`,`PMintingPolicy` & `PStakeValidator`.
+  
+  `mkValidator`, `mkStakeValidator` & `mkMintingPolicy` functions, for creating Plutus API compatible scripts.
+
+  `validatorHash`, `mintingPolicySymbol` & `stakeValidatorHash` to obtain script hashes.
+
+  `tryDecodeHex` & `encodeSerialise` (re-exported from `Data.Aeson.Extras`), for encoding & decoding Scripts and other Plutus values as hex.
```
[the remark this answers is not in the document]
t1lde: Probably, yeah

plutarch-test/src/Plutarch/Test.hs
t1lde: it's used as a helper for the new tests, too

Plutarch/Field.hs
t1lde: looks like it, I have no idea how this happened.

plutarch.cabal
t1lde: meant to remove this

Plutarch/Field.hs
t1lde: I think vscode must've restored an old cached file without asking :upside_down_face:

TotallyNotChase:
Do we have safe ways to convert typed validations functions into `PValidator` (and similar wrappers)?

t1lde:
> Do we have safe ways to convert typed validations functions into `PValidator` (and similar wrappers)?

Safe is complicated. It's up to a user to understand what invariants may have been enforced by other validators.  #277 should help though

TotallyNotChase:
> Safe is complicated. It's up to a user to understand what invariants may have been enforced by other validators. #277 should help though

We could, at the very least, have `ClosedTerm (PAsData a :--> PAsData b :--> PAsData PScriptContext :--> c) -> ClosedTerm PValidator` and similar functions. It'd make it easier to work with and probably simplify the documentation/demonstration. I do also think that we should use `PAsData` for the datum/redeemer/context args.

t1lde:
> We could, at the very least, have `ClosedTerm (PAsData a :--> PAsData b :--> PAsData PScriptContext :--> c) -> ClosedTerm PValidator` and similar functions.

This is rather unsafe, see @L-as 's earlier [comment](https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109)

TotallyNotChase:
> This is rather unsafe, see @L-as 's earlier [comment](https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109)

Huh, that's weird - so there's another layer of verification outside the control of the type system - didn't know about that. I suppose this means that smart contract writers are *expected* to verify the format of their datums/redeemers before assuming them to be what they are (as proclaimed by the type system)? Is this also what happens in PlutusTx land as well?

t1lde:
> > This is rather unsafe, see @L-as 's earlier [comment](https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109)
> 
> Huh, that's weird - so there's another layer of verification outside the control of the type system - didn't know about that. I suppose this means that smart contract writers are _expected_ to verify the format of their datums/redeemers before assuming them to be what they are (as proclaimed by the type system)? Is this also what happens in PlutusTx land as well?

PlutusTx cheats (when you use `wrapValidator`) by wrapping the arguments with `unsafeFromData`, which is the kind of inefficiency we want to make it possible to avoid in Plutarch.

L-as:
> > This is rather unsafe, see @L-as 's earlier [comment](https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109)
> 
> Huh, that's weird - so there's another layer of verification outside the control of the type system - didn't know about that. I suppose this means that smart contract writers are _expected_ to verify the format of their datums/redeemers before assuming them to be what they are (as proclaimed by the type system)? Is this also what happens in PlutusTx land as well?

We probably want to make it possible to have a`PAsDataMaybe` of some kind, such that we declare that it *may* be what it is saying it is, but may very well not be. If we had such an analogue for `PIsDataRepr`, then field access would also give you `PAsDataMaybe` or something like that. The issue is we might want to have both `PAsDataMaybe` and `PAsData` versions of the same data type, meaning we need some form of HKD probably.

TotallyNotChase:
> We probably want to make it possible to have a`PAsDataMaybe` of some kind, such that we declare that it _may_ be what it is saying it is, but may very well not be. If we had such an analogue for `PIsDataRepr`, then field access would also give you `PAsDataMaybe` or something like that. The issue is we might want to have both `PAsDataMaybe` and `PAsData` versions of the same data type, meaning we need some form of HKD probably.

Could you explain why we need `PAsDataMaybe`? Can't we solve all this once we have `PMaybeFrom`? I assumed that would essentially validate the encoding and give back the expected type.

L-as:
> > We probably want to make it possible to have a`PAsDataMaybe` of some kind, such that we declare that it _may_ be what it is saying it is, but may very well not be. If we had such an analogue for `PIsDataRepr`, then field access would also give you `PAsDataMaybe` or something like that. The issue is we might want to have both `PAsDataMaybe` and `PAsData` versions of the same data type, meaning we need some form of HKD probably.
> 
> Could you explain why we need `PAsDataMaybe`? Can't we solve all this once we have `PMaybeFrom`? I assumed that would essentially validate the encoding and give back the expected type.

That's not as efficienct.

L-as:
You always need to validate the redeemer in a sensible validator, because it should have no unused information. For the datum however, it is very often the case that we *don't* need to validate it, because it might already have been validated by something else.

t1lde:
Can we merge this yet? I think everything is resolved.

TotallyNotChase:
One more question:
```hs
type PValidator = PData :--> PData :--> PScriptContext :--> POpaque
```
can we use `PAsData PScriptContext` instead?

t1lde:
> One more question:
> 
> ```haskell
> type PValidator = PData :--> PData :--> PScriptContext :--> POpaque
> ```
> 
> can we use `PAsData PScriptContext` instead?

You could (I had it this way initially) but there's no reason to, since `pfromData` & `pdata` are both no-ops for `PIsDataReprInstances`-based `PIsData` instances

TotallyNotChase:
> You could (I had it this way initially) but there's no reason to, since `pfromData` & `pdata` are both no-ops for `PIsDataReprInstances`-based `PIsData` instances

For sure, there's no semantic difference. But I am curious whether it'd be better to consistently use `PAsData` everywhere whenever a data encoded value is expected, without having to know whether or not the `a` in `PAsData a` has a `PIsDataRepr` instance. It might make it easier to understand for users and possibly make documentation easier.

L-as (changes requested):
Sorry for being inactive! Yeah I don't see any issues, but I'm wondering about the tasty-golden dependency

Plutarch/Api/V1/Scripts.hs
```diff
@@ -64,3 +65,5 @@ deriving via
   (DerivePConstantViaNewtype Plutus.ValidatorHash PValidatorHash PByteString)
   instance
     (PConstant Plutus.ValidatorHash)
+
+-----------------------------
```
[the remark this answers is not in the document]
t1lde: It was originally sectioning functions that have now been moved, I can remove rq (just to avoid the chance of it causing a conflict elsewhere somehow)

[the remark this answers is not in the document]
t1lde: I did not expect that to break the format check :upside_down_face: , wow

TotallyNotChase:
> I can't vouch for _everywhere_ part, but currently the only way to write a non-trivial
> 
> ```
> type PValidator = PData :--> PData :--> PScriptContext :--> POpaque
> ```
> 
> is to use `punsafeCoerce`. I think it would make sense to mirror the Plutus design and have
> 
> ```
> type PTypedValidator dat red = PAsData dat :--> PAsData red :--> PAsData PScriptContext :--> POpaque
> ```

As I was made aware of by Tilde and Las, through the discussions on this PR - `PTypedValidator` would be unsafe as a mere type alias. But Tilde and Las can elaborate on that much better.

I wanted to clarify that by "everywhere", I really only mean validator/minting policy arguments. As those are the only places where one would explicitly expect data encoded arguments. For utility functions and such, no one really cares about the encoding - so they can go ahead and use whatever. It's the cross-language boundaries that concern me, from a documentation and comprehensibility perspective.

L-as:
> > > You could (I had it this way initially) but there's no reason to, since `pfromData` & `pdata` are both no-ops for `PIsDataReprInstances`-based `PIsData` instances
> > 
> > 
> > For sure, there's no semantic difference. But I am curious whether it'd be better to consistently use `PAsData` everywhere whenever a data encoded value is expected, without having to know whether or not the `a` in `PAsData a` has a `PIsDataRepr` instance. It might make it easier to understand for users and possibly make documentation easier.
> 
> I can't vouch for _everywhere_ part, but currently the only way to write a non-trivial
> 
> ```
> type PValidator = PData :--> PData :--> PScriptContext :--> POpaque
> ```
> 
> is to use `punsafeCoerce`. I think it would make sense to mirror the Plutus design and have
> 
> ```
> type PTypedValidator dat red = PAsData dat :--> PAsData red :--> PAsData PScriptContext :--> POpaque
> ```

There is no way to make such an abstraction safe. #303

PR #270, opened 2022-02-10
#235: automatic `pfromData`

### LINKED ISSUE #235: Add utility `pfield` and `pletFields` variants that automatically `pfromData` the result

MangoIV:
## Changes

- this resolves #235 
- changed the method `pfield` to be polymorphic over the output type 
- added a new method `hrecdfield` that wraps `hrecfield` but is polymorphic over the output type
- the record dot instance now works with that new method 
- I also adjusted some of the examples to show that it works with both the implicit and the explicit `pfromData`

## Issues

- implicit `pfromData` for `pfield` doesn't work well in large expressions as GHC isn't clever enough to infer the type in some cases, type annotations help in those cases. If anybody comes up with a trick how to ship around that problem, it would be much  appreciated :)


Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# OPTIONS_GHC -Wno-orphans#-}
```
L-as, line 4: Formatting

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -100,6 +122,41 @@ instance
...
+    even if the result it not originally wrapped into it 
+-}
+class PNoAsData (a::PType) (b::PType) | a -> b where 
+  premoveAsData :: Term s a -> Term s b
+
+instance {-# OVERLAPPABLE #-} PNoAsData a a where 
+  premoveAsData = id
+
+instance {-# OVERLAPPABLE #-} a ~ PAsData b => PNoAsData a b  where  
+  premoveAsData = punsafeCoerce
```
L-as, line 137: This isn't legal? It does `premoveAsData :: Term s (PAsData b) -> Term s b`. You want `pfromData`.

flake.nix
```diff
@@ -1,6 +1,7 @@
 {
   description = "plutarch";
 
+  inputs.nixpkgs-2111 = { url = "github:NixOS/nixpkgs/nixpkgs-21.11-darwin"; };
```
L-as, line 4: I think we want to use `follows` here.

t1lde (dismissed):
Looks good, though I'm a *little* nervous about the type inference issues...

A (syntax-abusive) idea would be to add `HasField` instances for field names with pre/post-fixed symbols, like:

```haskell
x.field_data
x.field_nodata
```
probably an unpopular idea though :sweat_smile:

L-as (changes requested):
Looks great!

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -100,6 +122,41 @@ instance
   type PFields (PAsData a) = PFields a
   ptoFields = ptoFields . pfromData
 
+
+{- | 
+    always makes sure the result is not wrapped in PAsData, 
+    even if the result it not originally wrapped into it 
+-}
+class PNoAsData (a::PType) (b::PType) | a -> b where
```
TotallyNotChase, line 130: Where is this used?

flake.nix
MangoIV: I think I'll just remove the change completely -- this PR shouldn't be concerned with it anyway. Especially since it's only a bug in Nix.

Plutarch/DataRepr/Internal/Field.hs
MangoIV: nowhere yet but I asked @L-as and he said that it might be useful somewhere in the future. it's basically the other class but vice versa.

nixinator (approved):

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -12,20 +14,38 @@ module Plutarch.DataRepr.Internal.Field (
...
+  PFromDataable,
+  pmaybeFromAsData,  
+  PNoAsData,
+  premoveAsData, 
+
   -- * Re-exports
   HRec (..),
   Labeled (Labeled, unLabeled),
   hrecField,
+  hrecdField,
```
TotallyNotChase, line 27: Let's export this from `Plutarch.DataRepr`. Would also be nice to export it from `Plutarch.Prelude`. Eventually the `hrecField` export should be removed from `Plutarch.Prelude` though.

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
```
TotallyNotChase, line 3: You can do `./bin/format` and it'll take care of all the formatting for you!

TotallyNotChase (changes requested):
Could you post some examples where type inference is less than ideal? We should track it as an issue if we can't resolve them immediately.

MangoIV:
yes, in the examples I added a type signature to where it's not ideal: 
https://github.com/MangoIV/plutarch/blob/b5a85197479d02f69a7323787830e48616b17291/examples/Examples/Api.hs#L181

Plutarch/DataRepr/Internal/Field.hs
TotallyNotChase: Actually I think it's better to change `hrecField` to have the same type signature as the current `hrecdField` and not have `hrecdField` at all.

L-as:
What's the error message when it goes wrong?

Plutarch/DataRepr/Internal/Field.hs
TotallyNotChase: `HRec` is meant specifically for plutus types regardless, hence it's within `DataRepr.Internal`. No need to worry about that. We want to have consistent behavior for users, ideally. See discord for my concerns.

For the cyclic dep, you can put the `PFromDataable` stuff into a separate module within `DataRepr.Internal`, everything directly related to `HRec` (including the `HasField` instance) can then be within `Internal.HRec`.

MangoIV:
This would be the error. It basically picks the "wrong" instance in this case. (So it's actually not an issue with instance resolution.) I actually had it pick the other instance more eagerly in my first try but that would break all the uses of `pfield`with `pfromData` so I changed it to pick the other instance more eagerly. 
``` • Couldn't match type: PBuiltinList (PAsData PPubKeyHash)
                     with: PAsData PPubKeyHash
        arising from a functional dependency between:
          constraint ‘Plutarch.DataRepr.Internal.Field.PFromDataable
                        (Plutarch.DataRepr.Internal.PUnLabel
                           (Plutarch.DataRepr.Internal.HList.IndexList
                              @PLabeledType
                              (Plutarch.DataRepr.Internal.PLabelIndex
                                 "signatories"
                                 (Plutarch.DataRepr.Internal.Field.PFields (PAsData PTxInfo)))
                              (Plutarch.DataRepr.Internal.Field.PFields (PAsData PTxInfo))))
                        (PAsData (PAsData PPubKeyHash))’
            arising from a use of ‘pfield’
          instance ‘Plutarch.DataRepr.Internal.Field.PFromDataable
                      a (PAsData a)’
            at <no location info>
    • In the first argument of ‘(#)’, namely ‘pfield @"signatories"’
      In the expression: pfield @"signatories" # ctx.txInfo
      In an equation for ‘signatories’:
          signatories = pfield @"signatories" # ctx.txInfo
    |
197 |   let signatories = pfield @"signatories" # ctx.txInfo
    |                     ^^^^^^```

File: .envrc [hunk 7]
```diff
@@ -0,0 +1 @@
+use flake
```
TotallyNotChase, line 1: I assume we don't need this?

Plutarch/DataRepr/Internal/Field.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE FunctionalDependencies #-}
```
TotallyNotChase, line 2: I *believe* this is redundant. I forgot to remove it during my cleanup.

TotallyNotChase (approved):
Looks good! You should make an issue so we can track the type inference issues post-merge.

.envrc
MangoIV: we do not need it but there's no `.envrc` yet and this gives you instant `nix` shell with `direnv-nix`, I quite like it. It's also practical for people who use vscode with that `direnv` plugin that just loads the environment of the project you're in. But, of course, you decide ;)

Plutarch/DataRepr/Internal/Field.hs
MangoIV: it was indeed redundant, thanks!

.envrc
MangoIV: Is this as yes? :D

TotallyNotChase: I don't mind it ¯\\\_(ツ)\_/¯

nixinator (approved):

MatthewCroughan (approved):

L-as (approved):
LGTM

srid:
@MangoIV Is [this](https://github.com/Plutonomicon/plutarch/pull/276#issuecomment-1039178561) expected? Should we apply the new golden?

TotallyNotChase:
> Is [this](https://github.com/Plutonomicon/plutarch/pull/276#issuecomment-1039178561) expected? Should we apply the new golden?

Just to chime in, there must be something else causing that change. This PR shouldn't cause any change on the existing compilation outputs whatsoever. `cabal test` on master succeeds as well.

But yes, whatever compilation output you get on `master` for those particular tests, should be the expected ones.

srid:
Yup, just to be clear: it is a new test in my branch. `master` passed only because it doesn't have that test.

MangoIV:
I have elaborated some more on Discord, I don't think it's caused by the PR I did: The issue persists if I introduce a new function 

```haskell
pfield2 ::
forall name p s a as n.
  ( PDataFields p
  , as ~ (PFields p)
  , n ~ (PLabelIndex name as)
  , KnownNat n
  , a ~ (PUnLabel (IndexList n as))
  , PIsData a 
  ) =>
  Term s (p :--> a)
pfield2 = plam $ \i ->
  pfromData $ pindexDataRecord (Proxy @n) $ ptoFields @p i
  ```
  
  that always removed the `PAsData`. This also changes the output, while it should not.
  This is the reason why I think the issue is actually not causes by this PR.

PR #274, opened 2022-02-10
Upstream some list functions

Geometer1729:


Plutarch/List.hs
```diff
@@ -363,3 +370,134 @@ plistEquals =
...
+      (pcon PNothing)
+      xs
+
+psort :: POrd a => Term s (PList a :--> PList a)
+psort = timSort
+
+-- TODO: decide on default sort
+-- probably by benchmarking them
+
+mergesort :: POrd a => Term s (PList a :--> PList a)
```
L-as, line 413: Naming

Plutarch/List.hs
```diff
@@ -363,3 +370,134 @@ plistEquals =
...
+              (pcons # y #$ self # xs # ys')
+
+-- most of the tricks from timsort are not implemented here
+-- just the central strategy of finding runs of
+-- consecutive data and then merging them together
+timSort :: POrd a => Term s (PList a :--> PList a)
+timSort = phoistAcyclic $
+  plam $ \xs -> merge2 #$ timSplit # xs
+
+timSplit :: POrd a => Term s (PList a :--> PList (PList a))
```
L-as, line 457: Naming here too

plutarch.cabal
```diff
@@ -167,19 +167,26 @@ test-suite examples
...
+    Properties.Utils
     Utils
 
   build-depends:
     , aeson
     , base
     , bytestring
     , generics-sop
+    , hedgehog
+    , mtl
```
L-as, line 182: Why mtl?

L-as (changes requested):
@srid What do you think about the testing?

Plutarch/List.hs
Geometer1729: Do you want all functions name psomeFunctionName or just exposed ones?

srid:
Just spoke with Brian. The haskPlutEquiv stuff is valuable for property testing Plutarch functions by piggybacking on their Haskell equivalent functions (it avoids having to write manual tests for each of them). 

I feel [Properties/Utils.hs](https://github.com/Plutonomicon/plutarch/pull/274/files#diff-f8ff3e451db7fa0415b03a4d0d75303fc0234aa9d0296b4c42def5519a3b0403)'s export interface could be simplified though. But it is probably not important right now, as all of these testing code will have to moved to `./plutarch-test` (and thereby refactored) anyway after #276 is merged.

Plutarch/List.hs
L-as: Any export that is a Plutarch function or type from any module should be prefixed with `p`.

plutarch-test/src/Util.hs
```diff
@@ -0,0 +1,286 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Util (
```
srid, line 5: You want to move this to `Plutarch.Test.Property` inasmuch as the whole thing is to do with property testing.

plutarch-test/src/Gen.hs
```diff
@@ -0,0 +1,54 @@
+module Gen (
```
srid, line 1: Should be `Plutarch.Test.Property.Gen`

plutarch-test/src/Plutarch/ListPropSpec.hs
```diff
@@ -0,0 +1,108 @@
...
+
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+import Plutarch.Test (PlutarchGolden (All), goldens)
+import Test.Syd (Spec, describe, it)
+import Test.Syd.Hedgehog ()
+
+spec :: Spec
+spec = describe "list spec" $ do
```
srid, line 21: By convention we don't have spaces in spec descriptions that get used as golden filenames. This should be therefore `describe "list"`.

srid (changes requested):

Geometer1729:
Moved to #304 so it can be from a different branch.

Issue #269, opened 2022-02-10
Measure sizing of non-applied scripts in Benchmark


t1lde:
The sizes measured by plutarch-benchmark are suitable for measuring in isolation, but when it comes to real scripts, we actually want to measure the size of the script *before* any arguments are applied - afaik this is how they are actually sent in the tx witnesses (also packaging the Datum & Redeemers separately), with the verifier applying the script arguments.

I have found that  ScriptContext itself adds upwards of 300bytes, even in pretty small examples, which is actually *larger* than many scripts.

TotallyNotChase:
We could have a `benchWithArgs`.

L-as:
Relevant to what @srid is doing, seems like there's overlapping work.

t1lde:
I got a *little* start on it today, it's pretty simple so shouldn't take too much to adapt to plutus-tests

srid:
I just realized I was doing something like this for the Bool spec in #276:

```haskell
      goldens
        All
        [ ("lam", popaque pnot)
        , ("app", popaque $ pnot #$ pcon PTrue)
        ]
```

<img width="923" alt="image" src="https://user-images.githubusercontent.com/3998/153639829-36fca4f0-ad2c-4cb7-ba02-17ee485802d6.png">

L-as:
@t1lde I get the motivation now, so you want to measure the size without the `ScriptContext` constant I assume? Wouldn't that just mean making a test without a concrete `ScriptContext` and then another that uses the first one but applies a concrete `ScriptContext`? Not sure we need some special function fori t.

PR #276, opened 2022-02-11
New tests: goldens; unified benchmark+examples

srid:
- [x] Use sydtest, and start porting ./examples
- [ ] Finish rest of ./examples
- [ ] Port benchmarks to ./plutarch-test
- [ ] Delete examples and benchmarks (but keep `plutarch-benchmark` library)
- [x] Make plutarch-test library


srid:
@L-as What are your thoughts on merging this right now -- so that when people open new PRs with tests, they can start adding the tests to ./plutarch-tests instead -- with the rest of the tasks done as separate PR?

https://github.com/Plutonomicon/plutarch/issues/269#issuecomment-1036411927

MangoIV:
I'm so sorry, I accidentally pushed to your branch

srid:
Ran CI locally, and it passes.

@L-as I've deleted already ported over code. Let's merge this now, and ask current/ future PR authors to add new tests to `./plutarch-test` going forward.

MatthewCroughan (approved):
Once @NorfairKing provides a flake for his projects, we can reduce the complexity of this code.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,139 @@
+{-# LANGUAGE ImpredicativeTypes #-}
```
L-as, line 1: lol

plutarch-test/src/Spec.hs
```diff
@@ -0,0 +1,3 @@
+{-# OPTIONS_GHC -F -pgmF sydtest-discover -optF --no-main #-}
```
L-as, line 1: What does this do?

L-as (dismissed):
Looks great, thanks!

L-as:
@srid @MangoIV there are conflicts...

plutarch-test/src/Spec.hs
MangoIV: Test discovery, afaik

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,139 @@
...
+      it "uplc" $
+        pureGoldenByteStringFile ("goldens" </> nUplc) $
+          multiGolden ps $ \p ->
+            T.pack $ printTerm p
+    -- Golden test for Plutus benchmarks
+    when (hasBenchGolden pg) $
+      it "bench" $
+        pureGoldenByteStringFile ("goldens" </> nBench) $
+          multiGolden ps $ \p ->
+            TL.toStrict $ Aeson.encodeToLazyText $ benchmarkScript' $ compile p
```
NorfairKing, line 133: I would use `sydtest-aeson`'s function for this.

plutarch-test/src/Main.hs
```diff
@@ -0,0 +1,10 @@
+module Main (main) where
+
+import GHC.IO.Encoding (setLocaleEncoding, utf8)
+import Spec (spec)
+import Test.Syd (sydTest)
+
+main :: IO ()
+main = do
+  setLocaleEncoding utf8
```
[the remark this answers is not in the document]
L-as: Inside builds there is no locale IIRC

plutarch-test/src/Plutarch/Test.hs
srid: Some goldens contain multiple entries, eg:

```
❯ cat plutarch-test/goldens/bool.pnot.bench.golden 
lam {"exBudgetCPU":29873,"exBudgetMemory":200,"scriptSizeBytes":14}
app {"exBudgetCPU":387150,"exBudgetMemory":1401,"scriptSizeBytes":16}
```

Looks like `sydtest-aeson` shows a JSON-friendly diff. Can we get that for the above case? `pureGoldenJSONValueFileMany :: (Show a, Eq a, FromJSON a, ToJSON a) => FilePath -> Map k a -> GoldenTest a`?

srid: The reason why some goldens have multiple JSON values in them is to group benchmarks for closely related programs in the same file, and thus avoid having gazillion golden files most of which corresponding to tiny programs.

srid:
CI is still down; but tests pass locally.

plutarch-test/src/Plutarch/Test.hs
NorfairKing: @srid if we have the name for each of those things in the multi-golden, we could make a json object out of them, otherwise a list...

srid: Using json object of assoc list (`[String, a]`), `sydtest-aeson` is producing:

image

compared to the current:

image

I don't like how the latter (ie., current behaviour) puts both the entries (`app` and `lam`) in the _same_ line; at the same time I find the former to be too verbose. 🤔

NorfairKing: I like the `pureGoldenTextFile` one best. But it's going to be messed up if we change the keys to a different name. Up to you.

srid: Keys and order shouldn't change unless developers purposefully change their name/order in the test source. The above golden is for:

```haskell
      goldens
        All
        [ ("lam", popaque pnot)
        , ("app", popaque $ pnot #$ pcon PTrue)
        ]
```

L-as:
So there are "golden" files for the generated UPLC, but there should be two such generated outputs: pre-evaluation term and post-evaluation term.

srid:
> So there are "golden" files for the generated UPLC, but there should be two such generated outputs: pre-evaluation term and post-evaluation term.

#285

nixinator (commented):
wow,  this is even bigger ... Srid now wins a cookie, for the biggest commit i've ever seen! This amazing!

L-as:
nitpick: I think uplc-eval is better than uplc.eval, otherwise it's harder to mentally parse the file name IMO. TBF you can also keep it like this.

plutarch-test/src/Main.hs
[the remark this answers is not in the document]
DarthPJB: I don't actually know, I would however imagine this is a safeguard against GB18030 encoding on Chinese systems; most Chinese distros set it as the default at kernel-level; and it trickles down - I've seen more than my fair share of software break due to systems expecting UTF-8 being force-fed half-byte-codes.

a statement of reality is that all software is locale dependant once it starts manipulating byte-strings, it's just hand-waved by generalised UTF-8 usage.

cabal.project
```diff
@@ -1,5 +1,9 @@
 packages:
   ./.
-  ./plutarch-benchmark/plutarch-benchmark.cabal
+  ./plutarch-*
```
DarthPJB, line 3: I don't understand cabal properly yet; however wildcards always make me nervous.

Are there any implications here for future additions to 'plutarch' causing unexpected behaviour?

DarthPJB (approved):
I cannot obviously vouch for the density of haskell here - so I've viewed it, read it, become confused, and assumed it's sane and workable.

This review relates only to Nix, besides one question:

why is the file "[plutarch-test/goldens/list.pmatch.bench.golden](https://github.com/Plutonomicon/plutarch/pull/276/files#diff-b70e9645e601fdb41fb7878149692963e431a0935826f1d37fd912a4b5af8932)" empty? I'd expected something, unless it's just a placeholder?

As far as the nix-code goes, it looks to my eye well ordered, and sane - while working around issues with the Haskell ecosystem in an effectively commented fashion. As such, I approve of it.

That said, I'd defiantly hope that @MatthewCroughan will pass his eye over it before merge; as he is without question more likely to find issues - if they exist.

cabal.project
srid: Using wildcards in cabal.project is not uncommon. In fact the default value already uses one. See https://cabal.readthedocs.io/en/3.6/cabal-project.html#cfg-field-packages

srid:
> why is the file "[plutarch-test/goldens/list.pmatch.bench.golden](https://github.com/Plutonomicon/plutarch/pull/276/files#diff-b70e9645e601fdb41fb7878149692963e431a0935826f1d37fd912a4b5af8932)" empty? I'd expected something, unless it's just a placeholder?

@DarthPJB Good catch! That got committed by accident. Removed now.

L-as:
The code should preferably err on unused golden files. We can do this in the future, but someone should make an issue after we merge this.

flake.nix
```diff
@@ -495,7 +563,7 @@
           #     builtins.attrValues self.project.${system}.hsPkgs
           #   )
           # );
-          hspkgs = builtins.map (x: self.project.${system}.hsPkgs.${x}.components.library) [
+          hspkgs = builtins.map (x: self.haddockProject.${system}.hsPkgs.${x}.components.library) [
```
L-as, line 566: Isn't this change entirely unrelated? What does this do?

L-as: OK, I see, it's for setting `-development` to false. You've put a lot of stuff into a single PR which makes it hard to merge.

plutarch-test/plutarch-test.cabal
```diff
@@ -0,0 +1,133 @@
...
+    , sydtest
+    , sydtest-discover
+    , filepath
+    , generics-sop
+    , plutarch
+    , plutarch-benchmark
+    , plutus-core
+    , plutus-ledger-api
+    , plutus-tx
+    , rank2classes        >=1.4.4
```
L-as, line 89: Do you use this?

plutarch-test/plutarch-test.cabal
```diff
@@ -0,0 +1,133 @@
...
+  main-is:            Main.hs
+  hs-source-dirs:     src
+
+  if impl(ghc >= 9.0)
+    other-modules:
+      Plutarch.MonadicSpec
+      Plutarch.FieldSpec
+  other-modules:
+    Spec
+    Plutarch.ApiSpec
```
L-as, line 114: Why are you using the same prefix as `Plutarch`?

flake.nix
srid: `haddockProject` is now used only for building haddocks (this rename makes that clear); the rest of the things use `flakeMatrix` (enumerates on 'development' flag as well as ghc version). All the Nix changes are all done to improve test coverage (across flag valus and compiler versions). I don't see any more Nix changes to be done in this branch.

I wanted 8.10 and development=true to be enabled in CI first, before basing other work on. Would you rather I open subsequent PRs (for porting existing tests) as separate PRs, instead of pushing to this branch?

L-as:
What do you do to handle the discrepancy in the strings passed to `fails` in GHC 8.10 and GHC 9.2?  To me it seems like a good solution might just be to use a placeholder string in `fails` when a `testing` flag is enabled.

plutarch-test/plutarch-test.cabal
srid: Mostly existing convention; `Foo.hs` vs `FooSpec.hs`. Example:

- https://github.com/NorfairKing/validity/blob/master/genvalidity-time/src/Data/GenValidity/Time.hs
- https://github.com/NorfairKing/validity/blob/master/genvalidity-time/test/Data/GenValidity/TimeSpec.hs

L-as: OK, makes sense.

srid:
> What do you do to handle the discrepancy in the strings passed to fails in GHC 8.10 and GHC 9.2?

This PR doesn't provide any general solution to this discrepancy. It sidesteps it entirely (for now), by rewriting that one example to handle the pmatch case exhaustively using perror, https://github.com/Plutonomicon/plutarch/pull/276/commits/e4245abb3603b34d454ee45a45dc16c10d685dbd#diff-3a97af4dc66f7003646b7980f30c5feefff6069a0f5857fa9bb9c1da173f7a3cL182-R193

> To me it seems like a good solution might just be to use a placeholder string in fails when a testing flag is enabled.

In principle it seems to be a bad idea to change application's behaviour[^1] differently when run in tests from when it is used in other contexts. cc @NorfairKing 

[^1]: Specifically, by changing the string here to some constant if run in tests only (but using existing behaviour when used by the user): https://github.com/Plutonomicon/plutarch/blob/a0cbe99921aad7c5df9239cb0240933e4d9b2eaa/Plutarch/TermCont.hs#L39-L41

L-as:
You're right wrt. that. But with this approach we don't test the monadic code properly.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,205 @@
...
+plutarchDevFlagDescribe m =
+#ifdef Development 
+  describe "dev=true" m
+#else
+  describe "dev=false" m
+#endif
+{- ORMOLU_ENABLE -}
+
+{- Asserts the term evaluates without success -}
+pfails :: forall (a :: PType). ClosedTerm a -> Expectation
```
L-as, line 129: You should remove the `forall`, it's not necessary here.

L-as: Same everywhere else where the types are simple.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,205 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE ImpredicativeTypes #-}
```
L-as, line 2: I thought you wanted to use GHC 8.10 with this.

plutarch-test/src/Plutarch/PIsDataSpec.hs
```diff
@@ -0,0 +1,103 @@
...
+  , PlutusTx.ToData (PLifted p)
+  , PLift p
+  , Eq (PLifted p)
+  , Show (PLifted p)
+  ) =>
+  PLifted p ->
+  IO ()
+pFromDataCompat x = plift (pfromData $ pconstantData @p x) `shouldBe` x
+
+pDataCompat ::
```
L-as, line 93: Naming. `pdataCompat` and such.

plutarch-test/src/Plutarch/MonadicSpec.hs
```diff
@@ -0,0 +1,49 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE QualifiedDo #-}
+
+module Plutarch.MonadicSpec (spec) where
+
+import Test.Syd
+
+import Plutus.V1.Ledger.Api
+
+import Plutarch.Api.V1
```
L-as, line 10: For the future, please try to avoid implicit imports like this. Since this isn't important code, we shouldn't block the PR on this though.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,205 @@
...
+  expectationFailure,
+  getTestDescriptionPath,
+  it,
+  pureGoldenTextFile,
+  shouldBe,
+ )
+
+import Plutarch
+import Plutarch.Benchmark (benchmarkScript')
+import Plutarch.Bool
```
L-as, line 44: Please use explicit imports, `(..)` is also implicit.

plutarch-test/goldens/unit.pmatch.bench.golden
```diff
@@ -0,0 +1 @@
+0 {"exBudgetCPU":29873,"exBudgetMemory":200,"scriptSizeBytes":6}
```
L-as, line 1: No `uplc` results for this?

plutarch-test/goldens/trace.dev=false.ptrace.uplc.eval.golden
```diff
@@ -0,0 +1,2 @@
+one (program 1.0.0 ())
```
L-as, line 1: How do you decide which tests to make versions of for both `dev=true` and `dev=false`?

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,205 @@
...
+
+{- | Like `describe`, but determines description from `Development` CPP flag
+
+  Useful to create two sets of othersise identical group of tests that differ
+  only by `Development` flag. This has the effect of creating two sets of golden
+  tests (with different filepaths) for corresponding flag values.
+
+  Typically meant to be used in conjunction with `ptraces`.
+-}
+plutarchDevFlagDescribe :: forall (outers :: [Type]) inner. TestDefM outers inner () -> TestDefM outers inner ()
```
L-as, line 116: Ah, you use this to choose which to test twice?

plutarch-test/src/Plutarch/BoolSpec.hs
```diff
@@ -0,0 +1,64 @@
+module Plutarch.BoolSpec (spec) where
```
L-as, line 1: This isn't your fault, but GH doesn't syntax-highlight this file for some reason...

plutarch-test/src/Plutarch/BoolSpec.hs
```diff
@@ -0,0 +1,64 @@
...
+        let p1 = pand # pcon PFalse # pdelay perror
+            p2 = pcon PFalse #&& perror
+        goldens All [("pand", popaque p1), ("op", popaque p2)]
+        it "pand" $ passert $ pnot # pforce p1
+        it "op" $ passert $ pnot # p2
+        it "pand.perror" $ do
+          pfails $ pand # pcon PFalse # perror
+          pfails $ pand # pcon PTrue # perror
+          pfails $ pcon PTrue #&& perror
+    describe "por" $ do
```
L-as, line 42: While not a blocking issue, I do wonder how we could make the testing API more ergonomic.

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -0,0 +1,205 @@
...
+golden' :: forall a. PlutarchGolden -> Maybe String -> ClosedTerm a -> Spec
+golden' pg mk p =
+  goldens' pg mk [("0", popaque p)]
+
+{- | Like `golden` but for multiple programs
+
+  Multiple programs use a single golden file. Each output separated from the
+  keyword with a space.
+-}
+goldens :: PlutarchGolden -> [(String, ClosedTerm a)] -> Spec
```
L-as, line 175: Perhaps this should be `ClosedTerm POpaque` for the sake of consistency and type inference? Your choice.

L-as (changes requested):
Looks mostly great, thanks!

flake.nix
```diff
@@ -516,29 +584,67 @@
...
+      # 
+      # In effect, this allows us to run an 'app' as part of the build process (eg: in CI).
+      flakeApp2Derivation = system: appName:
+        (nixpkgsFor system).runCommand appName { } "${self.apps.${system}.${appName}.program} | tee $out";
     in
-    {
+    rec {
       inherit extraSources cabalProjectLocal haskellModule tools;
 
-      project = perSystem projectFor;
```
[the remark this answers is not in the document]
L-as: A good example is `hsPkgs`.

plutarch-test/src/Plutarch/MonadicSpec.hs
L-as: Note that I do know that the tests originally had this problem due to laziness on my part.

plutarch-test/src/Plutarch/Test.hs
srid: The `a` requires ImpredicativeTypes. Keeping it polymorphic is nice, because it avoids popaque in most golden groups, eg:

image

plutarch-test/src/Plutarch/Test.hs
srid: It is needed for the `a` in `goldens :: PlutarchGolden -> [(String, ClosedTerm a)] `

plutarch-test/src/Plutarch/Test.hs
srid: Yup, the idea is to use `plutarchDevFlagDescribe` somewhere in test hierarchy where you know that the child sub-tree of tests also use Plutarch programs that will behave differently across 'development' flag sets (eg: with tracing). I'll add this to the README.

plutarch-test/src/Plutarch/BoolSpec.hs
srid: I'll think about this outside this PR. Maybe some sort of MonadWriter that collects programs on the goal, and batches golden tests at the end ...

plutarch-test/src/Plutarch/BoolSpec.hs
srid: 🤷🏿 
image

flake.nix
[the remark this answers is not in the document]
srid: I added `projectMatrix`. I suppose this would be useful from `nix repl`.

plutarch-test/goldens/trace.dev=false.ptrace.uplc.eval.golden
srid: Answered elsewhere, and explained in README.

plutarch-test/goldens/unit.pmatch.bench.golden
srid: @L-as Some tests specifically generate _only_ bench goldens (_or_ only uplc goldens), rather than both, as controlled by this type:

https://github.com/Plutonomicon/plutarch/blob/1ebeff5b2dcf5205c753a55b6efe430b2c7f5b00/plutarch-test/src/Plutarch/Test.hs#L135-L149

That's what is happening here, because of `golden Bench p` (instead of `golden All p`). Should we get rid of `Bench` and *always* generate printTerm regardless of the program? Likewise, are there cases where avoiding benchmarks make sense? It is used in a handful of cases, like:

```
plutarch-test/src/Plutarch/PLamSpec.hs
12:      golden PrintTerm $ plam (\x -> x)
14:      golden PrintTerm $ plam (\_ y -> y)
16:      golden PrintTerm $ plam (\x _ -> plet x $ \_ -> perror)
```

I figured we don't need to benchmark lambdas, inasmuch as the body of a lambda won't be evaluated anyway. But may be it is worth benchmarking the script size anyway.

plutarch-test/src/Plutarch/MonadicSpec.hs
srid: Sure. I assume we can relax on `Plutus.V1.Ledger.Api` and the like for testing code. Added explicit imports for Plutarch.*.

plutarch-test/goldens/api.example.getFields.uplc.golden
```diff
@@ -0,0 +1 @@
+0 (program 1.0.0 (\i0 -> force (force sndPair) (unConstrData i1)))
```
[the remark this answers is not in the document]
srid: Added in 48f439aea4e26213c834bf2eba0ce969f732e35f - but see comment below for larger picture.

plutarch-test/goldens/unit.pmatch.bench.golden
srid: Here's an example of something generating bench goldens, but no printTerm golden:

```haskell
    describe "phexByteStr" $ do
      let a :: [String] = ["42", "ab", "df", "c9"]
          p = pconstant @PByteString (BS.pack $ map readByte a) #== phexByteStr (concat a)
      golden Bench p
      it "relation" $ passert p
```

At least the eval uplc would be useless here, as it would be: `(program 1.0.0 True)`. The non-eval UPLC would be: `(program 1.0.0 ((\i0 -> equalsByteString i1 i1) #42abdfc9))` - however why should it be under "phexByteStr" tree (see above), rather a tree for equalsByteString?

srid: Here, we can do benchmark golden, because doing that will result in error:

```haskell
    describe "pmatch" $ do
      let p = pmatch (integerList [1, 3, 1]) $ \_ -> perror
      golden PrintTerm p
```

Changing `PrintTerm` -> `All` will result in:

image

srid: Another example, where we can't really benchmark a scriptcontext constant:

```haskell
golden PrintTerm ctx
```
image

srid: Anyway, uplc goldens added in [a001bde](https://github.com/Plutonomicon/plutarch/pull/276/commits/a001bdedb021b4314f3bebad570538ea39cfe4c0)

srid:
@L-as This is ready for merge now. Opened #289 and #290 as follow-up (in addition to porting the rest).

plutarch-test/src/Plutarch/Test.hs
L-as: That is true and is quite unfortunate. You can have a newtype wrapper for `ClosedTerm` if you want, i.e.
```haskell
newtype MyClosedTerm a = MyClosedTerm (ClosedTerm a)
```
which should fix it AFAICT, and then you could do `MyClosedTerm $ ...` in the list you pass to `goldens`. You could also hard-code it to be `POpaque`. It's quite unfortunate that we still support GHC 8.10, but we will probably drop it once GHC 9.2.2 comes out, since I've realised that there isn't much point to the older GHC 8.10 support.

plutarch-test/goldens/unit.pmatch.bench.golden
L-as: > At least the post-eval uplc would be useless here, as it would be: (program 1.0.0 True). The pre-eval UPLC would be: (program 1.0.0 ((\i0 -> equalsByteString i1 i1) #42abdfc9)) - however why should it be under "phexByteStr" tree (see above), rather a tree for equalsByteString?

I don't agree with this. the post-eval UPLC is still very important, because if it starts evaluating into something else, we have a bug! 9/10 times changes to the golden files are accidental and means we have a bug.

I don't see why we wouldn't support only `All`. Terms that err can still be benchmarked. It just means that the benchmarking result should include only the script size, and possibly still CPU and memory cost if that information isn't lost.

Wrt. the `ctx` case, I'm not sure why that's erring in the first place, but I think that we should still benchmark this too.

You could potentially include the error message *in* the benchmarking golden file.

plutarch-test/src/Plutarch/Test.hs
L-as: Good point.

L-as:
@srid 

I really really think we should use CPP to check for a testing flag in calls to `fails` and just substitute it with a placeholder string in those cases. As you've noted in #290 , even just moving lines around will cause the golden files to change. If a PR isn't buggy, then most often **no golden files should change**. This violates that principle and means that we'll have many unrelated changes to the golden files that don't indicate a bug.

L-as (changes requested):
See previous comment wrt. #290 

Removing everything but `All` would also be nice if you have time.

L-as:
Thanks for the work BTW, this is really great and pretty close to as good as it could get AFAICT. For the future, I think we should try to see if we can group the tests more so we don't have ~200 golden files, albeit that should be another PR.

PR #278, opened 2022-02-13
Fix development mode tests

TotallyNotChase:
Thanks to @MangoIV for flagging the failing tests!

It *seems* like the compilation outputs changed in response to a `plet` optimization.


L-as (approved):
This is actually the inlining optimisation. `λx. w ((λy. z) x)` is translated into `λx. w z`.

PR #280, opened 2022-02-14
Made some suggestions for Chase

jhodgdev:
Hiya @TotallyNotChase I have started work through reviewing your docs. Haven't finished yet but here is what I have done so far. Will tag you again when I am done. Thanks again for inviting my feedback.

### DIFF
--- a/docs/SYNTAX.md
+++ b/docs/SYNTAX.md
@@ -5,24 +5,30 @@ This document describes the syntax of Plutarch through code examples.
 <details>
 <summary> Table of Contents </summary>
 
-- [Constants](#constants)
-  - [Static building with `pconstant` and `pconstantData`](#static-building-with-pconstant-and-pconstantdata)
-  - [Dynamic building with `pcon`](#dynamic-building-with-pcon)
-  - [Overloaded literals](#overloaded-literals)
-  - [Miscellaneous](#miscellaneous)
-- [Lambdas](#lambdas)
-- [Delayed terms and Forcing](#delayed-terms-and-forcing)
+-   [Constants](#constants)
+    -   [Static building with `pconstant` and `pconstantData`](#static-building-with-pconstant-and-pconstantdata)
+    -   [Dynamic building with `pcon`](#dynamic-building-with-pcon)
+    -   [Overloaded literals](#overloaded-literals)
+    -   [Miscellaneous](#miscellaneous)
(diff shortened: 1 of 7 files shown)


jhodgdev:
Please note that through force-of-habit, I probably let my Markdown formatter change some syntax things that were not important. I have not made any (deliberate) changes to your work but have rather left suggestions as annotated comments.

TotallyNotChase:
I fixed the merge conflicts for you :)

docs/SYNTAX.md
```diff
@@ -98,17 +115,24 @@ x = phexByteStr "41"
 ```
 
 # Lambdas
+
 You can create Plutarch level lambdas by apply `plam` over a Haskell level lambda/function.
```
peter-mlabs, line 119: I think it may be useful here to highlight that `plam` works "over a Haskell level lamba/function _of Plutarch terms_".  I'm re-reading this for the first time, and stumbled again thinking that I could just yeet `plam` in front of any Haskell function, for a second.

docs/USAGE.md
```diff
@@ -84,7 +106,7 @@ It works as you would expect, though the type is scary. Think of it as the Haske
 fix :: ((a -> b) -> (a -> b)) -> a -> b
```
peter-mlabs, line 106: I _believe_ both of these signatures can be written as

```haskell
pfix :: Term s (((a :--> b) :--> (a :--> b)) :--> (a :--> b))
```

and

```haskell
fix :: ((a -> b) -> (a -> b)) -> (a -> b)
```

which, to me, makes it clear that there isn't some hidden difference between the signatures and highlights that they're working on functions in particular. This is also how `pfix` is listed in GHCi.

peter-mlabs:
Also, I think the "implementing `PlutusType`" section should come before "PConstant"/"PLift", because it seems that the method you use to derive the latter is determined by the former. Is that correct?

jhodgdev:
Hiya @TotallyNotChase I will make use the browser to continue my review as/when I get the chance :)

TotallyNotChase:
> Hiya @TotallyNotChase I will make use the browser to continue my review as/when I get the chance :)

Sure. Let me know when you think this PR is ready to merge. I'll update the docs with your suggestions post merge and we can hopefully get this overhaul finished by the end of the week.

TotallyNotChase:
Thanks!

PR #281, opened 2022-02-14
Benchmarking size of un-applied Scripts

### LINKED ISSUE #269: Measure sizing of non-applied scripts in Benchmark

t1lde:
Added functions to `plutarch-benchmark` which measure the size of an 'unapplied' version of a script, and some helpers for creating benchmarks with arguments applied in various ways.

I haven't incorporated it into the plutarch-test golden tests API, as I'm not sure it's necessary.

Closes #269 , Merge after #276 #267


t1lde:
cc: @srid on whether it's worth incorporating into the plutarch-test golden benchmarking

L-as:
Wait, had we been doing this wrong all along? This is problematic.

plutarch-benchmark/src/Plutarch/Benchmark.hs
```diff
@@ -49,29 +59,120 @@ import Data.Maybe (fromJust)
...
+  benchmarkScriptUnapplied name unApplied $
+    Plutus.applyArguments
+      unApplied
+      [ Plutus.toData datum
+      , Plutus.toData redeemer
+      , Plutus.toData ctx
+      ]
+
+-- | Benchmark a MintingPolicy with the provided arguments
+benchmarkMintingPolicy ::
```
L-as, line 121: I'm honestly not in favor of adding functions specific to each "type" of script. Rather than using `Plutus.Validator`, etc., these should just take the term, and the user should just specify what the type of the term is, e.g. `PValidator`.

nixinator (commented):
too big for me to review!

plutarch-benchmark/src/Plutarch/Benchmark.hs
t1lde: removed them, benchUnapplied should suffice I think.

MatthewCroughan (dismissed):
As long as it builds and passes, I'm happy. I can't say much about the Nix additions, because they require a lot of context that I don't have for the Plutonomicon anymore, as it has advanced so quickly.

t1lde:
Should be ready to merge after #267

t1lde:
Accidentally puashed the fixes to #267 to this branch instead

L-as:
I have to be honest, I'm not sure what this PR is trying to do.

t1lde:
> I have to be honest, I'm not sure what this PR is trying to do.

See #269 benchmarking the size of scripts without arguments applied, since they would otherwise throw the measurement off.

Issue #191, opened 2022-01-26
Add CI for development and non-development build


L-as:

TotallyNotChase:
`cabal test -f development`

Should ideally also accompany the build for GHC 8.10.7 in CI

srid:
I just ran some tests with development flag set, and the UPLC generated is different as one would expect:

<img width="1707" alt="image" src="https://user-images.githubusercontent.com/3998/153934801-7d7abfb7-080c-4c86-b786-e48d75c15b08.png">

Pattern match failure in line 171 translates to a perror automatically:

https://github.com/Plutonomicon/plutarch/blob/c71322f84d2594dd49fb433e71837acd1f3c0512/plutarch-test/src/Plutarch/ApiSpec.hs#L170-L172

This means we'll need two sets of goldens for the values of the flag. This is obvious for tracing tests (they always use ptrace*), but not so obvious for other tests (which, right now, seems to be those using monadic syntax).

srid:
Yes.

PR #283, opened 2022-02-14
Port examples using ptrace* to plutarch-test

srid:
(Incidentally, this PR puts us one step ahead for doing #191)


cabal.project
```diff
@@ -3,3 +3,7 @@ packages:
   ./plutarch-*
 
 benchmarks: true
+
+-- Enabling the flag here allows TraceSpec.hs and such tests to actually test
+-- the `ptrace` family of functions.
+flags: -development
```
srid, line 9: This doesn't change existing behaviour. `development` remains turned off by default.

When doing #191 however we should figure out a way to toggle this at project level in flake.nix checks.

srid:
The goldens for scenario where `development` flag is set will be addressed as part of #191.

CI passes locally, so merging.

Issue #198, opened 2022-01-26
Make plutarch.cabal cleaner and add tests for GHC 8.10


L-as:
https://github.com/Plutonomicon/plutarch/issues/145#issuecomment-1022629754

srid:
Dup in part: #145

TotallyNotChase:
Shouldn't this be closed already?

L-as:
We can close this when https://github.com/Plutonomicon/plutarch/blob/staging/plutarch.cabal#L153 is removed entirely.

srid:
> We can close this when https://github.com/Plutonomicon/plutarch/blob/staging/plutarch.cabal#L153 is removed entirely.

Should be done in #302

PR #284, opened 2022-02-14
Run tests in development mode as well

### LINKED ISSUE #191: Add CI for development and non-development build

srid:
Run `plutarch-test` with development flag set to false (current behaviour) _as well as_ with it set to true.

Resolves #191

Still GHC 9 only, but #198 will be the next step from here - so flake.nix has been refactored somewhat in preparation for that.


flake.nix
```diff
@@ -481,6 +481,8 @@
               packages.plutarch-test.components.exes.plutarch-test.build-tools = [
                 pkgSet.hsPkgs.sydtest-discover
               ];
+              packages.plutarch-test.flags.development = flagDeveloment;
+              packages.plutarch.flags.development = flagDeveloment;
```
L-as, line 485: Develoment?

nixinator (commented):
flaketastic

Issue #145, opened 2022-01-19
CI for GHC 8.10


srid:
Once staging is merged, will Plutarch continue to support GHC 8.10?

I think it should, because not all smart contracts will be moving over to GHC 9.2.

L-as:
I think yes, and AFAICT the core library should still compile fine, even if the tests don't.

TotallyNotChase:
So, when using the `flake.nix` from [this PR](https://github.com/mlabs-haskell/plutus-scaffold/pull/17), and changing `ghcVersion` to `ghc8107` - I seem to be getting compile errors for fourmolu during `nix develop`-
```
src/Ormolu/Utils/Extensions.hs:24:1: error:
    Could not find module ‘Distribution.Utils.Path’
    Perhaps you meant
      Distribution.Utils.IOData (from Cabal-3.2.1.0)
      Distribution.Utils.MD5 (from Cabal-3.2.1.0)
      Distribution.Utils.Base62
```
is it expecting a more up to date cabal version?

TotallyNotChase:
FWIW I did get that flake and everything to work with 8.10.7 by removing fourmolu.

srid:
@TotallyNotChase Why not modify Plutarch's `flake.nix`? Did you notice it requiring more than compiler version change?

TotallyNotChase:
¯\_(ツ)_/¯ I'm just not very good with nix. It works flawlessly on my testbed with ghc8107 after removing fourmolu. I'm sure you could make it work with fourmolu as well by modifying the nix a bit.

srid:
For the record, it only requires changing compiler version, removing fourmolu/ format checks and adding back HLS.

Diff: https://github.com/Plutonomicon/plutarch/compare/staging...srid/ghc8107

Everything but the examples build:

https://hercules-ci.com/github/Plutonomicon/plutarch/jobs/353

I wonder if all of this can be parameterized so we can do `nix develop` using either of the compilers.

L-as:
I think we should at the very least in CI make sure compilation with GHC 8.10.7 works.

L-as:
I changed the name of the issue.

srid:
#188 adds CI for the library, but we are not running tests.

L-as:
The tests can't be run. They depend on GHC 9.2.1.

blamario:
[...]

PR #287, opened 2022-02-15
Run tests in GHC 8.10

srid:
Tests for #198


srid:
To understand the reason behind golden change, see https://github.com/Plutonomicon/plutarch/issues/145#issuecomment-1040577889

srid:
Tests are now run in 4 different configurations. Look for the output in the "Log" tab,

image

flake.nix
```diff
@@ -467,9 +467,12 @@
...
             # Remove stanzas from .cabal that won't work in GHC 8.10
             sed -i '/-- Everything below this line is deleted for GHC 8.10/,$d' $out/plutarch.cabal
-            # Remove packages that won't work in GHC 8.10 (yet)
+
+            # Prevent `sydtest-discover` from using GHC9 only modules when building with GHC810
+            # https://github.com/NorfairKing/sydtest/blob/master/sydtest-discover/src/Test/Syd/Discover.hs
             chmod -R u+w $out/plutarch-test
-            rm -rf $out/plutarch-test
+            rm -f $out/plutarch-test/src/Plutarch/MonadicSpec.hs
+            rm -f $out/plutarch-test/src/Plutarch/FieldSpec.hs
```
srid, line 475: @Norfairking FYI - this ^ is done to prevent sydtest-discover from using ghc9-only modules in ghc8. It would be great if it used the .cabal file to filter the modules.

NorfairKing: I don't think making `sydtest-discover` more complex like that will be a good tradeoff.
I think this is an okay hack for now, or we could use CPP

PR #292, opened 2022-02-16
Resolve #291

### LINKED ISSUE #291: Always benchmark and print unevaluated and evaluated forms of all tests
#276

https://github.com/Plutonomicon/plutarch/pull/276#discussion_r807748248

We always want to benchmark and print all terms. If it errs, we note that in the benchmark result along with the size.

L-as:


srid:
@L-as One of the failures is because of not removing `src/` in https://github.com/Plutonomicon/plutarch/blob/c77fcd605269bd8183d5496e297eb38503ea0e29/flake.nix#L471-L475

L-as:
Finally updated plutus! That was painful.

L-as:
I will fix this later because of changes in #307

Issue #262, opened 2022-02-09
Use `TermCont` more


L-as:
This might conflict with #202 , unsure at this point.

TotallyNotChase:
What do you mean? Use `TermCont` more where?

L-as:
Anything that returns `(a -> Term s b) -> Term s b` should just use `TermCont s a`.

blamario:
Rather than change the type of `pmatch` and other functions, it would be better to export their synonyms from `Plutarch.TermCont`.

Geometer1729:
I find pmatch is still sometimes useful even using TermCont fairly heavily. I'd preffer a solution where both the original pmatch and something that uses TermCont are available. Currently we're using `tmatch = tcont. pmatch`, and that's been working pretty well.

L-as:
Bad idea for Plutarch 1

PR #294, opened 2022-02-16
Guide overhaul by splitting into several documents + more

jhodgdev:
I am finishing off reviewing Chase's guide for ease of discussion/use.


docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+# Prefer Plutarch level functions
+
+Plutarch level functions have a lot of advantages - they can be hoisted; they are strict so you can [use their arguments however many times you like without duplicating work](#dont-duplicate-work); they are required for Plutarch level higher order functions etc. Unless you _really_ need laziness, like `pif` does, try to use Plutarch level functions.
+
+Also see: [Hoisting](./CONCEPTS.md#hoisting-metaprogramming--and-fundamentals).
+
+# When to use Haskell level functions?
+Although you should generally [prefer Plutarch level functions](#prefer-plutarch-level-functions), there are times when a Haskell level function is actually much better. However, figuring out *when* that is the case - is a delicate art.
+
+There is one simple and straightforward usecase though, when you want a function argument to be lazily evaluated. In such a case, you should use a Haskell level functions that `pdelay`s the argument before calling some Plutarch level function. Recall that [Plutarch level functions are strict](#plutarch-functions-are-strict).
```
jhodgdev, line 85: use case

jhodgdev, line 85: function

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+  PNil -> match_nil
+```
+It takes in a Plutarch level function, let's see a typical usage-
+```hs
+pelimList
+  (plam $ \x xs -> pcons # x # (self # xs))
+  pnil
+  ls
+```
+This is rather redundant, the above snippet will exhibit inlining to produce-
```
jhodgdev, line 103: Replace 'exhibit'.

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+  pnil
+  ls
+```
+would turn into-
+```hs
+pmatch ls $ \case
+  PCons x xs -> pcons # x # (self # xs)
+  PNil -> match_nil
+```
+It turns out that `pelimList` usages *almost always* use a one-off Haskell level function with a redundant `plam` - so `pelimList` would benefit greatly from just taking a Haskell level function directly.
```
jhodgdev, line 129: Consider:

It turns out that using `pelimList` *almost always* involves using a one-off Haskell level function (and therefore a redundant `plam`). In such cases, `pelimList` benefits greatly from just taking a Haskell level function directly.

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+
+However, **not all higher order functions** benefit from taking Haskell level functions. In many HOF usages, you could benefit from passing a commonly used function argument, rather than a one-off function argument. Imagine `map`, you don't always map with one-off functions - often, you `map` with existing, commonly used functions. In these cases, that commonly used function ought to be a Plutarch level function, so it can be hoisted and `map` can simply reference it.
```
jhodgdev, line 131: HOF?

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+
+# List iteration is strict
+Chained list operations (e.g a filter followed by a map) are not very efficient in Plutus Core. In fact, the iteration is not lazy at all! For example, if you did a `pfilter`, followed by a `pmap`, on a builtin list - the entire `pmap` operation would be computed first, the whole list would be iterated through, and *only then* the `pfilter` would start computing. Ridiculous!
+
+# Let Haskell level functions take responsibility of evaluation
+We've discussed how a Haskell level function that operates on Plutarch level terms needs to [be careful](#dont-duplicate-work) about [work duplication](./CONCEPTS.md#plet-to-avoid-work-duplication). Related to this point, it's good practice to design your Haskell level functions so that *it takes responsibility* for evaluation.
+
+The user of your Haskell level function doesn't know how many times it uses the argument it has been passed! If it uses the argument multiple times without `plet`ing it - there's duplicate work! There's 2 solutions to this-
```
jhodgdev, line 177: two solutions

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+
+Instead, try to offload the responsbility of evaluation to the Haskell level function - so that it only `plet`s when it needs to.
+
+Of course, this is not applicable for recursive Haskell level functions!
+
+# The isomorphism between `makeIsDataIndexed`, Haskell ADTs, and `PIsDataRepr`
+When [implementing `PIsDataRepr`](#implementing-pisdatarepr-and-friends) for a Plutarch type, if the Plutarch type also has a Haskell synonym (e.g `ScriptContext` is the haskell synonym to `PScriptContext`) that uses [`makeIsDataIndexed`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx.html#v:makeIsDataIndexed) - you must make sure the constructor ordering is correct.
+
+> Aside: What's a "Haskell synonym"? It's simply the Haskell type that *is supposed to* correspond to a Plutarch type. There doesn't *necessarily* have to be some sort of concrete connection (though there can be, using [`PLift`/`PConstant`](./TYPECLASSES.md#pconstant--plift)) - it's merely a connection you can establish mentally.
```
jhodgdev, line 190: I like this aside. I would consider replicating it somewhere more prominent.

TotallyNotChase:
(you should retarget this to `guide`)

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+  , txInfoValidRange  :: POSIXTimeRange
+  , txInfoSignatories :: [PubKeyHash]
+  , txInfoData        :: [(DatumHash, Datum)]
+  , txInfoId          :: TxId
+  }
+```
+The *field names* don't matter though. They are merely labels that don't exist in runtime.
+
+# Prefer statically building constants whenever possible
+Whenever you can build a Plutarch constant out of a pure Haskell value - do it! Functions such as `pconstant`, `phexByteStr` operate on regular Haskell synonyms of Plutarch types. Unlike `pcon`, which potentially works on Plutarch terms (ex: `pcon $ PJust x`, `x` is a `Term s a`). A Plutarch term is an entirely "runtime" concept. "Runtime" as in "Plutus Core Runtime". They only get evaluated during runtime!
```
jhodgdev, line 252: replace ex with e.g.

docs/TRICKS.md
```diff
@@ -0,0 +1,290 @@
...
+ in pcon $ PMinting $ pdcons # pdata currSym # pdnil
+```
+The semantics are both are the same. But the former (`pconstant`) compiles to a constant term directly. Whereas the latter compiles to some code that *builds* the constant during Plutus Core runtime.
+> Aside: Remember that Haskell runtime is actually compile-time for Plutarch! Even if you have a dynamically computed variable in the Haskell world, it's still a *constant* in the Plutarch world. So you can use it just as well as an argument to `pconstant`!
+
+Whenever you need to build a Plutarch term of type `a`, from a Haskell value, use `pconstant`. Whenever you need to build a Plutarch term of type `PAsData a`, use `pconstantData`!
+
+# Figuring out the representation of a Plutarch type
+As discussed in other sections of guide, Plutarch types are merely tags to underlying semantic representations. This is an eDSL after all! Their data declarations *actually* don't matter as far as internal semantics are concerned. It's actually the `PlutusType` instance that *really* determines the representation. So how do you figure out the representation of this seemingly transient tag? By following *conventions*.
```
jhodgdev, line 275: of the guide

TotallyNotChase:
This PR is ready to be merged.

docs/examples/VALIDATOR.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+# Validator that always fails
+
+```hs
+import Plutarch.Prelude
+import Plutarch.Api.V1.Contexts
+import Plutarch.Api.V1.Scripts
+
+alwaysFails :: Term s (PAsData PDatum :--> PAsData PRedeemer :--> PAsData PScriptContext :--> PUnit)
```
L-as, line 41: It should be `PData :--> PData :--> PAsData PScriptContext :--> POpaque`

docs/examples/VALIDATOR.md
```diff
@@ -0,0 +1,126 @@
...
+import Plutarch.Prelude
+import Plutarch.Api.V1.Contexts
+import Plutarch.Api.V1.Crypto
+import Plutarch.Api.V1.Scripts
+
+pmatchC :: PlutusType a => Term s a -> TermCont s (a s)
+pmatchC = tcont . pmatch
+
+checkSignatory :: Term s (PPubKeyHash :--> PAsData PDatum :--> PAsData PRedeemer :--> PAsData PScriptContext :--> PUnit)
+checkSignatory = plam $ \ph _ _ ctx' -> unTermCont $ do
```
L-as, line 71: Why not use `P.do`?

docs/examples/VALIDATOR.md
```diff
@@ -0,0 +1,126 @@
...
+    )
+    (Spending (TxOutRef "" 1))
+
+> evalWithArgsT (checkSignatory # pubKeyHash) [PlutusTx.toData (), PlutusTx.toData (), PlutusTx.toData mockCtx]
+Right (Program () (Version () 1 0 0) (Constant () (Some (ValueOf unit ()))))
+```
+
+# Using custom datum/redeemer in your Validator
+
+All you have to do is [implement `PIsDataRepr` and friends](./../TYPECLASSES.md#implementing-pisdatarepr-and-friends) for your custom datum/redeemer and you can use it just like `PScriptContext` in your validators!
```
L-as, line 126: This isn't true FWIW.

L-as (commented):
I'm unsure of what has changed because of the lack of organisation in the commits, so I'm not sure whether what I've commented on is new or old.

docs/examples/VALIDATOR.md
TotallyNotChase: When are the `PDatum` and `PRedeemer` types supposed to be used? Also why use `POpaque` in a concrete validator function? I think that should really be necessary when you wrap the validator function to get a `PValidator`.

docs/examples/VALIDATOR.md
TotallyNotChase: What should be changed here?

srid:
Some thoughts:

This is a good start, but if we split this further (recursively) it would make linking more straightforward. For eg., 

- docs/TYPECLASSES/PlutusType.md
- docs/TYPES/PList.md

I also think it would make future PRs easier to review, since we don't have to scroll a long document to locate the diff deep inside it ... or multiple (disparate) thunks spread across the same .md file.

2-3 levels of directory hierarchy would be reasonable I think? And we can generate the website like: https://input-output-hk.github.io/adrestia/ or https://input-output-hk.github.io/cardano-wallet/ - with sidebar for free (from directory hierarchy).  You can preview this on our docs, using:

```sh-session
$ cachix use srid # optional
$ nix run github:srid/emanote -- run --port 8081 -L ./docs/
```

At the moment the sidebar looks flat:

image

L-as:
I have to be honest, I'm not in favor of preferring `TermCont` in the guide. It's a suboptimal solution because of the need for all the extra syntax (or extra definitions), and GHC 8.10 support isn't really very relevant anymore. It is only important for legacy projects that haven't bothered to switch to GHC 9.2 because of the associated work necessary for decoupling the off-chain and on-chain code bases (which should be done anyway!).

L-as:
In the future we *will* have to drop GHC 8.10 support anyway.

TotallyNotChase:
Marking this as draft for the time being as I want to merge #310 first. I'm also going to split this further afterwards as @srid suggests.

TotallyNotChase:
@L-as I really do think `TermCont` is more expressive, flexible, and familiar - if solely due to it being a real monad. Maybe `QualifiedDo` is slightly more terse but I personally think `TermCont` is easier to grasp and extend.

I was especially delighted by some of the examples @srid showed me-
```hs
-- | Terminate with given value on empty list, otherwise continue.
nonEmpty :: Term s r -> PList a s -> TermCont @r s (Term s a, Term s (PList a))
nonEmpty x0 list = TermCont $ \k ->
  case list of
    PSCons x xs -> k (x, xs)
    PSNil -> x0
```
Example usage within do syntax-
```hs
(x, xs) <- nonEmpty r =<< tcon (pmatch l)
```

I'm sure there are many more really cool things you could do with `TermCont`, while it being also easier to grasp due to the guarantees of a monad.


I *really* believe that `TermCont` should be the way going forward, but I'm going to modify the guide according to what you and others actually want. Perhaps we should ask users what they prefer?

(for users and others to judge) In summary:
## `QualifiedDo` with `Plutarch.Monadic`
* It has more convenient syntax (discussed [in the guide](https://github.com/Plutonomicon/plutarch/blob/master/docs/GUIDE.md#do-syntax-with-qualifieddo-and-plutarchmonadic)) in the simple and common cases.

  Whereas you write `x <- tcont $ pmatch ...` for `TermCont`, you merely write `x <- pmatch` for this. (subject to change due to #262).

  Whereas you have to `unTermCont` to obtain the `Term` out of a `TermCont` computation, you don't have do any such thing here.

  Whereas you have to use `pure` on regular terms to create and yield a `TermCont` within monadic bindings - you don't have do any such thing here. Just `pconstant ...` is enough. No need for `pure $ pconstant ...`, for example.
* It is not a *monad* at all. It just leverages do syntax. Extending it with other monadic combinators may not be as familiar/ergonomic.
* It requires GHC 9. This will be the future for Plutarch anyway (that is, Plutarch 2.0).

Example snippet:
```hs
{-# LANGUAGE QualifiedDo #-}

import Plutarch.Api.V1.Contexts
import qualified Plutarch.Monadic as P
import Plutarch.Prelude

f :: Term s (PScriptPurpose :--> PUnit)
f = plam $ \x -> P.do
  PSpending _ <- pmatch x
  ptrace "matched spending script purpose"
  pconstant ()
```

## `TermCont`
* This is a *real monad* that follows the expected laws. It """composes""" the way you would expect other monads and is extensible in the same way. See `nonEmpty` example above.
  
  [Relevant section in the guide](https://github.com/Plutonomicon/plutarch/blob/guide/docs/USAGE.md#do-syntax-with-termcont) (note: this one is on the `guide` branch).
* Arguably less convenient since you need to write the `unTermCont`, `pure` and similar things you might be accustomed to when using other monads.
* Does not require GHC 9, whether or not that's relevant.

Example snippet:
```hs
import Plutarch.Api.V1.Contexts
import Plutarch.Prelude

pmatchC :: PlutusType a => Term s a -> TermCont s (a s)
pmatchC = tcont . pmatch

ptraceC :: Term s PString -> TermCont s ()
ptraceC s = tcont $ \f -> ptrace s (f ())

test :: Term s (PScriptPurpose :--> PUnit)
test = plam $ \x -> unTermCont $ do
  PSpending _ <- pmatchC x
  ptraceC "matched spending script purpose"
  pure $ pconstant ()
```
> Notice how you need to define `pmatchC` and `ptraceC` currently. Subject to change with #262 

Feel free to extend this with further points as you please.

L-as:
Having to define your own functions for pmatch, etc. is simply not acceptable IMO. I agree that TermCont has the potential to be a more useful abstraction, but see #262.

It is possible that we can have something that is almost strictly more ergonomic, but if we are to use it in the guide, then we should fix #262 first. There is also the issue of linearity that might possibly mess with this.

TotallyNotChase:
> Having to define your own functions for pmatch, etc. is simply not acceptable IMO.

That makes complete sense; I had failed to realize that. But perhaps we can overcome this by simply using `tcont` inline? Like we have in our examples/tests? I can also use regular `ptrace` instead of `ptraceC` and perhaps have a note somewhere about `ptraceC` definition - indicating how users can build useful `TermCont` utilities.

L-as:
You could explain the concept using TermCont, then show the QualifiedDo syntax

PR #296, opened 2022-02-16
Make output directory of `benchMain` CSV file configurable

emiflake:
Summary:
- Make it possible for consumers of `plutarch-benchmark` to pass `--output` flag that determines where CSV file will go. By default it will still go to the local `bench.csv`
- Use `optparse-applicative` instead of `getArgs`

Should we make outputting CSV files not the default behavior?

### DIFF
--- a/plutarch-benchmark/benchmark-diff/Main.hs
+++ b/plutarch-benchmark/benchmark-diff/Main.hs
@@ -3,21 +3,25 @@ module Main (main) where
 import Plutarch.Benchmark (decodeBenchmarks, diffBenchmarks, renderDiffTable)
 
 import qualified Data.ByteString.Lazy as BSL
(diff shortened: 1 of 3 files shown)


srid:
Yes, the benchmark executables (not the library) will eventually go away once all the examples/benchmarks are ported over to plutarch-test. The latter uses golden tests, which are superior. Smart contract projects can use `Plutarch.Test` to do testing in similar manner (@Geometer1729 and I do this in our project).

emiflake:
Oh, right. Then this PR is useless, since I don't think anyone requires this functionality *right now* and it will quickly be deprecated.

In what issue is the inclusion of benchmarks in the goldens tracked?

srid:
> In what issue is the inclusion of benchmarks in the goldens tracked?

#302

Issue #290, opened 2022-02-16
Remove flakiness in `fail` pattern match messages


srid:
This can happen as a result of just shifting the program by a line or two:

<img width="590" alt="image" src="https://user-images.githubusercontent.com/3998/154170456-bf50b215-c985-478c-8574-a86996328825.png">

https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fvx3q5p6blavkkqaab5a4r3jvlharfmw8-test-ghc9-dev.drv/log?via-job=02f9845c-7185-48b7-800d-0dac66e9dad1

L-as:
https://github.com/Plutonomicon/plutarch/pull/276#issuecomment-1041320034

We really need to replace the string in `fails` in tests.

PR #297, opened 2022-02-16
test: make `compile` deterministic

srid:
Add a `compileD :: ClosedTerm a -> Scripts.Script` and use it throughout golden tests. All it does is replace `Trace "Pattern match failure<flaky part>"` with `Trace "Pattern match failure..."` making the compiled script deterministic, and thus resolving #290.

Merge after #293


srid:
image

plutarch-test/src/Plutarch/Test/Deterministic.hs
```diff
@@ -0,0 +1,68 @@
...
+  walkScript $ \term -> do
+    -- Replace the 's' in `trace s`.
+    Apply () b@(Force _ (Builtin _ Trace)) (Constant () (Some (ValueOf DefaultUniString s))) <- pure term
+    let s' = T.pack . streamEdit ghcPatternMatchReplacement id . T.unpack $ s
+    pure $ Apply () b (Constant () (someValueOf DefaultUniString $ s'))
+  where
+    ghcPatternMatchReplacement :: M.Parsec Void String String
+    ghcPatternMatchReplacement = do
+      -- What's being replaced (which is varying).
+      s <- M.string "Pattern match failure" <* M.takeRest
```
L-as, line 41: You can just pattern match rather than using parsec

plutarch-test/src/Plutarch/Test/Deterministic.hs
```diff
@@ -0,0 +1,68 @@
...
+{- Like `compile`, but the result is deterministic -}
+compileD :: ClosedTerm a -> Scripts.Script
+compileD p = rewriteTraces $ compile p
+
+{- Rewrite the string used by `Trace` so that the script becomes deterministic. -}
+rewriteTraces :: Scripts.Script -> Scripts.Script
+rewriteTraces =
+  walkScript $ \term -> do
+    -- Replace the 's' in `trace s`.
+    Apply () b@(Force _ (Builtin _ Trace)) (Constant () (Some (ValueOf DefaultUniString s))) <- pure term
```
L-as, line 34: Very often the trace function will be hoisted, so this won't work unfortunately.

plutarch-test/src/Plutarch/Test/Deterministic.hs
```diff
@@ -0,0 +1,68 @@
...
+    let s' = T.pack . streamEdit ghcPatternMatchReplacement id . T.unpack $ s
```
L-as, line 35: You don't need to use `streamEdit` since it's always in the beginning AFAIK.

L-as (changes requested):
This unfortunately doesn't fix the problem. Try having a test case with more than 1 use of `ptrace`.

plutarch-test/src/Plutarch/Test/Deterministic.hs
srid: True; that's an accidental artifact from printTerm rewrite commit. Removed now.

plutarch-test/src/Plutarch/Test/Deterministic.hs
srid: You are right, but I don't understand why we are hoisting `ptrace` in the first place The only solution here seems to be to inline it: #301

srid:
Alternative approach: do the rewrite in Plutarch (the library), and add `compile' :: CompileOpts -> ClosedTerm a -> Script`, where `CompileOpts` specifies "rewrite traces to be deterministic" (or a general variation of that; such as rewrite plugins?), and possibly more in future.

L-as:
We will likely need another solution for this, possibly including rewriting. I'll mark this as draft since the solution isn't clear

L-as:
See #316

L-as:
Replaced by #479

PR #301, opened 2022-02-16
Optimize `ptrace`

srid:
Optimize `ptrace` by inlining its use. Merge after #297 

Improves on cpu, mem and script size for the case where `ptrace` is used more than once.

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Unreleased
 
+- Optimisations:
+  - `ptrace'` is now inlined, and is slightly more efficient in cpu, mem and script size.
+
 - `TermCont`: Parametrize by result type; add `MonadFail` instance; etc.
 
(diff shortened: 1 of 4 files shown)


TotallyNotChase:
That's weird. I don't get why hoisting would make it worse. 
```hs
ptraceCore :: Term s (PString :--> a :--> a)
ptraceCore = pforce $ punsafeBuiltin PLC.Trace

ptraceHoisted :: Term s (PString :--> a :--> a)
ptraceHoisted = phoistAcyclic $ pforce $ punsafeBuiltin PLC.Trace

test :: Term s PByteString
test = ptraceCore # "foo" # phexByteStr "41"

test' :: Term s PByteString
test' = ptraceHoisted # "foo" # phexByteStr "41"
```
Both `test` and `test'` generate the same code here - can we have a minimal repro to understand what's happening here?

L-as:
There is a better solution than thin BTW. I will make a PR later. Waiting for CI to work again.

L-as:
Replaced by #316

PR #304, opened 2022-02-17
Upstream some list functions

Geometer1729:
(#274 had to be closed so this pr could be from a different branch.)


Geometer1729:
Replaced by #306

Issue #303, opened 2022-02-17
Add something like `PAsDataMaybe`


L-as:
See 
- https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109
- https://github.com/Plutonomicon/plutarch/pull/267#issuecomment-1043042454
- #277

MangoIV:
Hi, I have the following: 



```haskell 
class PTryFromRecur (a::PType) where
  ptryFromRecur :: Term s (PData :--> a)

instance PTryFromRecur PInteger where 
  ptryFromRecur = punsafeBuiltin PLC.UnIData

instance PTryFromRecur  PByteString where 
  ptryFromRecur = punsafeBuiltin PLC.UnBData
```

```haskell
-- this is the one I'm not sure about: 
instance (PTryFromRecur a, PLift a) => PTryFromRecur (PBuiltinList (PAsData a)) where 
  ptryFromRecur = phoistAcyclic $ plam $ \opq -> 
    let lst :: Term _ (PBuiltinList (PAsData PData))
        lst = punsafeBuiltin PLC.UnListData #$ opq
        res :: Term _ (PBuiltinList (PAsData a))
        res = pmap # (plam $ \e -> punsafeCoerce $ ptryFromRecur @a #$ pfromData e) # lst
     in res
```


Is the latter instance legal? I'm especially asking because of the punsafeCoerce I wrote. My reasoning is that if I ptryFromRecur @a whatever comes out must be verified and I can therefore just wrap it back into a PAsData. Is that correct?

Issue #190, opened 2022-01-26
Monoid/Semigroup instance for (Term s PValue)


nazrhom:
Would it make sense to have these instances available in Api.V1? I stumbled into this while trying to implement a Plutarch version of `valuePaidTo`.

L-as:
Sure, but it probably makes sense to split the module up internally.

Issue #139, opened 2022-01-18
plutarch-extra


Geometer1729:
At the meeting today we talked about creating a separate cabal project (probably in this repo) for libraries in Plutarch.

kozross:
For what it's worth, my experience with `plutus-extra` suggest that multiple independent repoes, with one per sublibrary, is a much better idea than a megarepo.

L-as:
> At the meeting today we talked about creating a separate cabal project (probably in this repo) for libraries in Plutarch.

What meeting?

L-as:
> For what it's worth, my experience with plutus-extra suggest that multiple independent repoes, with one per sublibrary, is a much better idea than a megarepo.

Elaborate?

Geometer1729:
> What meeting?

The sprint meeting this morning.

Geometer1729:
Is this a just timezone thing and the meeting was last night for you or something? Or did I misunderstand the conversation?

kozross:
Big repoes with multiple projects in them end up becoming quite unwieldy, especially when you have a large pool of overall contributors, but where most of them are focused only on one or two sub-libraries.

L-as:
@Geometer1729 it's possible I'm forgetting the conversation.

In any case, @kozross I agree, but we're in a state where Plutarch changes rapidly. If `plutarch-extra` is in this repository, then the changes can happen together.

Geometer1729:
I mentioned that plutarch is missing a lot of prelude analogs ie `pfst` and `psnd` for `PPair` and asked if you intended plutarch to eventually have functions like that and iirc it was your idea that there should be a separate module for library functions like that.

L-as:
@Geometer1729 Those things should just be in `Plutarch.Pair`. It's more things like `Natural` that might want a separate library potentially.

PR #306, opened 2022-02-17
Upstream some list functions.

Geometer1729:

#274 and #304 were both closed to change the source branch.


Plutarch/List.hs
```diff
@@ -363,3 +370,134 @@ plistEquals =
...
+          PSCons y ys' ->
+            pif
+              (x #<= y)
+              (pcons # x #$ self # xs' # ys)
+              (pcons # y #$ self # xs # ys')
+
+-- most of the tricks from timsort are not implemented here
+-- just the central strategy of finding runs of
+-- consecutive data and then merging them together
+ptimSort :: POrd a => Term s (PList a :--> PList a)
```
L-as, line 453: Why do we need this?

Plutarch/List.hs
```diff
@@ -363,3 +370,134 @@ plistEquals =
...
+      ( \y ys ->
+          pif
+            (f # y)
+            (pcon $ PJust y)
+            (self # f # ys)
+      )
+      (pcon PNothing)
+      xs
+
+psort :: POrd a => Term s (PList a :--> PList a)
```
L-as, line 407: When do you actually use sorting?

plutarch.cabal
```diff
@@ -169,12 +169,14 @@ test-suite examples
...
     , bytestring
     , generics-sop
+    , hedgehog
     , plutarch
     , plutus-core
     , plutus-ledger-api
     , plutus-tx
     , rank2classes       >=1.4.4
     , tasty
+    , tasty-hedgehog
```
L-as, line 179: We've committed to using sydtest rather than tasty.

plutarch-test/src/Plutarch/Test/Property/Util.hs
```diff
@@ -0,0 +1,286 @@
...
+import Control.Monad.IO.Class (liftIO)
+import Data.Proxy (Proxy (..))
+import Data.Text (Text)
+
+import Plutus.V1.Ledger.Scripts (Script (..), ScriptError)
+import PlutusCore.Evaluation.Machine.ExBudget (ExBudget)
+
+import Hedgehog (Gen, Property, PropertyT, annotate, annotateShow, assert, forAll, property, (===))
+
+class Marshal h (p :: PType) | h -> p where
```
L-as, line 37: How is this class related to this PR?

L-as (changes requested):
.

DarthPJB (commented):
I can't speak to @L-as 's questions, however the Nix code changes are acceptable.

plutarch.cabal
Geometer1729: sorry about that, I forgot to remove the import when moving to sydtest.

plutarch-test/src/Plutarch/Test/Property/Util.hs
Geometer1729: When testing agreement between Haskell functions and Plutarch functions it's convenient to be able to generate the arguments to the function as Haskell terms and marshal them into Plutarch terms, that way you have both representations and can run both functions. Marshal is also used to turn the output of the Haskell function into Plutarch to test equality at the end.

Plutarch/List.hs
Geometer1729: I originally implemented this because I thought we would need sorting to convert values into sorted list so that functions like unionWith can be more efficient. Hopefully the values come sorted or can be required to be sorted, but we haven't tested that yet.

Plutarch/List.hs
Geometer1729: I was worried that merge sort might be unnecessarily slow if a lot of the lists it gets called with are nearly sorted. I plan to benchmark the project using both and see what the tradeoff looks like.

L-as:
Two thoughts:
- I don't think we should have sorting honestly, because all parts of ScriptContext are either pre-sorted, or can be required to be sorted rather than actually sorting them. We discussed this in an internal meeting some time ago. We should make plutarch-extra with the helpers needed for this.
- I'll take a deeper look at Marshal

nixinator (commented):
don't eat hedge hogs , they are engaged, no matter how tasty they are.

TotallyNotChase:
I stumbled upon the lack of efficient sorting options when implementing `Semigroup` for `PValue`. That could use `unionWith`, and `unionWith` could benefit from sorted lookups and such. I can't speak to whether or not that belongs in this repo but `Semigroup (Term s PValue)` (or simply `unionWith`) is sort of high priority IMO.

Geometer1729:
> I stumbled upon the lack of efficient sorting options when implementing `Semigroup` for `PValue`. That could use `unionWith`, and `unionWith` could benefit from sorted lookups and such. I can't speak to whether or not that belongs in this repo but `Semigroup (Term s PValue)` (or simply `unionWith`) is sort of high priority IMO.

This was for the same reason (in fact we even have a monoid instance for PValue we want to upstream) but if Las is right and the values can be required to come sorted then unionWith can pretty easily preserver sortedness.

TotallyNotChase:
> > I stumbled upon the lack of efficient sorting options when implementing `Semigroup` for `PValue`. That could use `unionWith`, and `unionWith` could benefit from sorted lookups and such. I can't speak to whether or not that belongs in this repo but `Semigroup (Term s PValue)` (or simply `unionWith`) is sort of high priority IMO.
> 
> This was for the same reason (in fact we even have a monoid instance for PValue we want to upstream) but if Las is right and the values can be required to come sorted then unionWith can pretty easily preserver sortedness.

Do you think it's possible to upstream some of these `Value` utilities (not necessarily instances, just `unionWith` and similar) to another library that people can target? I actually need this stuff as well and I've been handwriting all of this. There's no reason for all users to go through this much definitions just to be able to use maps..... of all things. We could have a place for utility functions now that we can merge to `plutarch-extra` later as Las suggests.

emiflake:
+1 on having a clear place to upstream utility functions pre-`plutarch-extra`. What's blocking getting `plutarch-extra` going, though?

TotallyNotChase:
screw it, let's just get plutarch-extra ready alongside 1.2 release

@MangoIV could you make the initial plutarch-extra scaffold in this repo please? It should be a separate library - doesn't necessarily need to contain any real code immediately.

@Geometer1729 could you prepare the PRs for the code of this please? Most of the functions in this PR should also go to `plutarch-extra`, and whatever else utilities necessary - be sure to separate the PRs though. However, you should add the instances and utilities for `PMap` (and `PValue`) in Plutarch itself. We'll then have the fundamental utilities for the fundamental data types - though we're probably missing a few for `Plutarch.Pair`.

@L-as are you ok with this plan? anything you want to change?

Relevant issues
* https://github.com/Plutonomicon/plutarch/issues/139
* https://github.com/Plutonomicon/plutarch/issues/190

please make sure to mark any relevant PRs with this issues.

MangoIV:
Sure, I can do that, I'm currently working on #303 but I can do this at <some idea of a deadline you insert ;)>.

L-as:
@TotallyNotChase sounds good to me

L-as:
#357

L-as:
Closing this in favor of #357

PR #307, opened 2022-02-17
Improve test spec ergonomics

### LINKED ISSUE #289: Improve ergonomics of golden testing
> While not a blocking issue, I do wonder how we could make the testing API more ergonomic.

> I'll think about this outside this PR. Maybe some sort of MonadWriter that collects programs on the goal, and batches golden tests at the end ...

_Originally posted by @srid in https://github.com/Plutonomicon/plutarch/pull/276#discussion_r807401889_

srid:
Resolves #289 

- Add monadic syntax with operators to define goldens hierarchically _in addition to_ also asserting evaluation output. This enables us to _always_ write golden for Plutarch programs as well as (optionally) assert them in the same place.
  - Old golden code, including `PrintTerm` and `Bench` conditionals, are no longer used, but are nevertheless marked as deprecated for backwards compatibility (They should be removed eventually, in 1.2+)
    - `benchmarkScript'` now returns on failing scripts
- Reduce the number of golden files by grouping at higher level

[^1]: I'm hoping that we make traces inline (#301) such that it makes #297 correct.

## Syntax

Goldens, without assertion:

image

Goldens, with optional assertion:

image

image


blamario:
Do we have an ETA for the merge? I need to touch many golden tests and I'd prefer to do it once.

srid:
Total golden file count has been halved -- from 183 -> 91.

L-as:
Great, is this missing something for it to be merged?

L-as:
You're not merging into master BTW

srid:
@L-as @blamario This is now ready for review and merge.

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
...
+
+mkGoldenValue' :: ClosedTerm a -> Maybe Expectation -> GoldenValue
+mkGoldenValue' p mexp =
+  GoldenValue
+    (T.pack $ printScript $ compileD p)
+    (T.pack $ printScript $ evaluateScriptAlways $ compileD p)
+    (TL.toStrict $ Aeson.encodeToLazyText $ benchmarkScript' $ compileD p)
+    mexp
+
+instance HasGoldenValue (Term s a) where
```
L-as, line 68: This is really bad IMO. This will fail at run-time when you try to use it on non-closed terms.
Can I ask why you need this?

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
...
+{- A `Term` paired with its evaluation expectation
+
+  Example:
+  >>> TermExpectation (pcon PTrue) $ \p -> pshouldBe (pcon PTrue)
+
+-}
+data TermExpectation s a = TermExpectation (Term s a) (Term s a -> Expectation)
+
+{- Test an expectation on a golden Plutarch program -}
+(@->) :: Term s a -> (ClosedTerm a -> Expectation) -> TermExpectation s a
```
L-as, line 80: This should be `ClosedTerm a -> .(ClosedTerm a -> ...)`

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
+{-# LANGUAGE ImpredicativeTypes #-}
+
+module Plutarch.Test.Golden (
+  pgoldenSpec,
+  (@|),
+  (@\),
+  (@->),
+
+  -- * Internal
```
L-as, line 9: Internal in what way?

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
+{-# LANGUAGE ImpredicativeTypes #-}
```
L-as, line 1: If you don't want to support GHC 8.10, then why not make the instance for `ClosedTerm`?

L-as (changes requested):

L-as:
Looks great mostly!

plutarch-test/src/Plutarch/Test.hs
```diff
@@ -69,26 +86,17 @@ pshouldBe x y = do
 passert :: ClosedTerm a -> Expectation
 passert p = p #@?= pcon PTrue
 
+{- Asserts the term to be false -}
```
blamario, line 89: Shouldn't this be `{- | Asserts the term to be false -}`, so Haddock picks it up?

blamario (dismissed):
Looks good to me, except several comments in `Plutarch.Text.Golden` should be haddocked.

plutarch-test/src/Plutarch/Test/Golden.hs
srid: It exists mainly to provide the optional "And then check that this Plutarch program matches this expectation" syntax

```
    "por" @\ do
      "tf" @| pcon PTrue #|| pcon PFalse @-> passert
```

That `@-> passert`, basically. The type class treats the following (that comes after `@|`) polymorphically):

- `pcon PTrue #|| pcon PFalse`
- `pcon PTrue #|| pcon PFalse @-> \p -> passert p`

---

> This will fail at run-time when you try to use it on non-closed terms.

IIRC GHC prevented me from deriving for `ClosedTerm a`, forcing me to use `Term s a` ... let me check this once again.

plutarch-test/src/Plutarch/Test/Golden.hs
srid: Well, `plutarch-test` the library should be supported in 8.10 -- we are using this in our smart contract.

plutarch-test/src/Plutarch/Test/Golden.hs
srid: Yup, I can't do this:

image

L-as: If you can make `@|` and such take closed terms and not export the instance, then it should be safe actually.

plutarch-test/src/Plutarch/Test/Golden.hs
L-as: Well then you should probably remove this.

plutarch-test/src/Plutarch/Test/Golden.hs
srid: Internal to `Plutarch.Test` (and not re-exported, like the list above), but since this module is not exposed, it makes no sense I guess.

plutarch-test/src/Plutarch/Test/Golden.hs
srid: Okay, should be good now. `ClosedTerm` everywhere except in this instance and `unsafeClosedTerm`.

L-as:
Why not use a newtype wrapper over ClosedTerm?

srid:
> Why not use a newtype wrapper over ClosedTerm?

I had thought about it, but it had me wondering - what's the benefit of that? 

Switching to using a newtype would still have me derive `instance HasGoldenValue (Term s a)` anyway because `(@|)` has to polymorphically accept a term directly (if not `TermExpectation`), assuming we don't want to make the user wrap their terms in this newtype.

srid:
Ah, hold on - I got something working to get rid of this unsafe function.

srid:
Uh. I thought of using type families:

```haskell
class HasGoldenValue a where
  type GoldenValueFrom a :: Type
  mkGoldenValue :: GoldenValueFrom a -> GoldenValue

instance HasGoldenValue (Term s a) where
  type GoldenValueFrom (Term s a) = ClosedTermN a
  mkGoldenValue (ClosedTermN p) = mkGoldenValue' p Nothing
```

But that leads ambiguity at use sites:
image

srid:
After fce969cc31e0475946c226bb3c7750cef82f826c `unsafeClosedTerm` is used in exactly one place - that "Term s a" instance. I think this is as good as it gets.

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
...
+  baseDir </> T.unpack k <> ".golden"
+
+-- | Group multiple goldens values in the same file
+combineGoldens :: [(GoldenKey, Text)] -> Text
+combineGoldens xs =
+  T.intercalate "\n" $
+    (\(GoldenKey k, v) -> k <> " " <> v) <$> xs
+
+-- | Specify goldens for the given Plutarch program
+(@|) :: HasGoldenValue v => GoldenKey -> v -> ListSyntax (GoldenKey, GoldenValue)
```
L-as, line 111: So this is still unsafe unfortunately. `v` could be a `Term s a` that references free variables.
You could split this up into two functions, one for `ClosedTerm a` and one for `TermExpectation a`.
Another possible solution is to add the `s` to `HasGoldenValue`, such that it is `(forall s. HasGoldenValue s v) =>`. The instance for `TermExpectation` would not make use of the `s`.

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -0,0 +1,184 @@
...
+  Expectation,
+  Spec,
+  TestDefM,
+  describe,
+  getTestDescriptionPath,
+  it,
+  pureGoldenTextFile,
+ )
+
+import Plutarch
```
L-as, line 35: You are BTW supposed to use `Plutarch.Prelude`, and not `Plutarch`.

plutarch-test/src/Plutarch/Test/Golden.hs
L-as:
```haskell
bad' :: forall s a. GoldenKey -> Term s a -> ListSyntax (GoldenKey, GoldenValue)
bad' k v = listSyntaxAdd (k, mkGoldenValue @s v)

xc :: GoldenKey -> (ListSyntax (GoldenKey, GoldenValue) -> Term s a) -> Term s (POpaque :--> a)
xc key f = plam $ \(x :: Term _ POpaque) -> f (bad' key x)
```

L-as: This is quite obviously nonsense.

L-as: I fixed this myself.

L-as:
I'll merge this so others can build on top of it, though my latest commits might not be perfect.

PR #309, opened 2022-02-18
Changed the Scott encoding to strict nullary constructors

blamario:
I've hit a snag with the FFI regarding the difference in Scott encodings of ADTs. Take for example the encoding of `Maybe a`. In Plutarch type notation, the two Scott encodings of its type are

    type PlutarchMaybe = (a :--> r) :--> PDelayed r :--> r
    type PlutusTxMaybe = PDelayed ((a :--> r) :--> r :--> r)

Unsurprisingly, the PlutusTx encoding sucks because it requires lots of forcing and delayings. But I also wonder why Plutarch needs that `PDelayed` on the nullary constructor `PNothing`. It's useful only if the `Nothing` branch of the destructor needs to be lazy, but I'd expect that to be a minority of cases - and those can be explicitly lazified.

This draft PR demonstrates the effects. I've only fixed the test suite on the root project, leaving alone `plutarch-test` and `plutarch-benchmark`. For those tests in `plutarch-test` that continue to pass, the effect is a reduced budget overall.


L-as:
This completely breaks do notation FYI. `PJust x <- y` will always fail with this

Issue #105, opened 2022-01-13
Improve the Developers' guide


TotallyNotChase:
Right now, the [Developers' guide](https://github.com/Plutonomicon/plutarch/blob/master/docs/DEVGUIDE.md) is quite barebones. Ideally it should explain all the concepts that Plutarch is built upon, under the hood. All the low level details about terms, `PDataRepr`, `PLift`, and more.

The following topics need to be covered, as a starting point:-
- [ ] `PConstant`/`PLift`. This should include all the low level details. Be sure to include info on `DefaultUni`. Though you can also use existing cross references as relevant.
- [ ] `PDataRepr`. This should include the following:-
  - [ ] Anatomy of the pattern matching in general and `DataReprHandlers`.
  - [ ] Anatomy of the derivers (generic and `PIsDataReprInstances`)
  - [ ] Anatomy of the field accessors
- [ ] `Term`/`ClosedTerm` and how code gen works in general. Ideally, it should provide some information on how Plutarch types are merely "tags", and their Haskell ADT hardly matters, and how these tags are actually used to "tag" concrete Plutus Core constants (or lambdas, for scott encoding).
- [ ] The machinery behind `PLam` (`plam`).

srid:
See also #107

TotallyNotChase:
I'll probably work on this issue post Plutarch 1.1. Self assigning for now.

TotallyNotChase:
I added a task list to this. If you have sufficient familiarity with any of this topics, please feel free to contribute and help make the developers guide better!

PR #310, opened 2022-02-18
expand beginner guide material and rearrange some sections

peter-mlabs:
I felt that the Plutarch guide lacked a solid, introductory entry-point. This commit is an attempt to rectify that. I've removed the `SYNTAX.md` and corresponding entry in the top-level `GUIDE.md` and replaced it with an `INTRO.md`. `INTRO.md` contains substantially the same material, but introduced at a slower pace with a focus on building intuition one concept at a time. I also felt that certain sections were misplaced among `INTRO.md`, `USAGE.md`, and `CONCEPTS.md`; this PR also contains my suggestions there.

Things I tried to emphasize:
- Providing context for code snippets/examples before the snippet, and additional description after. Some existing examples dropped immediately into a snippet before indicating what the reader should be looking for, or provided too much description before the actual code was diplayed.
- Explicitly defining and Using consistent formatting and terminology. This includes stylizing Plutarch terms as "`Term`" rather than "term" or "Term", being explicit about when `PType` refers to a tag vs. a type vs. a kind, and being strict about distinguishing between functions-between-terms and lambda terms.
- Providing a high-level overview, but otherwise introducing concepts one-by-one and not using later concepts in earlier prose or examples.
- Providing more rudimentary and fleshed-out examples, including displaying the intermediate types and kinds
- Building intuition via comparison with beginner Haskell.

Edits may be necessary, both for validity (as I'm learning most of these concepts for the first time myself) and grammar, flow, etc. Feedback is appreciated.


peter-mlabs:
Just pushed some additional edits focused on reducing wordiness and using the active voice. Content is still the same.

L-as:
Most of these changes are really good, but due to it being largely automated, some of the meaning has been mangled.

peter-mlabs:
> concrete != tangible in this context

I wasn't sure whether I should have accepted this suggestion or not. I'll revert.

L-as:
You might want to wait for #294 , though really great work @peter-mlabs !

peter-mlabs:
> You might want to wait for #294 , though really great work @peter-mlabs !

Thanks!

I did branch off Jack's guide branch, so most of the work in #294 should be included. Once that's finalized I'll incorporate any final changes

L-as:
I didn't realise you're targetting `guide`! @TotallyNotChase maybe you want to merge this so the work isn't lost?

TotallyNotChase:
Definitely want to merge this before #294 @L-as

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+```hs
+-- | Closed term is a type synonym 
+type ClosedTerm (a :: PType) = forall (s :: S). Term s a 
+
+-- | Compile operates on closed terms to produce usable UPLC scripts.
+compile :: ClosedTerm a -> Script
+```
+`Term`s are constructed from Haskell values and are tagged with `PType`s.
+
+## Plutarch Constant `Term`s
```
TotallyNotChase, line 137: I think the details of the typeclasses (`PConstant`/`PLift` and `PlutusType`/`PCon`/`PMatch`) should stay at their own section in the Typeclasses document. The constant building could stay surface level here, like it was before. What do you think?

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+it _does_ translate to a lazy Plutarch lambda.
+-}
+hif :: Term s PBool -> Term s a -> Term s a -> Term s a
+hif cond whenTrue whenFalse = pforce $ pif' # cond # pdelay whenTrue # pdelay whenFalse
+
+{- | Plutarch-level `if`. We delay the arguments in `hif` so they are not applied
+until forced.
+-}
+pif :: forall {s :: S} {b :: PType}. Term s (PBool :--> (b :--> (b :--> b)))
+pif = plam pif'
```
TotallyNotChase, line 428: Not quite - I've been bitten by this before as well! Once you make a function into a Plutarch level function - there is no delaying the arguments anymore.  See, when you apply this `pif` - the arguments are already evaluated as soon as it enters the body! There's no point in delaying the arguments inside the body of a Plutarch lambda - the deed has been done!

I had a note about this in the "Delay and Force" section of Concepts-
> Delaying the argument to a Plutarch level function, within the function body, is not very useful - since the argument has been evaluated before the function body has been entered! Instead, you'll often notice usage of `pdelay` and `pforce` in Haskell level function arguments to simulate laziness.

Let's make sure we keep a similar note on this overhaul as well just so users don't make the same mistake!

Also I assume you meant to use `plam hif` - otherwise this review comment makes no sense lol

docs/USAGE.md
```diff
@@ -5,50 +5,22 @@ This document describes various core Plutarch usage concepts.
...
-infixl 8 #
-
-infixr 0 #$
-```
-
-`#$` is pretty much just `$` for Plutarch functions. But `#` is left associative and has a high precedence. This essentially means that the following:
-
-```haskell
-f # 1 # 2
+<small><i><a href='http://ecotrust-canada.github.io/markdown-toc/'>Table of contents generated with markdown-toc</a></i></small>
```
TotallyNotChase, line 19: what's this?

TotallyNotChase (changes requested):
This is awesome! This is the introduction the guide needed, for sure. There's a few things we can improve though.

Also, as a general review note: if possible, I think we should cut down the intro a bit. We don't have to provide *too much* information in it. Ideally, it should be pretty basic (but not as basic as my original `SYNTAX.md` lol)

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
+> Note: If you spot any mistakes/have any related questions that this guide lacks the answer to, please don't hesitate to raise an issue. The goal is to have high-quality documentation for Plutarch users!
+
+<details>
+<summary> Table of Contents </summary>
```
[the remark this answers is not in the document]
peter-mlabs: Ah interesting -- I used an online TOC generator, and it seems like it mangled more than that. I'll fix.

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+
+When we talk about "Plutarch scripts," we are referring to values of type `Term (s :: S) (a :: PType)`. `Term` is a `newtype` wrapper around a more complex type, the details of which Plutarch end-users can ignore. A `Term` is a typed lambda term; it can be thought of as representing a not-yet-evaluated computation that, if successfully evaluated, will return a value of type `a`. 
+
+The two type variables of the `Term s a` declaration have particular kinds: 
+
+- `s:: S` is like the `s` of `ST s a`. It represents the computation context in a manner that mimics mutable state while providing a familiar functional interface. Sections 1 through 4 of [1] give an accessible introduction to how this works. `s` is never instantiated with a concrete value; it is merely a type-level way to ensure that computational contexts remain properly encapsulated (i.e., different state threads don't interact). For more in-depth coverage of this and other eDSL design principles used in Plutarch, see [2].
```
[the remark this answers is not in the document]
peter-mlabs: fixed

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+- `a :: PType` is short-hand for "Plutarch Type". We prefix these types with a capital `P`, such as `PInteger`, `PBool`, and so forth. _Tagging_ a `Term` with a `PType` indicates the type of the `Term`'s return value. Doing this allows us to bridge between the simple type system of Plutarch and the untyped UPLC -- the tags facilitate efficient conversions to and from the `Data` representation required by UPLC terms and the simple type system of Plutarch ` Term`'s. On the other hand, it also can provide an easier way to convert between Haskell values and Plutarch ` Term`'s via type synonyms.
```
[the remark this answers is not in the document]
peter-mlabs: Got it -- this makes sense. It just clicked to me that Scott encoding isn't just a "different" way to represent the `Data`, its _actually_ using lambdas. Cool!

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+- B.) Working with _Plutarch `Terms`_, which are values of the type `Term (s :: S) (a :: PType)`. These are the Plutarch scripts themselves, from which we build up more complex scripts before compiling and executing them on-chain.
+- C.) Writing Haskell-level functions _between Plutarch Terms_ (i.e., with types like `Term s a -> Term s b`). Doing so allows us to leverage Haskell's language features and ecosystem.
+- D.) Efficiently Converting the functions from _(C.)_ to _Plutarch-level functions_, which are of the type `Term s (a :--> b)`. We can _directly_ convert the functions from (C.) to Plutarch-level functions at the most naive level using `plam`. Additional Plutarch utilities provide for optimization opportunities.
+- E.) Compiling and executing the functions from _(D.)_, targetting UPLC for on-chain usage. 
+
+As a preview, the bridge Plutarch provides between Haskell and UPLC looks something like this:
+
+```
+------------------------------------------------------
```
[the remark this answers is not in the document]
peter-mlabs: Yeah, I wasn't super happy with this chart myself. I'll revisit these suggestions later and try to draw up something better. Happy to incorporate more suggestions if you think of them

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+
+# Untyped Plutus Core (UPLC)
+
+Plutarch compiles to UPLC. Most Plutarch end-users will not need to concern themselves with the details of UPLC, but a brief overview will aid in building a mental model of how Plutarch works.
+
+Unlike Haskell, UPLC is a low-level and untyped language implementing a basic lambda calculus. Consequently, it supports only a handful of built-in values and functions which may be strung together in lambda applications. The built-in types provided by UPLC include the usual primitive types -- integers, byte strings and strings, booleans, and so forth -- and a special `Data` value that can encode representations of arbitrary sum-of-products Haskell types.
+
+While the _semantic_ meaning of a Haskell type such as `Maybe Integer` is missing in UPLC, it still can be _represented_ as `Data` in UPLC. UPLC doesn't track what differentiates semantically distinct `Data` values and will not prevent a programmer from operating on the `Data` representation in non-sensical ways.
```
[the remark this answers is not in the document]
peter-mlabs: As above -- this is clicking for me now. I'll go through and make sure that this is less `Data` centric, perhaps with an additional line/link or two somewhere to explain that `Data` and Scott encodings are the encodings we're concerned with.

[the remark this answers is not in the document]
peter-mlabs: fixed

docs/INTRO.md
```diff
@@ -0,0 +1,441 @@
...
+`PType` is defined as `type PType = S -> Type`; that is, it is a _kind synonym_ for `S -> Type` (where `S` and `Type` are themselves kinds). This synonym is important to keep in mind because when querying the kind of something like `PBool` in, say, GHCi, we will _not_ see `PType` as the kind. Instead, we get 
+
+```hs
+ghci> :k PBool
+PBool :: S -> Type
+```
+
+Thus, any time we see the kind `S -> Type`, we should mentally substitute its kind synonym `PType`. We reiterate: types of kind `PType`, should be considered as _tags_ on computation. They do not represent types of values in the same way as standard Haskell types. 
+
+The kind of basic types such as `Integer` in Haskell is `*`; the corresponding "basic" kind in Plutarch is simply `PType`. Higher-kinded types in Haskell, such as `Maybe`, will kinds such as `* -> *`. In Plutarch, the corresponding kind is:
```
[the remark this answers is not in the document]
peter-mlabs: This is a good call.  Is there actually a difference between `Type` and `*`, or are they synonyms? I hadn't encountered the `Type` kind before MLabs, so I think it may be worthwhile to mention whether or not there's a distinction.

docs/INTRO.md
peter-mlabs: > Also I assume you meant to use plam hif - otherwise this review comment makes no sense lol

Yep! Oops :)

> Not quite - I've been bitten by this before as well! Once you make a function into a Plutarch level function - there is no delaying the arguments anymore. See, when you apply this pif - the arguments are already evaluated as soon as it enters the body! There's no point in delaying the arguments inside the body of a Plutarch lambda - the deed has been done!

Ah! On review, I see the difference. Just to be clear -- `hif` is the lazy evaluated version, and `pif` will still evaluate its arguments as they're passed in. Is that right?

Two comments here:

- As you can see, I tried to adopt the convention of using `h(...)` to indicate that it was function between `Term`s and `p(...)` to indicate that it was a `Term s (a :--> b)` Plutarch lambda. Is this convention sensible to you -- either strictly for the purposes of the guide, or as a style choice in production code? I'm a little shaky still on whether doing things this way is bad practice/causes bloat
- There's another section "When to use Haskell level functions?" section of Tips & Tricks (which I hadn't re-read before writing the Intro) that indicates that we can use these functions, but I'm not sure that I totally understand how/when we can actually compile Haskell functions (if that's the right way of thinking about it?). I think a brief indication that something like this is possible should be in `INTRO.md` so reaaders don't walk away with the wrong idea. Thoughts?

docs/USAGE.md
peter-mlabs: Ah -- hyperlink from the TOC generator I used.

BTW, it appears that it had some different anchor rewriting rules than GitHub uses. There were a bunch of broken links across the various files that I touched, but they should all be fixed now.

docs/INTRO.md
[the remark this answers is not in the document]
TotallyNotChase: They are one and the same. `*` is Haskell 2010's way of pronouncing `Type`. Times have changed a lot since then: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/poly_kinds.html#the-kind-type

docs/INTRO.md
TotallyNotChase: I do agree with that. I'm just a little concerned about too much duplicate info and the guide getting too big and unwieldy. But I think you'd actually be a much better judge for whether or not the guide, at its current stage, seems too big.

My concern is that there's *a lot* of really crucial and helpful information in all the corners of the guide - but when it gets too big, no one's going to read all of it and hence the information just sits there. I'm hopeful that the splitting effort will make it easier to grasp - and perhaps I'll scrape off some info from other sections if a lot of it will be covered in intro.

docs/INTRO.md
TotallyNotChase: yes, `hif` will take care of delaying the arguments before calling the strict plutarch level function.

`h(...)` makes sense to me. There is no bloat, not from a codegen point. I'd like to emphasize that there's no magic going on in Plutarch - but that's sort of a misnomer, isn't it? The Haskell runtime is Plutarch compile time - so any "bloat" from Haskell couldn't be a detriment for Plutarch. I don't know if it's easy to convey this concept of "there's no magic here - it's all rather simple" into the guide (I haven't been able to do so) - but I do think it would be of tremendous help. I'll try to log some of my thoughts here.

Since the Haskell is evaluated at Plutarch compile time - you can use the well defined evaluation order (just graph reduction) to realize exactly how the codegen is going to work. When you write:
```hs
hf x y = ...

pf = plam hf
```
You can follow the evaluation order until you reach a "Plutarch primitive" (somewhat self explanatory, I guess? `plam`, `pdelay`, `pforce`, synonyms to builtin functions (low level details) etc). According to that, this becomes `pf = plam (\x y -> ...)` - and there you go, `hf` is gone. It got "inlined" - according to reduction.

I showcased another example of this through `pmatch` above.

This is also why there's nothing to discuss around "using/compiling haskell level functions". You just kind of.... do it ¯\\\_(ツ)_/¯. You use it in some places, maybe in some of your Plutarch functions, or maybe in another Haskell function. In the end, you put it all together into one single validator function maybe - and compile that. As long as you end up with a Plutarch level term in the end, you can compile it. It doesn't matter what you used in its body - if you did something wrong, the type system will catch it. If you use a Haskell level function in your Plutarch term, it'd simply be inlined in Haskell runtime, then Plutarch codegen can go on its merry way to operate on the Haskell function's body.

Feel free to let me know if something unclear in there - I promise that the codegen is *very* simple if taught correctly (or observed). I haven't managed to actually explain it correctly anywhere - I wanted to do that in the `DEVGUIDE` sometime (see: #105). But at the end of the day, it's possible to follow Haskell's simple evaluation order with pen and paper, and Plutarch codegen will be crystal clear - we just have to document those "pen & paper" steps somewhere heh.

L-as:
`Type` is preferred

docs/INTRO.md
peter-mlabs: Yeah, I think my (personal) ideal guide has a pretty big intro. My goal is that you read this big intro/overview once to get a picture of how everything is laid out, and then don't really need to go back to it ever again because you've got tutorials, how-tos, and reference documentation that is better suited for actual development.

Perhaps we should make this clear, and think about what goes where. I think redundancy and repetition can be good in some scenarios -- its annoying to have to hop around from resource to resource -- but each time the same information is stated should be A.) be consistent, and B.) serve a distinct, explicit purpose.

Perhaps we should consider just going full-bore on the diataxis framework and plan on separating out 

- Explain: call this "The Guide" and include `INTRO.md` and some sections from concepts/usage
- Reference: The typeclass/types docs, and maybe some additional function reference material that aren't fully suitable for haddock comments, plus some of the shorter examples
- Tutorials: when we get there, building an end-to-end app using plutarch, or showing real-world cases of optimization, common mistakes, etc. Long-form examples.
- HOWTOs: fold in the tips & tricks, common issues and trouble shooting, short-form examples, etc.

I think that would be out of scope for this PR, but useful going forward. Thoughts?

docs/INTRO.md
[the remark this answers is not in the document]
peter-mlabs: Got it. I've added a blockqutoe note with this info and a link to basic type level programming concepts. I found it a helpful post to skim when I was trying to get up to speed on these things a few months ago.

docs/INTRO.md
peter-mlabs: Oh right -- we _can_ go through and make `Term s b` by applying a `Term s a` to a `Term s (a :--> b)`, but we can also just get it from any function that yields a `(...) -> Term s b`...

I think that it may be instructive to go through annotated examples of what things actually compile in UPLC terms. Its probably not suitable for `INTRO.md`, but it could help build that intuition. 

Admittedly, I still lack that intuition myself. It's not clear when a `Term s (a :--> b)` is _needed_ versus a `Term s a -> Term s b`, but I'm sure this will become more clear when I start writing more code. 

FWIW, this is covered in "When to use a haskell function" and "Prefer plutarch level functions", but I'm still struggling with it an can't totally pin down why. If I had to guess, I think its because I don't totally understand hoisting/plet, and the phrase:

> Haskell level functions will always be inlined when generating the Plutus Core. 

is non-obvious and kinda breaks my mental model of Plutarch (its not immediately clear that we can just inline arbitrary haskell functions or how that actually looks). This makes the example and the paragraph

> However, not all higher order functions benefit from taking Haskell level functions. In many higher order function use cases, you could benefit from passing a commonly used function argument, rather than a one-off function argument. Imagine map, you don't always map with one-off functions - often, you map with existing, commonly used functions. In these cases, that commonly used function ought to be a Plutarch level function, so it can be hoisted and map can simply reference it.

less informative than I assume they would be otherwise.

docs/INTRO.md
TotallyNotChase: That sounds good to me. I'll keep that in mind. But I don't think I'll be able to do it during this split. Perhaps you could make an issue about this as well that we can track. Otherwise, I will put it in my todo list.

peter-mlabs: Done, spun it off. into a new issue. Happy to contribute as time allows -- or as my own needs necessitate it :smile:

docs/INTRO.md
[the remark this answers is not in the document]
TotallyNotChase: Removing the `pfromData/pdata` part is enough in my eyes, as those are nothing special and don't establish any bridges between worlds. You can fill that gap with `compile` if you want - that fills the role of "Plutarch -> UPLC". I'm not too sure about the reverse - @blamario is working on FFI that might achieve such a relation eventually.

You *could* perhaps have another box showcasing Haskell functions working on `Term`s - these can be converted to Plutarch term world using `plam`, of course. The reverse relation is also possible, something like:
```hs
pf :: Term s (a :--> b :--> c)

-- recovering a haskell level function from a plutarch function
hf :: Term s a -> Term s b -> Term s c
hf x y = pf # x # y
```

I don't think that box is necessary though - I personally don't like treating functions and regular values differently. But if you believe it could potentially help users - I have absolutely nothing against it.

docs/GUIDE.md
```diff
@@ -28,41 +28,52 @@ The Plutarch guide is your one stop shop for getting up to speed on Plutarch!
...
-## Usage
-
-- [Applying functions](./USAGE.md#applying-functions)
-- [Conditionals](./USAGE.md#conditionals)
-- [Recursion](./USAGE.md#recursion)
-- [Do syntax with `TermCont`](./USAGE.md#do-syntax-with-termcont)
-- [Do syntax with `QualifiedDo` and `Plutarch.Monadic`](./USAGE.md#do-syntax-with-qualifieddo-and-plutarchmonadic)
-- [Deriving typeclasses for `newtype`s](./USAGE.md#deriving-typeclasses-for-newtypes)
-- [Deriving typeclasses with generics](./USAGE.md#deriving-typeclasses-with-generics)
+- [Overview](#overview)
```
TotallyNotChase, line 35: could you update these hyperlinks please?

docs/INTRO.md
TotallyNotChase: > its not immediately clear that we can just inline arbitrary haskell functions or how that actually looks

yeah.... it's likely a gap in my explanation. Maybe it helps to realize that this isn't anything specific to Plutarch or how it works: this is just how you can evaluate any pure functional language (since with the lack of side effects, all your functions are just equations that you can keep inlining and reducing to obtain a result).

Perhaps I can demonstrate the full codegen step, from the user facing level to the absolute low level with a simple example-
```hs
-------------------------
------- your code -------
-------------------------

add :: Term s PInteger -> Term s PInteger -> Term s PInteger
add x y = x + y

padd5 :: Term s (PInteger :--> PInteger)
padd5 = plam (hf 5)

-------------------------
------- evaluation ------
-------------------------

{- this is what happens during Haskell runtime
it's semantically equivalent to STG evaluation
but it's much simpler since we don't care about
GHC optimizations in an explanation
-}

-- note: PLC refers to `import qualified PlutusCore as PLC`

padd5
plam (hf 5)
plam (\y -> 5 + y) -- (∵ `hf x y = x + y`)
plam (\y -> (\a b -> punsafeBuiltin PLC.AddInteger # x # y) 5 y) -- (∵ `x + y = punsafeBuiltin PLC.AddInteger # x # y`)
plam (\y -> punsafeBuiltin PLC.AddInteger # 5 # y)
plam (\y -> punsafeBuiltin PLC.AddInteger # pconstant (5 :: Integer) # y) -- (∵ Overloaded `Num` literals, `fromInteger = pconstant` for `instance Num PInteger`)
plam
  (\y ->
    punsafeBuiltin PLC.AddInteger
      -- "PLC.Some (PLC.ValueOf PLC.DefaultUniInteger 5)" is how you represent a Plutus Core builtin integer constant with value 5
      # punsafeConstant (PLC.Some (PLC.ValueOf PLC.DefaultUniInteger 5))
      # y
  )
-- reached a saturation point - everything here is a "Plutarch primitive"
-- now those plutarch primitives are going to be reduced to build an AST representing UPLC
{- I'm going to skip most of the details here - but evaluation of these functions are all the same
All haskell functions (doesn't matter if they are related to this repo) can be thought of as evaluating
in this manner.
Note that by "haskell functions", I'm not referring to "Haskell functions operating on Plutarch terms" here.
We're hardly talking about Plutarch anymore!
-}
plam
  (\y ->
    Term (\_ -> TermResult (RBuiltin PLC.AddInteger) [])
      -- "PLC.Some (PLC.ValueOf PLC.DefaultUniInteger 5)" is how you represent a Plutus Core builtin integer constant with value 5
      # Term (\_ -> RConstant (PLC.Some (PLC.ValueOf PLC.DefaultUniInteger 5)) [])
      # y
  )
```
The lambda applications get significantly more complicated and bloated - but I assure you that in the end, it just inlines and inlines until it reduces to a data structure: [`Term`](https://github.com/Plutonomicon/plutarch/blob/master/Plutarch/Internal.hs#L135). That's the AST that we can now serialize into UPLC. Plutus itself provides machinery to do such a thing.

References:
* [`PLC.AddInteger`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore.html#t:DefaultFun) - this is a builtin function. That is, UPLC builtin.
* [Constant building (internal)](https://github.com/Plutonomicon/plutarch/blob/master/docs/DEVGUIDE.md#plutus-core-constants-unsafe)
* [`Some`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore.html#t:Some), [`ValueOf`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore.html#t:ValueOf), and [`DefaultUni`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore.html#t:DefaultUni).
* [`RawTerm`](https://github.com/Plutonomicon/plutarch/blob/master/Plutarch/Internal.hs#L70)

docs/GUIDE.md
peter-mlabs: Yep, fixed a bunch of these. I think I got them all.

docs/INTRO.md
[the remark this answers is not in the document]
peter-mlabs: Edited the diagram and added your example. I think its as good a place as any to introduce the convention the `h` indicates a function-between-terms and `p` indicates a lambda term

docs/INTRO.md
peter-mlabs: I'm assuming the first two lines should be 

```
hf :: Term s PInteger -> Term s PInteger -> Term s PInteger
hf x y = x + y
```

yeah?

TotallyNotChase: ah, yeah it is. I was supposed to replace all the `hf`s with `add` but only one was actually affected

peter-mlabs: This is a good example. I haven't looked into UPLC much at all, but I found it intuitive. I guess the idea is that as long as we reach a "Term" it doesn't matter if that's through lambda application or function application.

Next step (for me) is building intuition for how how hoisting and `plet` work in terms of the (high level) UPLC codegen. 

I think both these things could be placed somewhere in an intermediate document if they don't already exist, but are out of scope for the intro right now.

peter-mlabs:
Okay, I think I've covered all points raised so far. Let me know if I've missed anything?

docs/INTRO.md
TotallyNotChase: > Next step (for me) is building intuition for how how hoisting and plet work in terms of the (high level) UPLC codegen.

It's just a function application. Exploit the fact that function application is strict and therefore the argument to the function is a strictly evaluated binding and you have the exact semantics needed by `plet`.

> I think both these things could be placed somewhere in an intermediate document if they don't already exist, but are out of scope for the intro right now.

Eventually, yeah - it should all be in `DEVGUIDE`.

TotallyNotChase (approved):
Thanks for the incredible contribution to the guide!

docs/INTRO.md
```diff
@@ -0,0 +1,462 @@
...
+it _does_ translate to a lazy Plutarch lambda.
+-}
+hif :: Term s PBool -> Term s a -> Term s a -> Term s a
+hif cond whenTrue whenFalse = pforce $ pif' # cond # pdelay whenTrue # pdelay whenFalse
+
+{- | Plutarch-level `if`. We delay the arguments in `hif` so they are not applied
+until forced.
+-}
+pif :: forall {s :: S} {b :: PType}. Term s (PBool :--> (b :--> (b :--> b)))
+pif = plam hif
```
TotallyNotChase, line 450: @peter-mlabs  One thing that I didn't notice but might still be helpful for your understanding. This `pif` has the same issues as the previous one we talked about. Its type is `Term s (PBool :--> b :--> b :--> b)`. All of these types are immediately evaluated before the body of `hif` is even entered - so we're back to square one. No use delaying inside the body.

You can't make lazy Plutarch level functions ergonomically, at all. And there's no point in doing so. A lazy Plutarch level function would have a signature like: `Term s (PBool :--> PDelay b :--> PDelay b :--> b)`

peter-mlabs: Yeah, after spending some time working with it, this is all starting to sink in. Thanks for the help and feedback!

PR #314, opened 2022-02-21
Update plutus and haskell-language-server

L-as:
This was a bit of a PITA, and required a lot of changes, some backward incompatible.

### DIFF
--- a/Plutarch/Lift.hs
+++ b/Plutarch/Lift.hs
@@ -15,6 +15,7 @@ module Plutarch.Lift (
   PLift,
   DerivePConstantDirect (..),
   DerivePConstantViaNewtype (..),
+  DerivePConstantViaBuiltin (..),
 
   -- * Internal use
   PUnsafeLiftDecl (..),
@@ -23,12 +24,14 @@ module Plutarch.Lift (
 import Data.Coerce
 import Data.Kind (Type)
 import GHC.Stack (HasCallStack)
-import Plutarch.Evaluate (evaluateScript)
+import Plutarch.Evaluate (EvalError, evalScript)
 import Plutarch.Internal (ClosedTerm, PType, Term, compile, punsafeConstantInternal)
 import qualified Plutus.V1.Ledger.Scripts as Scripts
 import qualified PlutusCore as PLC
-import PlutusCore.Constant (readKnownConstant)
+import PlutusCore.Builtin (ReadKnownError, readKnownConstant)
 import PlutusCore.Evaluation.Machine.Exception (ErrorWithCause, MachineError)
+import PlutusTx (BuiltinData, Data, builtinDataToData, dataToBuiltinData)
+import PlutusTx.Builtins.Class (FromBuiltin, ToBuiltin, fromBuiltin, toBuiltin)
 import qualified UntypedPlutusCore as UPLC
 
 class (PConstant (PLifted p), PConstanted (PLifted p) ~ p) => PUnsafeLiftDecl (p :: PType) where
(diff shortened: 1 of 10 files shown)


plutarch-test/plutarch-test.cabal
```diff
@@ -96,11 +96,18 @@ library
   hs-source-dirs:  src
   exposed-modules: Plutarch.Test
 
+  -- FIXME remove -Wwarn=deprecations
+  ghc-options:     -Wwarn=deprecations
```
TotallyNotChase, line 100: Should we have an issue for this perhaps? Using deprecated functions in examples that users might look at is probably not the best look.

plutarch-test/src/Plutarch/ScriptsSpec.hs
```diff
@@ -20,9 +20,14 @@ module Plutarch.ScriptsSpec (
 import Data.Text (Text)
 
 import qualified Plutus.V1.Ledger.Api as Plutus
-import qualified Plutus.V1.Ledger.Crypto as Plutus
 
-import Data.Aeson.Extras (encodeSerialise)
+import GHC.Prim (coerce)
```
TotallyNotChase, line 24: I was under the impression `coerce` can be obtained from `Data.Coerce` as well, which wouldn't require the `ghc-prim` dependency maybe. Could you explain the difference to me please?

TotallyNotChase (changes requested):
Could you please add the user facing changes on the changelog?

TotallyNotChase:
The `PConstant`/`PLift` derivation stuff will also need to be updated in the guide. I was curious why `DerivePConstantViaNewtype` was still used for `PPOSIXTime` though. Isn't `PInteger` a builtin as well? Or was it not affected by those `newtype -> data` changes upstream?

L-as:
> The `PConstant`/`PLift` derivation stuff will also need to be updated in the guide. I was curious why `DerivePConstantViaNewtype` was still used for `PPOSIXTime` though. Isn't `PInteger` a builtin as well? Or was it not affected by those `newtype -> data` changes upstream?

:shrug: Maybe they didn't do the change for that. I only fixed what broke.

plutarch-test/plutarch-test.cabal
L-as: Maybe. wrt. the examples, they will all be ported to the new testing framework anyway, and wrt. plutarch-test, that will be fixed in another PR as it's a lot of work.

Plutarch/Lift.hs
```diff
@@ -109,3 +113,29 @@ instance (PLift p, PLift p', Coercible h (PLifted p')) => PConstant (DerivePCons
   type PConstanted (DerivePConstantViaNewtype h p p') = p
   pconstantToRepr x = pconstantToRepr @(PLifted p') $ coerce x
   pconstantFromRepr x = coerce $ pconstantFromRepr @(PLifted p') x
+
+class ToBuiltin' a arep | a -> arep where
```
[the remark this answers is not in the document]
L-as: Yep, I have no idea why they don't have an instance for `Data`.

Plutarch/Lift.hs
```diff
@@ -64,30 +67,31 @@ pconstant x = punsafeConstantInternal $ PLC.someValue @(PConstantRepr (PLifted p
...
-  Left e -> Left $ LiftError_ScriptError e
+      Left e -> Left $ LiftError_ReadKnownError e
+  (Left e, _, _) -> Left $ LiftError_EvalError e
 
 -- | Like `plift'` but throws on failure.
 plift :: forall p. (HasCallStack, PLift p) => ClosedTerm p -> PLifted p
 plift prog = case plift' prog of
   Right x -> x
-  Left e -> error $ "plift failed: " <> show e
+  Left LiftError_FromRepr -> error "plift failed because of pconstantFromRepr"
```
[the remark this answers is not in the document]
L-as: Nope, it indicates that `pconstantFromRepr` returned `Nothing`, which is what will happen if you e.g. do `plift $ punsafeCoerce (pconstant ()) :: Term _ PInteger`

Plutarch/Lift.hs
```diff
@@ -64,30 +67,31 @@ pconstant x = punsafeConstantInternal $ PLC.someValue @(PConstantRepr (PLifted p
...
+  Left (LiftError_ReadKnownError _) -> error "plift failed because of an internal error in plutus-core"
```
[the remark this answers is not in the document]
L-as: Good question, no idea.

plutarch-test/src/Plutarch/ScriptsSpec.hs
L-as: You're right! I was thinking that too but couldn't remember which it was.

plutarch-test/plutarch-test.cabal
TotallyNotChase: is @srid working on this? or did you mean you're working on it in another PR? or does someone else need to work on it?

Plutarch/Lift.hs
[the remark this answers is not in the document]
TotallyNotChase: I guess stringifying `ReadKnownError` might only be possible with prisms
<https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore-Evaluation-Machine-Exception.html#v:_UnliftingError>
<https://playground.plutus.iohkdev.io/doc/haddock/plutus-core/html/PlutusCore-Evaluation-Machine-Exception.html#v:_UnliftingErrorE>

that's really unfortunate. Best we can do to make this more informative might be:
```hs
Left (LiftError_ReadKnownError (ErrorWithCause _ errTerm)) ->
  error $ "plift failed because of an internal error in plutus-core"
    <> "\n"
    <> maybe "" (\x -> "cause: " <> show x) errTerm
```

plutarch-test/plutarch-test.cabal
L-as: @srid is working on this

L-as:
https://github.com/input-output-hk/plutus/blob/71b1c4337dd97dd6b33daf89c4920f629efdbfbb/plutus-ledger-api/src/Plutus/V1/Ledger/Time.hs#L43

They just use a newtype of `Integer`

Plutarch/Lift.hs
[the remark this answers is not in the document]
L-as: #319

PR #316, opened 2022-02-22
Don't hoist forced built-ins

L-as:
Depends on #307


L-as:
See latest two commits

srid:
@L-as Some of these benchmarks end up being slightly worse (eg: [rational.bench.golden](https://github.com/Plutonomicon/plutarch/pull/316/files#diff-2288bb1ecdfbf1f52dacef50f1f5a6512204444ac89eea5e5fc936429e973ed3)), which tells me that we should be selective as to which builtin to inline?

L-as:
That's probably the double force inlining.

srid:
Looking at [api.ctx.bench.golden](https://github.com/Plutonomicon/plutarch/pull/316/files#diff-ceacda6adaf5735c4a8724f5b4c7d1b3b2ace24cac97bbb5ddd7106a1387e1b4) ... oddly, the use of inlined list/pair builtins improves cpu/mem benchmarks for `get.mint` and `get.sym`, while being worse off for `get.credentials`.

TotallyNotChase:
Huh, that behavior is really weird and we should definitely look into it further

L-as:
So I'm not sure of what to do here. The issue is that there's both the cost of the let-bindings, and the cost of the reference. Let-bindings seem to be very expensive. We probably want to focus making sure *bigger* scripts are smaller, rather than microoptimising the already small ones.

TotallyNotChase:
It *does* seem like a balancing act. In particular, for the case for `Trace` (and potentially other builtins that take one force) - 4 or more usages are better of hoisted. Less than 4, rather be inlined.

If we can inline hoisted builtins depending on this "saturation point" - we can have the best of both worlds. Is that something we can trivially do right here?

TotallyNotChase:
> So I'm not sure of what to do here. The issue is that there's both the cost of the let-bindings, and the cost of the reference. Let-bindings seem to be very expensive. We probably want to focus making sure _bigger_ scripts are smaller, rather than microoptimising the already small ones.

I'm struggling to understand why/how references are costlier. Could you explain it to me please? By references, you mean the lambda args I assume? Something like `(\\i0 -> \\i0 -> 1)` seems to only have a size of `3`. The constant is `1`, and it seems like each lambda is `1` as well. Meanwhile, `(\\i0 -> \\i0 -> i1)` is also giving me a size of `3` - what's going on here?

```hs
test :: Term s (PInteger :--> PInteger :--> PInteger)
test = plam $ \a b -> b

test' :: Term s (PInteger :--> PInteger :--> PInteger)
test' = plam $ \a b -> 1

-- > scriptSize $ compile test
-- 3
-- > scriptSize $ compile test'
-- 3
```

L-as:
@TotallyNotChase https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/data/cekMachineCosts.json#L3

L-as:
Can someone make a test with a ton of traces, and a ton of ifs?

TotallyNotChase:
> https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/data/cekMachineCosts.json#L3

They're all the same? references, constants, and the rest

TotallyNotChase:
> Can someone make a test with a ton of traces, and another with a ton of ifs?

For `ptrace`:
```hs
{- sample code
  ptraceHoisted/Inline # "foo" #$
  ... many more ...
  ptraceHoisted/Inline # "foo" #$ phexByteStr "41"
-}

-- 1: hoisted: 6
-- 1: inlined: 6
-- 2: hoisted: 16
-- 2: inlined: 14
-- 3: hoisted: 20
-- 3: inlined: 19
-- 4: hoisted: 24
-- 4: inlined: 24
-- .... hoisted wins in all cases afterwards ...
-- 57: hoisted: 236
-- 57: inlined: 289
-- ... hoisted keeps winning ...
-- 100: hoisted: 408
-- 100: inlined: 509
-- ... hoisted keeps winning ...
-- 1000: hoisted: 4012
-- 1000: inlined: 5009
```
it's pretty clear that 4 is the turning point for single force builtins

will also check for double force builtins

TotallyNotChase:
Here's also the `IfThenElse` case, which is also a single force builtin:-
```hs
{- sample code
  pifHoisted/Inline # pfalse # ptrue #$
  ... many more ...
  pfalse
-}

-- id: method: size
-- 1: hoisted: 8
-- 1: inlined: 8
-- 2: hoisted: 17
-- 2: inlined: 15
-- 3: hoisted: 23
-- 3: inlined: 22
-- 4: hoisted: 29
-- 4: inlined: 29
-- .... hoisted wins in all cases afterwards ...
-- 57: hoisted: 347
-- 57: inlined: 400
-- ... hoisted keeps winning ...
-- 100: hoisted: 605
-- 100: inlined: 701
-- ... hoisted keeps winning ...
-- 1000: hoisted: 6005
-- 1000: inlined: 7001
```

TotallyNotChase:
for `ChooseList` (takes two forces):

```hs
{- sample code
  pcatLHoisted/Inline # pnilInt # ptrue #$
  ... many more ...
  pfalse
-}

-- id: method: size
-- 1: hoisted: 9
-- 1: inlined: 9
-- 2: hoisted: 21
-- 2: inlined: 20
-- 3: hoisted: 27
-- 3: inlined: 28
-- 4: hoisted: 33
-- 4: inlined: 36
-- .... hoisted wins in all cases afterwards ...
-- 57: hoisted: 351
-- 57: inlined: 460
-- ... hoisted keeps winning ...
-- 100: hoisted: 609
-- 100: inlined: 804
-- ... hoisted keeps winning ...
-- 1000: hoisted: 6009
-- 1000: inlined: 8004
```
3 is the turning point.

t1lde:
> For `Trace`:

You should also measure cpu & memory here

srid:
https://github.com/Plutonomicon/plutarch/pull/316#issuecomment-1048758465

> 3 is the turning point.

You mean 2 is the turning point.

---

Looks like there is a pattern here: single-forced builtins are best inlined up to 4 uses. And double-forced builtins are best inlined up to 2 uses. Needs to be verified.

TotallyNotChase:
> CPU/Mem usage

You guys are right! On it!

TotallyNotChase:
CPU/Mem for double force:

```hs
{- sample code
  pcatLHoisted/Inlined # pnilInt # ptrue #$
  ... many more ...
  pfalse
-}
```

|  Id/Method   |    CPU    |    Memory    |    Size    |
| ------------ | --------- | ------------ | ---------- |
|  1/hoisted   |    418057 |      1032    |     9      |
|  1/inlined   |    418057 |      1032    |     9      |
|  2/hoisted   |    925333 |      2264    |     21     |
|  2/inlined   |    895560 |      2164    |     20     |
|  3/hoisted   |   1253971 |      2896    |     27     |
|  3/inlined   |   1283744 |      2996    |     28     |
|  4/hoisted   |   1582609 |      3528    |     33     |
|  4/inlined   |   1671928 |      3828    |     36     |
| 1000/hoisted | 328906057 |    633000    |   6009     |
| 1000/inlined | 388303192 |    832500    |   8004     |

pretty linear like last time - makes complete sense to me

TotallyNotChase:
CPU/Mem for single force (`IfThenElse`):

```hs
{- sample code
  pifHoisted/Inlined # pfalse # ptrue #$
  ... many more ...
  pfalse
-}
```

|  Id/Method   |    CPU    |    Memory    |    Size    |
| ------------ | --------- | ------------ | ---------- |
|  1/hoisted   |    238285 |       901    |     8      |
|  1/inlined   |    238285 |       901    |     8      |
|  2/hoisted   |    506243 |      1802    |     17     |
|  2/inlined   |    446697 |      1602    |     15     |
|  3/hoisted   |    684882 |      2403    |     23     |
|  3/inlined   |    655109 |      2303    |     22     |
|  4/hoisted   |    863521 |      3004    |     29     |
|  4/inlined   |    863521 |      3004    |     29     |
| 1000/hoisted | 178787965 |    601600    |   6005     |
| 1000/inlined | 208441873 |    701200    |   7001     |

L-as:
I will make it such that it is only inlined if there are less than 5 uses.

L-as:
Do note that  it's not always beneficial under 5 even to inline, because it might be that it's in a loop and is evaluated a million times.

TotallyNotChase:
> Do note that it's not always beneficial under 5 even to inline, because it might be that it's in a loop and is evaluated a million times.

That's kind of a big problem. Why inline at all then?

L-as:
It's a good question. The problem only applies to forced built-ins, because they're not actually in WHNF.

L-as:
The conclusion is that we need clever inlining that counts how many times it's used.

L-as:
Fixing flakiness of tests should not rely on this in any case, it would be a horrible hack.

L-as:
This will not be done, `pinline` and such will be available instead.

PR #318, opened 2022-02-22
Mario/strict nullary constructors

blamario:
Updated the tests and benchmarks, have a look.


Plutarch/Internal/PlutusType.hs
```diff
@@ -140,9 +141,13 @@ class PCon a where
   pcon :: a s -> Term s a
 
 class PMatch a where
-  -- | Pattern match over Plutarch Terms via a Haskell datatype
+  -- | Pattern match over Plutarch Terms via a Haskell datatype. All branches are strictly evaluated, use `pmatchSum` if you need them lazy.
   pmatch :: Term s a -> (a s -> Term s b) -> Term s b
 
+-- | A less efficient delayed 'pmatch' for use with partial pattern matches in @do@ notation.
+pmatchSum :: PlutusType a => Term s a -> (a s -> Term s b) -> Term s b
+pmatchSum x f = PI.pforce (pmatch x (PI.pdelay . f))
```
L-as, line 149: The issue is that it's easy to accidentally use `pmatch` and have a non-obvious bug.

TotallyNotChase:
I thought it's more common for people to expect short circuiting in their "pattern matches" though. I remember doing that a lot in our validators. Is the opposite more common for others?

L-as:
I agree with Chase. Maybe there's another viable approach?

blamario:
I don't see any way around the problem with the `PlutusType` class as it is. And I have no suggestion on how to change it either. This PR is a breaking change, obviously, so this is a good time to consider it.

blamario:
Can we have a decision here, one way or another? It kinda affects how I proceed with the FFI branch.

TotallyNotChase:
> Can we have a decision here, one way or another? It kinda affects how I proceed with the FFI branch.

yes please. I also want to try and test it out on some of my existing validators to get some idea about the difference.

L-as:
Why does FFI depend on this?

blamario:
It doesn't strictly depend on it, it just becomes easier. Without this PR, I'll have to add something like `data PStrictMaybe` with a Scott encoding that matches PlutusTx's.

L-as:
I don't think we'll ever change the semantics of `pmatch` to strict.

blamario:
Ok, it was just an experiment.

PR #320, opened 2022-02-22
Capture more information on `plift` `ReadKnownError`

### LINKED ISSUE #319: More informative errors on `plift`
The constructors for `ReadKnownError` aren't exported, but some prisms are. https://github.com/input-output-hk/plutus/blob/0397f83b8298ba1b3c531511172637651b624164/plutus-core/plutus-core/src/PlutusCore/Builtin/KnownType.hs#L158

TotallyNotChase:
quick fix since I had the code lying around from reviewing #314 

Resolves #319

### DIFF
--- a/Plutarch/Lift.hs
+++ b/Plutarch/Lift.hs
@@ -21,6 +21,7 @@ module Plutarch.Lift (
   PUnsafeLiftDecl (..),
 ) where
 
+import Control.Lens ((^?))
 import Data.Coerce
 import Data.Kind (Type)
 import GHC.Stack (HasCallStack)
@@ -29,7 +30,7 @@ import Plutarch.Internal (ClosedTerm, PType, Term, compile, punsafeConstantInter
 import qualified Plutus.V1.Ledger.Scripts as Scripts
 import qualified PlutusCore as PLC
 import PlutusCore.Builtin (ReadKnownError, readKnownConstant)
-import PlutusCore.Evaluation.Machine.Exception (ErrorWithCause, MachineError)
+import PlutusCore.Evaluation.Machine.Exception (ErrorWithCause (ErrorWithCause), MachineError, _UnliftingErrorE)
 import PlutusTx (BuiltinData, Data, builtinDataToData, dataToBuiltinData)
 import PlutusTx.Builtins.Class (FromBuiltin, ToBuiltin, fromBuiltin, toBuiltin)
 import qualified UntypedPlutusCore as UPLC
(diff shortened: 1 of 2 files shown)


Plutarch/Lift.hs
```diff
@@ -89,9 +90,15 @@ plift' prog = case evalScript (compile prog) of
...
   Right x -> x
-  Left LiftError_FromRepr -> error "plift failed because of pconstantFromRepr"
-  Left (LiftError_ReadKnownError _) -> error "plift failed because of an internal error in plutus-core"
-  Left (LiftError_EvalError e) -> error $ "plift failed because of an erring term: " <> show e
+  Left LiftError_FromRepr -> error "plift failed: pconstantFromRepr returned 'Nothing'"
+  Left (LiftError_ReadKnownError (ErrorWithCause e causeMaybe)) ->
+    let unliftErrMaybe = e ^? _UnliftingErrorE
+     in error $
+          "plift failed: internal plutus-core error: "
+            <> maybe "ReadKnownEvaluationFailure" show unliftErrMaybe
```
TotallyNotChase, line 98: bit of an assumption here, `ReadKnownError` (the type of `e`) has 2 unexported constructors. One of them corresponds to `AsUnliftingError` (i.e `_UnliftingErrorE` prism) and the other corresponds to `AsEvaluationFailure` (not to be confused with `AsEvaluationError`). The latter is only useful to *confirm* that the constructor is indeed `ReadKnownEvaluationFailure` - as it holds no fields, and no real information.

Thus, if `unliftErrMaybe` is `Nothing`, the only possibility for `e` is to be `ReadKnownEvaluationFailure` - which is what I'm defaulting to here, without ever invoking `_EvaluationFailure :: Prism' ReadKnownError ()`.

With the lack of pattern matching, exhaustion checks are no longer simple and if they add another constructor - this kinda flies under our nose, potentially. Is there a way to achieve such exhaustion safety with prisms?

TotallyNotChase:
What's up with that "error: input 'haskell-nix/nixpkgs' follows a non-existent input 'nixpkgs-unstable'"?

L-as:
Update your version of Nix to latest master

TotallyNotChase:
> Update your version of Nix to latest master

The error is in our CI >:)

srid:
> > Update your version of Nix to latest master
> 
> The error is in our CI >:)

It is coming from benchmark-diff effect. It is not worth addressing, as the whole thing is going away when #322 is finished today.

Plutarch/Lift.hs
L-as: `ReadKnownEvaluationFailure` should never happen FWIW because we do `evalScript` AFAICT

TotallyNotChase: Interesting, should the `Nothing` case use a different string than `ReadKnownEvaluationFailure` perhaps (to indicate that case should be absurd)?

Still doesn't solve the constructor exhaustion issue, but I don't think we can actually solve that. Might just have to put a note here.

L-as: Rather than `"internal plutus-core error"`, it should be something like `"incorrect type"`.

TotallyNotChase: Sure, changed.

L-as (approved):
LGTM

PR #321, opened 2022-02-22
Use `evalScript` to fill in benchmarks for failing programs

srid:


srid:
CI passed, except for benchmark-diff failing. I should get rid of this diff thingy now...

L-as:
This is technically correct, but you're not making full use of this. With this, we always have access to the used budget and can thus always include it.

srid:
> This is technically correct, but you're not making full use of this. With this, we always have access to the used budget and can thus always include it.

I see what you mean ... okay done.

plutarch-benchmark/src/Plutarch/Benchmark.hs
```diff
@@ -65,42 +66,46 @@ import qualified Plutus.V1.Ledger.Api as Plutus
...
+mkBenchmark :: ExBudget -> ScriptSizeBytes -> Benchmark
+mkBenchmark (ExBudget cpu mem) = Benchmark cpu mem
+
 newtype ScriptSizeBytes = ScriptSizeBytes Int64
   deriving stock (Eq, Ord, Show, Generic)
   deriving newtype (Num, ToField)
   deriving newtype (ToJSON)
 
+scriptSize :: Script -> ScriptSizeBytes
+scriptSize = ScriptSizeBytes . fromInteger . toInteger . SBS.length . serialiseScriptShort
```
L-as, line 104: `fromInteger . toInteger`? `fromIntegral`?

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -168,17 +170,21 @@ currentGoldenKey = do
         Just path ->
           pure $ sconcat $ fmap GoldenKey path
 
-{- | Like `evaluateScript` but doesn't fail. Also returns `Script`.
+{- | Like `evalScript` but doesn't fail, and returns `Benchmark`.
```
L-as, line 173: Comment seems outdated.

plutarch-benchmark/src/Plutarch/Benchmark.hs
```diff
@@ -65,42 +66,46 @@ import qualified Plutus.V1.Ledger.Api as Plutus
...
+
+serialiseScriptShort :: Script -> SBS.ShortByteString
+serialiseScriptShort = SBS.toShort . LB.toStrict . serialise -- Using `flat` here breaks `evalScriptCounting`
```
L-as, line 107: Can you elaborate on the comment BTW?

L-as (changes requested):
Looks mostly good to me

plutarch-test/src/Plutarch/Test/Golden.hs
srid: It is accurate?

plutarch-benchmark/src/Plutarch/Benchmark.hs
srid: This looks to be an outdated comment; I remember trying to use the `flat` library for some reason, but it is not used anymore. Removed.

plutarch-test/src/Plutarch/Test/Golden.hs
L-as: `evalScript` doesn't fail either

srid: By "faIl" I meant it returning `Left EvalError`; I've clarified the comment.

L-as (approved):
LGTM

PR #322, opened 2022-02-22
WIP: Port and remove benchmark-diff

srid:


srid:
Thanks for [golden testing](https://kseo.github.io/posts/2016-12-15-golden-tests-are-tasty.html), we are able to delete a bunch of Nix code.

plutarch-test/goldens/plutustype.deconstr.bench.golden
```diff
@@ -0,0 +1,26 @@
...
+matching.raw.sumtype(ignore-fields) {"exBudgetCPU":1122748,"exBudgetMemory":2466,"scriptSizeBytes":33}
+matching.raw.sumtype(partial-match) {"exBudgetCPU":1481159,"exBudgetMemory":3198,"scriptSizeBytes":39}
+matching.raw.sumtype(exhaustive).minting {"exBudgetCPU":2739406,"exBudgetMemory":6502,"scriptSizeBytes":83}
+matching.raw.sumtype(exhaustive).spending {"exBudgetCPU":1749116,"exBudgetMemory":4098,"scriptSizeBytes":93}
+matching.raw.sumtype(exhaustive).rewarding {"exBudgetCPU":2244261,"exBudgetMemory":5300,"scriptSizeBytes":90}
+matching.raw.sumtype(exhaustive).certifying {"exBudgetCPU":2739406,"exBudgetMemory":6502,"scriptSizeBytes":85}
+matching.raw.sumtype(exhaustive)(ignore-fields).minting {"exBudgetCPU":2291676,"exBudgetMemory":5470,"scriptSizeBytes":75}
+matching.raw.sumtype(exhaustive)(ignore-fields).spending {"exBudgetCPU":1301386,"exBudgetMemory":3066,"scriptSizeBytes":85}
+matching.raw.sumtype(exhaustive)(ignore-fields).rewarding {"exBudgetCPU":1796531,"exBudgetMemory":4268,"scriptSizeBytes":82}
+matching.raw.sumtype(exhaustive)(ignore-fields).certifying {"exBudgetCPU":2291676,"exBudgetMemory":5470,"scriptSizeBytes":77}
```
srid, line 22: @TotallyNotChase Some of these don't match between typed and raw; could you review?

srid:
@Geometer1729 FYI I found myself needing `PlutarchGoldens`, so I took the opportunity to also export it from `Plutarch.Test`.

L-as:
Doesn't it make more sense to remove `plutarch-benchmark` entirely and move the needed (tiny amount of) code into plutarch-test? I don't think anyone needs the CSV stuff anymore.

Geometer1729:
> @Geometer1729 FYI I found myself needing `PlutarchGoldens`, so I took the opportunity to also export it from `Plutarch.Test`.
Should I close https://github.com/Plutonomicon/plutarch/pull/323 as redundant then?

srid:
> > @Geometer1729 FYI I found myself needing `PlutarchGoldens`, so I took the opportunity to also export it from `Plutarch.Test`.
> 
> Should I close #323 as redundant then?

Depends on which gets merged first.

srid:
> Doesn't it make more sense to remove `plutarch-benchmark` entirely and move the needed (tiny amount of) code into plutarch-test? 

Ideally, yes. Do we care about backwards compat here? @TotallyNotChase 

> I don't think anyone needs the CSV stuff anymore.

Right, I'll remove that.

L-as:
well this is for 1.2 anyway

TotallyNotChase:
> Ideally, yes. Do we care about backwards compat here? @TotallyNotChase

I guess since none of my teammates have actually started using `plutarch-benchmark`, it's ok. I assume no one here knows anybody that's actually using this library? Otherwise they might notice an entire library vanishing from under their feet.

plutarch-test/goldens/plutustype.deconstr.bench.golden
TotallyNotChase: Yeah all of these numbers are exactly the same as before - no worries here. We do have some potential inefficiencies in a few places of the `pmatchRepr` machinery that we cannot fix without linear types.

plutarch-test/goldens/plutustype.deconstr.uplc.eval.golden
```diff
@@ -0,0 +1,27 @@
...
+matching.raw.sumtype(exhaustive).rewarding (program 1.0.0 #03)
+matching.raw.sumtype(exhaustive).certifying (program 1.0.0 #04)
+matching.raw.sumtype(exhaustive)(ignore-fields).minting (program 1.0.0 #01)
+matching.raw.sumtype(exhaustive)(ignore-fields).spending (program 1.0.0 #02)
+matching.raw.sumtype(exhaustive)(ignore-fields).rewarding (program 1.0.0 #03)
+matching.raw.sumtype(exhaustive)(ignore-fields).certifying (program 1.0.0 #04)
+fields.typed.extract-single (program 1.0.0 #d87a9f41abff)
+fields.raw.extract-single (program 1.0.0 #d87a9f41abff)
+combined.typed.toValidatorHash (program 1.0.0 (\i0 -> \i0 -> i2 (unBData (force headList [#41ab]))))
+combined.raw.toValidatorHash (program 1.0.0 (\i0 -> \i0 -> i2 (unBData (force headList (force (force sndPair) ( 1
```
TotallyNotChase, line 26: that's interesting, why are these different?

TotallyNotChase (approved):

TotallyNotChase:
btw it'd be nice for some of these tests to measure the unapplied lambda as well as the full application (especially the deconstr tests). Probably something to consider later.

plutarch-test/goldens/plutustype.deconstr.uplc.eval.golden
TotallyNotChase: oh boy, it's because the `pmatchRepr` machinery does an extra `plet` to explicitly evaluate the `SndPair` before applying the `pcon . PJust`. But that sort of means the raw version is cheating by doing less work! That's non obvious and I'm not sure whether the raw version should be intentionally crippled here to match the result of the typed one.

t1lde:
I have been making use of `plutarch-benchmark` as a library - mostly out of convenience. IMO it's fine to be removed

srid:
This PR is ready; there is nothing left to do AFAICT.

L-as (approved):
LGTM

PR #326, opened 2022-02-23
Issue 303 and 344

### LINKED ISSUE #303: Add something like `PAsDataMaybe`

MangoIV:
## What is this 
This will resolve #303 and #344

## Tasks 

- [x] implementation for basic datatypes
- [x] implementation for `PDataRecord`
- [x] implementation for `PDataSum`
- [x] instances for `PValue` and `PMap`
- [x] tests for `PValue` (how to mock PValue)
- [x] return explicit continuation instead of `TermCont` by request of @L-as
- [x] make it possible to track *all* the types that have been recovered 
- [x] implementation for `PIsDataRepr`
- [x] access all types that have been recovered via Record update

## Issues
- I have some issues concerning the stuff I wrote until now. I want some confidence in what I did before I continue, that's why I opened this draft. 
- Currently I'm struggling with tree indexing, I wonder whether it is possible to polymorphically return either a tree or a PType, I have had my issues with that. I'll probably have to rewrite everything that exists for `Term` now to also work with `HSTree`. 

## Persisting and new issues after this has been merged
- add infallible conversion `PFrom` and conversion that returns Maybe `PMaybeFrom` 
```haskell
class PMaybeFrom (a :: PType) (b :: PType) where
  type PMaybeFromExcess a b :: PType
  pmaybeFrom :: Term s a -> (((Term s (PMaybe b), PMaybe (PMaybeFromExcess a b) s) -> Term s r) -> Term s r)
```


TotallyNotChase:
general thought: the conversion functions should probably be Haskell level functions - as I expect certain conversions to be no-ops.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,99 @@
...
+import qualified PlutusCore as PLC
+
+{- |
+    Each POpaque can be of any representation as it represents
+    the universe of Types. With PTryFrom we establish trust
+    in the Data by verifying the requested type matches its
+    representation.
+-}
+class PTryFrom (a :: PType) where
+  ptryFrom :: Term s (PData :--> a)
```
t1lde, line 45: This should be a multiparam class IMO,
and we can have `ptryFromData` as a synonym for convenience

Plutarch/Verify.hs
```diff
@@ -0,0 +1,99 @@
...
+
+instance PTryFromRecur PByteString where
+  ptryFromRecur = punsafeBuiltin PLC.UnBData
+
+instance (PTryFromRecur a, PLift a, PIsData a) => PTryFromRecur (PBuiltinList (PAsData a)) where
+  ptryFromRecur = phoistAcyclic $
+    plam $ \opq ->
+      let lst :: Term _ (PBuiltinList (PAsData PData))
+          lst = punsafeBuiltin PLC.UnListData #$ opq
+       in pmap # (plam $ \e -> pdata $ ptryFromRecur @a #$ pfromData e) # lst
```
t1lde, line 88: We want to avoid encouraging users to use this by default, perhaps comments + the name are enough :thinking:

L-as:
We want some examples of how to use this, probably something like what I described in https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109

Plutarch/Verify.hs
MangoIV: I will have to think about what the application of it being a MultiParamTypeClass would be (As normally you don't want to have Types that are unverified anyway and `PData` / `POpaque` are the only types that directly occur to me that this could be needed for.)

MangoIV:
> we want some examples of how to use this, probably something like what I described in https://github.com/Plutonomicon/plutarch/pull/267#discussion_r804499109

> Could add this to plutarch-test as VerifySpec.hs. More generally, all the (new) examples should go in plutarch-test - and we get tests and goldens for free.

I already started doing that as I need to test my changes anyway.

L-as:
We need examples of how to use this before we settle on the design

plutarch-test/plutarch-test.cabal
```diff
@@ -66,7 +66,7 @@ common c
 
   ghc-options:
     -Wall -Wcompat -Wincomplete-uni-patterns -Wno-unused-do-bind
-    -Wno-partial-type-signatures -Wmissing-export-lists -Werror
+    -Wno-partial-type-signatures -Wmissing-export-lists
```
MangoIV, line 69: happens all the time, will fix :eyes:

Plutarch/Verify.hs
TotallyNotChase: this is also safe for `PlutusType a => Term s a -> Term s (PInner a)`. We already have a function for this (`pto`) but it might be worth including in the conversion typeclasses.

Going further, `Term s (PInner a) -> Term s a` is also safe in certain cases but this may be harder to accomodate for. The conversion itself is given by `punsafeFrom`

Plutarch/Verify.hs
```diff
@@ -0,0 +1,107 @@
...
+    Each POpaque can be of any representation as it represents
+    the universe of Types. With PTryFrom we establish trust
+    in the Data by verifying the requested type matches its
+    representation.
+-}
+class PTryFrom (a :: PType) where
+  ptryFrom :: Term s (PData :--> a)
+
+instance PTryFrom PInteger where
+  ptryFrom = punsafeBuiltin PLC.UnIData
```
TotallyNotChase, line 49: you should be able to use `pasInt`, `pasList`, and such from `Plutarch.Builtin` in these places

MangoIV: Did that, thanks. :)

MangoIV:
I think I have a state which can be looked up to decide on how we further proceed. @L-as and @t1lde , would you be so kind and have a look at what I did and tell me what your opinion is? 
I'm personally much in favor with the `PTryFrom` class as it behaves in a lawful way, although I do get the concern, that it wraps and unwraps more. However, I think it can certainly offer the better user experience, removing `pfromData . pdata` and vice versa is possibly a task for another issue. I can repeat my concerns with the instances for `PTryFromUnwrap` in case T1lde wants me to. (Las already has them.)

Plutarch/DataRepr/Internal.hs
```diff
@@ -174,6 +175,21 @@ type family PLabelIndex (name :: Symbol) (as :: [PLabeledType]) :: Nat where
...
+
+instance
+  {-# OVERLAPPABLE #-}
+  ( All PIsData bs
+  , bs ~ PUnLabelAll xs
+  ) =>
+  PIsData (PDataRecord xs)
+  where
+  pfromData x = punsafeBuiltin PLC.UnListData # pforgetData x
+  pdata x = punsafeBuiltin PLC.ListData # x
```
t1lde, line 191: I don't think this is correct, `PDataRecord` is not a `List` but a `Constr`, and this would be inconsistent with the existing `PIsDataRepr` instances, where `pfromData` is a no-op.

L-as: `PDataRecord` is actually a `List`. `PDataSum` is `Constr`.

L-as: See `PlutusType` instance for `PDataRecord`

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+    If you don't care about parts of the datastructure
+    don't verify those parts, just let it return a PData
+    instead
+    Be aware this might get really expensive, so only
+    use it if you cannot establish trust otherwise
+    (e.g. via only checking a part of your Data with
+    PTryFrom)
+-}
+class PTryFrom (a :: PType) (b :: PType) where
+  --     "safest" type --^            ^-- target type
```
L-as, line 87: safest?

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  type PExcess a b :: PType
+  ptryFrom :: Term s a -> TermCont s (Term s b, Term s (PExcess a b))
```
L-as, line 89: do note that we haven't switched to `TermCont` project-wide yet. See #262 . I recommend you switch this to a manual continuation (by expanding the definition of `TermCont`), and switching to `TermCont` everywhere in another PR.

L-as: OTOH this function will be very misleading without `TermCont`. Hard choice. Anyway, doing the project-wide switch should be in another PR.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+    This checks the datastructure for validity.
+    If you don't care about parts of the datastructure
+    don't verify those parts, just let it return a PData
+    instead
+    Be aware this might get really expensive, so only
+    use it if you cannot establish trust otherwise
+    (e.g. via only checking a part of your Data with
+    PTryFrom)
+-}
+class PTryFrom (a :: PType) (b :: PType) where
```
L-as, line 86: We should try to note the laws for the type class.
`fst <$> ptryFrom x ≡ pure $ punsafeCoerce x` if `x` is indeed also in `b`.

Plutarch/Verify.hs
t1lde, line 89: I'm not clear what `PExcess` is for, if I'm understanding right, it's so you can choose between checking vs unwrapping the structure of `PData`? Is the first item returned always `punsafeCoerce`? 
Something feels a bit awkward about this. 
I think it might make more sense to make the `PExcess` type more primary (as the second param of the typeclass)

L-as, line 89: You should IMO return `PExcess a b s` instead of `Term s (PExcess a b)`. Otherwise you can't return e.g. a pair of terms.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  --     "safest" type --^            ^-- target type
+  type PExcess a b :: PType
```
L-as, line 88: I think the naming should be `PTryFromExcess`, otherwise it's unclear IMO.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  ptryFrom opq = pure $ (opq, pcon PUnit)
+
+instance {-# OVERLAPPING #-} PTryFrom PData (PAsData (PBuiltinList PData)) where
+  type PExcess PData (PAsData (PBuiltinList PData)) = PBuiltinList PData
+  ptryFrom opq = do
+    ver <- tcont $ plet (pasList # opq)
+    pure $ (punsafeCoerce opq, ver)
+
+instance
+  {-# OVERLAPS #-}
```
L-as, line 116: Please use `OVERLAPPING` or `OVERLAPPABLE`, no `OVERLAPS`.

Plutarch/Verify.hs
t1lde: A data type rather than a tuple might make things clearer, too. Something like:
```haskell
data CheckedAs a b s = CheckedAs { outer :: Term s a, inner :: Term s b } 
```
(naming is hard, but I think `outer` / `inner` is pretty good)

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+    -- I've not obtained a `PBuiltinList (PBuiltinPair a b)`
+    let verifyPair :: Term _ (PBuiltinPair PData PData :--> PBuiltinPair (PAsData a) (PAsData b))
+        verifyPair = plam $ \tup -> unTermCont $ do
+          (verfst, _) <- ptryFrom @PData @(PAsData a) $ pfstBuiltin # tup
+          (versnd, _) <- ptryFrom @PData @(PAsData b) $ psndBuiltin # tup
+          pure $ ppairDataBuiltin # verfst # versnd
+    ver <- tcont $ plet $ pmap # verifyPair # verMap
+    pure (punsafeCoerce opq, ver)
+
+instance {-# OVERLAPPING #-} PTryFrom PData (PAsData (PBuiltinMap POpaque POpaque)) where
```
L-as, line 134: This should be `PBuiltinMap PData PData` rather than `PBuiltinMap POpaque POpaque`.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+{-# OPTIONS_GHC -Wredundant-constraints #-}
+
+module Plutarch.Verify (
```
L-as, line 6: The module should probably be called `Plutarch.TryFrom`.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
```
L-as, line 3: You should remove this pragma.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  ptail,
+ )
+
+import Plutarch.Unsafe (punsafeBuiltin, punsafeCoerce)
+import qualified PlutusCore as PLC
+
+import Plutarch.TermCont (TermCont, tcont, unTermCont)
+
+{- |
+    Note: PAsData POpaque ~ PData
```
L-as, line 71: Do you make use of this fact anywhere? `PAsData PData ~ PData` seems more relevant.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+-}
+class PTryFrom (a :: PType) (b :: PType) where
+  --     "safest" type --^            ^-- target type
+  type PExcess a b :: PType
+  ptryFrom :: Term s a -> TermCont s (Term s b, Term s (PExcess a b))
+
+----------------------- PData instances -------------------------------------------------
+
+instance PTryFrom PData (PAsData PInteger) where
+  type PExcess PData (PAsData PInteger) = PInteger
```
t1lde, line 94: Could this be generalized:

```haskell
instance (PIsData p) => PTryFrom PData (PAsData p) where
  type PExcess PData (PAsData p) = p
  ptryFrom d = do
    inner <- tcont $ plet $  pfromData d
    pure (punsafeCoerce d, inner)
    
```

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+
+instance PTryFrom PData (PAsData PByteString) where
+  type PExcess PData (PAsData PByteString) = PByteString
+  ptryFrom opq = do
+    ver <- tcont $ plet (pasByteStr # opq)
+    pure $ (punsafeCoerce opq, ver)
+
+instance PTryFrom PData PData where
+  type PExcess PData PData = PUnit
+  ptryFrom opq = pure $ (opq, pcon PUnit)
```
t1lde, line 107: this feels strange

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+{-# OPTIONS_GHC -Wredundant-constraints #-}
+
+module Plutarch.Verify (
+  PTryFrom (PExcess, ptryFrom),
+  pcheckType,
+  pcheckByteStr,
+  pcheckInt,
+  pcheckList,
+  pcheckMap,
```
L-as, line 12: Do we need to export these helper functions? Seem like they'd be used only rarely.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+    ver <- tcont $ plet (pasList # opq)
+    pure $ (punsafeCoerce opq, ver)
+
+instance
+  {-# OVERLAPS #-}
+  ( PTryFrom PData (PAsData a)
+  , PTryFrom PData (PAsData b)
+  ) =>
+  PTryFrom PData (PAsData (PBuiltinMap a b))
+  where
```
t1lde, line 121: This instance is something to be careful of IMO - in many cases the efficient approach would be instead of verifying the whole structure 'outside in' by checking the whole structure, only to use parts of it, checking 'inside out' by first breaking down the structure then checking the relevant parts.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+instance
+  {-# OVERLAPPABLE #-}
+  ( PTryFrom PData a
+  , a ~ PAsData a'
+  , PIsData a'
+  , PTryFrom PData b
+  , b ~ PAsData b'
+  , PIsData b'
+  ) =>
+  PTryFrom PData (PAsData (PBuiltinPair a b))
```
TotallyNotChase, line 170: it's better to have a verifier for `PTuple` rather than `PAsData (PBuiltinPair ... )` - they are both the same and no op conversions exist. The former is certainly easier to work with without overlapping instances and stuff though.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  , PIsData a
+  ) =>
+  PTryFrom POpaque a
+  where
+  type PExcess POpaque a = PAsData a
+  ptryFrom opq = do
+    let prop :: Term _ a
+        prop = punsafeCoerce opq
+    ver' <- fst <$> (ptryFrom @PData @(PAsData a) $ pforgetData $ pdata prop)
+    ver <- tcont $ plet ver'
```
L-as, line 253: NB: I don't think you need the `plet` here technically, albeit it should be optimised out anyway, since `plet x` where `x` is an `RVar _` inlines the binding on use sites. Because of this you should just let this be as it is now.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  where
+  type PExcess PData (PAsData (PBuiltinMap a b)) = PBuiltinList (PBuiltinPair (PAsData a) (PAsData b))
+  ptryFrom opq = do
+    verMap <- tcont $ plet (pasMap # opq)
+    -- I've not obtained a `PBuiltinList (PBuiltinPair a b)`
+    let verifyPair :: Term _ (PBuiltinPair PData PData :--> PBuiltinPair (PAsData a) (PAsData b))
+        verifyPair = plam $ \tup -> unTermCont $ do
+          (verfst, _) <- ptryFrom @PData @(PAsData a) $ pfstBuiltin # tup
+          (versnd, _) <- ptryFrom @PData @(PAsData b) $ psndBuiltin # tup
+          pure $ ppairDataBuiltin # verfst # versnd
```
t1lde, line 130: I guess, it might be convenient to extract this out as 

```haskell
instance PTryFrom (PBuiltinPair PData PData) (PBuiltinPair (PAsData a) (PAsData b))
```

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+  ( PTryFrom PData (PAsData a)
+  , PIsData a
+  ) =>
+  PTryFrom POpaque a
+  where
+  type PExcess POpaque a = PAsData a
+  ptryFrom opq = do
+    let prop :: Term _ a
+        prop = punsafeCoerce opq
+    ver' <- fst <$> (ptryFrom @PData @(PAsData a) $ pforgetData $ pdata prop)
```
L-as, line 252: This is nice.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+instance {-# OVERLAPPING #-} PTryFrom PData (PAsData (PBuiltinList PData)) where
+  type PExcess PData (PAsData (PBuiltinList PData)) = PBuiltinList PData
+  ptryFrom opq = do
+    ver <- tcont $ plet (pasList # opq)
+    pure $ (punsafeCoerce opq, ver)
+
+instance
+  {-# OVERLAPS #-}
+  ( PTryFrom PData (PAsData a)
+  , PTryFrom PData (PAsData b)
```
t1lde, line 118: Maybe these can be generalized as `PTryFrom PData a, PTryFrom PData b`?

Plutarch/Verify.hs
t1lde, line 134: I think some generalization as I suggested might make this instance (and its overlapping) unneeded

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+    ver' <- fst <$> (ptryFrom @PData @(PAsData a) $ pforgetData $ pdata prop)
+    ver <- tcont $ plet ver'
+    pure $ (punsafeCoerce opq, ver)
+
+----------------------- Helper functions ------------------------------------------------
+
+pchooseData :: Term s (PData :--> a :--> a :--> a :--> a :--> a :--> a)
+pchooseData = phoistAcyclic $ pforce $ punsafeBuiltin PLC.ChooseData
+
+pcheckType :: (Term s PInteger) -> Term _ (PData :--> PAsData b)
```
L-as, line 261: I think this is dead code if we don't export it.

L-as (changes requested):
Looks great in general!

Plutarch/Verify.hs
L-as: Nope, `pfromData` already assumes the input is of the correct type

Plutarch/Verify.hs
L-as: See my comment about the law. It's not about outer vs inner, it's about extra information that the checking process finds, but isn't actually relevant for the law.

L-as: The semantics differ per instance.

Plutarch/Verify.hs
L-as: you could do `PTryFrom a a`. Not sure we need this instance thoughh.

Plutarch/Verify.hs
L-as: i.e. fusing the verification and processing I suppose. We could reintroduce the whole deep-shallow schism. Not sure it's worth it.

Plutarch/Verify.hs
```diff
@@ -0,0 +1,277 @@
...
+pcheckType i = plam $ \d ->
+  let con :: Term _ PInteger
+      con = pchooseData # d # 0 # 1 # 2 # 3 # 4
+   in pif (con #== i) (punsafeCoerce d) perror
```
TotallyNotChase, line 265: this is essentially equivalent to the `UnConstrData`/`UnListData`/`UnMapData`/`UnIData`/`UnBData` functions since it `perror`s anyway. `ChooseData` is really only relevant when you _want_ a recovery path.

Plutarch/Verify.hs
t1lde: Ah that's right, it only coincidentally works for some of these instances

Plutarch/Verify.hs
t1lde: Yeah, exactly.

Plutarch/Verify.hs
MangoIV: @t1lde The main idea about it is that I prove that 
`PData ~ PAsData PSomething` holds, in the process I will, in many cases, unwrap the `PAsData PSomething`, to safe work, I use the associated type to safe the actual work done and then just `punsafeCoerce` the input instead of rewrapping it, because if we have the second element of the pair without crashes we can actually be sure that the verification succeeded.

Plutarch/Verify.hs
MangoIV: I can just remove it but I don't see the issue with it, maybe we'll need it as a base case for a structure that doesn't rely on the inner type being wrapped in `PAsData`? (Although I can't come up with a usecase, either :D )

Plutarch/Verify.hs
MangoIV: I don't know whether the shallow thing would be feasible in this case tho, because after verifying that the outer type is a `PBuiltinList`, you end up with a list of `PBuiltinPair` (*not* `PAsData BuiltinPair`) which you can't verify standalone (at least I don't know of a way to do it).

Plutarch/Verify.hs
MangoIV: I just marked it as `OVERLAPPABLE` because there will be another instance for `PBuiltinPair PData PData`, as well.

Plutarch/Verify.hs
MangoIV: copied

Plutarch/Verify.hs
MangoIV: I think so, as well, but @t1lde wanted something like `PTryFromMaybe` which would need to use those because otherwise the script would just fail upon using `pAsSomething`, so maybe, if we need that class that returns a maybe instead, we may as well need those?

Plutarch/Verify.hs
MangoIV: well yes, I mean, it's the safest type "at the start", I will just remove it if it's misleading

Plutarch/Verify.hs
MangoIV: copied

Plutarch/Verify.hs
MangoIV: Your decision, just tell me whether I should leave it or not... :)

MangoIV: copie

Plutarch/Verify.hs
MangoIV: copied, I thought there would be an instance that this is being overlapped by but that's not possible, I will change it to OVERLAPPABLE

Plutarch/Verify.hs
MangoIV: I don't know if you can but I think you for sure shouldn't as you can't prove it for any x or y but only for `PAsData x` or `PAsData y`.

Plutarch/Verify.hs
MangoIV: but that's not correct, is it? 
`PBuiltinMap a b` is a `PBuiltinList (PBuiltinPair (PAsData a) (PAsData b))` so neither `a` nor `b` are actually `PData` but rather `POpaque \not\in PData`?

Plutarch/Verify.hs
MangoIV: that's why I exported the functions for future use in a recovering version of this. Perhaps I should refactor it out to another module tho?

TotallyNotChase: You can leave them out of this PR for now. I suspect there's a better way but we don't need this right now.

Plutarch/Verify.hs
L-as: Do we really need `PTryFromMaybe`? I don't think we do, or at the very least, we shouldn't try to do it in this PR.

Plutarch/Verify.hs
L-as: Just remove `TermCont` for now and make it take a continuation explicitly.

L-as:
Can we add a functional dependency `b -> a`?

L-as:
We might want `data PHopefully a b s = PHopefully (Term s a)` where `phopefully :: Term s (PHopefully a b) -> Term s b ; phopefully = ptryFrom` for use in `PValidator` and similar, so users can easily specify what format they expect.

L-as:
If we have fundep `b -> a`, we can make `a` implicit using the same trick used for `TermCont`.

Plutarch/Verify.hs
L-as: `POpaque \not\in PData` doesn't make sense? Unsure what you mean by this. You want `PData` because you want `PAsData PData`.

Plutarch/Verify.hs
L-as: We're not proving that `PData ~ PAsData PSomething` holds, though I assume you're writing that as a lazy notation.
To make it entirely clear and to avoid confusion, we're proving that for an `x` where `x \in PData`, `x \in PAsData something`.

Plutarch/Verify.hs
MangoIV: Yeah, I'm pretty much of the same opinion, I don't really see the use case for not failing after a e.g. a redeemer has been dishonest.

MangoIV:
> We might want data PHopefully a b s = PHopefully (Term s a) where phopefully :: Term s (PHopefully a b) -> Term s b ; phopefully = ptryFrom for use in PValidator and similar, so users can easily specify what format they expect.

I wanted to have `PData` as the basecase, i.e. just stop as the user "stops" caring what is in the structure, I think this is quite intuitive. I will definitely add documentation for it though. I have thought about `PHopefully` and I think it's a bit worse readability wise. 

```haskell
-- e.g.: 
-- "I, the user, just want to know, whether my Data is of type list, I don't care about 
-- anything else."
ver opq = unTermCont $ snd <$> (ptryFrom @PData @(PAsData (PBuiltinList PData)) opq)
```

Plutarch/Verify.hs
MangoIV: copied.

MangoIV:
An instance like 
```haskell
( PTryFrom a b
, PTryFrom b c) => 
PTryFrom a c where 
ptryFrom x = (fst <$> ptryFrom @a @b) >>= ptryFrom @b @c 
``` 
Would be really nice, as it would allow to do things like this: 
`ptryFrom @PData @PMap` 
where the instance for that arises from the instance `PTryFrom PData (PAsData PBuiltinMap)` and `PTryFrom (PAsData PBuiltinMap) (PAsData PMap)`, however, this is not possible because the instance overlaps with *everything* and there is no such as "ranked overlapping" in GHC. If anyone knows a solution to this, it would be really helpful...

Plutarch/Verify.hs
MangoIV: `PTuple` should automatically be an instance of `PTryFrom` because it's just a type alias for a `PDataSum`.

Plutarch/Verify.hs
MangoIV: > Going further, Term s (PInner a) -> Term s a is also safe in certain cases but this may be harder to accomodate for. The conversion itself is given by punsafeFrom

I couldn't see any, but if there is a rule for when  `punsafeFrom` is safe that is implementable for `PTryFrom`, that would be very handy because we would get the `PFrom` instance for free, if not there will just be a `PFrom` instance that aliases to `pto`

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -56,3 +61,88 @@ instance
...
       y' <- Plutus.fromData y
       Just (x', y')
+
+----------------------- Smart constructor to to create onchain instances ----------------
+
+{- |
+    Smart constructor to enforce PMap invariants on the Haskell level for writing
+    onchain code
+-}
+pmkPMap ::
```
L-as, line 71: This function seems weird to me. You could just take an actual map from `containers` as input

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -56,3 +61,88 @@ instance
...
+  forall a b s.
+  ( PLifted (PConstanted a) ~ a
+  , PLifted (PConstanted b) ~ b
```
L-as, line 74: #354

Plutarch/DataRepr/Internal.hs
```diff
@@ -174,6 +175,10 @@ type family PLabelIndex (name :: Symbol) (as :: [PLabeledType]) :: Nat where
 type family PUnLabel (a :: PLabeledType) :: PType where
   PUnLabel (name ':= a) = a
 
+instance {-# OVERLAPPABLE #-} PIsData (PDataRecord xs) where
+  pfromData x = punsafeBuiltin PLC.UnListData # pforgetData x
```
L-as, line 179: Please don't use the built-ins directly, they have aliases elsewhere.

Plutarch/Maybe.hs
```diff
@@ -16,3 +24,9 @@ data PMaybe (a :: PType) (s :: S)
   | PNothing
   deriving stock (GHC.Generic)
   deriving anyclass (Generic, PlutusType, PEq)
+
+pfromMaybe :: Term s (PMaybe a :--> a)
```
L-as, line 28: `pfromJust`

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,400 @@
...
+        - all conversion are fallible, this happens if the representation doesn't match
+          the expected type.
+      - the operation `ptryFrom` proves equality between the less expressive `PType` `a` and
+        the more expressive `PType` `b`, hence the first element of the resulting Tuple
+        must always be wrapped in `PAsData` if the origin type was `PData` (see law 1)
+      - the result type `b` must always be safe than the origin type `a`, i.e. it must carry
+        more information
+-}
+class PTryFrom (a :: PType) (b :: PType) where
+  type PTryFromExcess a b :: PType
```
L-as, line 99: I think the issue I have is that rather than returning this from `ptryFrom`, we really want to somehow splice in some code into the verification process, such that we're traversing while verifying essentially.

L-as (changes requested):
I'm still thinking about this

Plutarch/TryFrom.hs
L-as: You can emulate this for simple types by having `PTryFromExcess` be a list of variables referring to all the intermediately generated terms (this isn't trivially possible, but let's ignore the irrelevant difficulties for now). Maybe rather than having users do `ptryFrom` on complex types, we should somehow expose types that aren't verified yet. Let me think more about this.

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,404 @@
...
+  ptryFrom :: Term s a -> ((Term s b, Term s (PTryFromExcess a b)) -> Term s r) -> Term s r
```
L-as, line 100: We should still be returning `PTryFromExcess a b s` instead of `Term s (PTryFromExcess a b)`

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,404 @@
...
+
+----------------------- The class PFrom -------------------------------------------------
+
+{- |
+    Represents infallible conversion.
+    Laws:
+    - The result type must always carry less information than the origin
+      type
+-}
+class PFrom a b where
```
L-as, line 340: I don't think we should have this honestly

plutarch-test/src/Plutarch/TryFromSpec.hs
```diff
@@ -0,0 +1,482 @@
...
+  type PTryFromExcess PData (PAsData PNatural) = PNatural
+  ptryFrom opq = runTermCont $ do
+    (wrapped, unwrapped) <- TermCont $ ptryFrom @PData @(PAsData PInteger) opq
+    ver <- tcont $ plet $ pmkNatural # unwrapped
+    pure $ (punsafeCoerce wrapped, ver)
+
+validator :: Term s PValidator
+validator = phoistAcyclic $
+  plam $ \dat red ctx -> unTermCont $ do
+    (_, trustedRedeemer) <- TermCont $ ptryFrom @PData @(PAsData (PBuiltinList (PAsData PNatural))) red
```
L-as, line 321: Can we use a custom data type that is PIsDataRepr for both redeemer and datum? That would make the example much more convincing

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,404 @@
...
+  {-# OVERLAPPABLE #-}
+  ( PTryFrom PData (PAsData b)
+  , PTryFrom PData (PAsData (PDataRecord xs))
+  , x ~ (s ':= b)
+  , PIsData (PDataRecord xs)
+  , PTryFromExcess PData (PAsData (PDataRecord xs)) ~ PDataRecord xs
+  ) =>
+  PTryFrom PData (PAsData (PDataRecord (x ': xs)))
+  where
+  type PTryFromExcess PData (PAsData (PDataRecord (x ': xs))) = PDataRecord (x ': xs)
```
L-as, line 211: so an issue here is that we're actually discarding the "excess" of the inner verification. When you make it such that `PTryFromExcess a b s` is returned instead, you can return a Haskell pair of the excess stuff, such that everything is included.

L-as (changes requested):
Looks mostly good to me

plutarch-test/src/Plutarch/TryFromSpec.hs
MangoIV: I will try that.

Plutarch/Api/V1/AssocMap.hs
MangoIV: copied.

Plutarch/Maybe.hs
MangoIV: copied

Plutarch/TryFrom.hs
MangoIV: copied

Plutarch/TryFrom.hs
MangoIV: copied

Plutarch/TryFrom.hs
MangoIV: copied, I removed `PFrom` and `PMaybeFrom` for now

plutarch-test/src/Plutarch/TryFromSpec.hs
```diff
@@ -0,0 +1,482 @@
...
+    (_, trustedDatum) <- TermCont (ptryFrom @PData @(PAsData (PBuiltinList (PAsData PNatural))) dat)
```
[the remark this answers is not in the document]
MangoIV: so in reality we want to just `punsafeCoerce` it? or do we want to add a type alias for `PData`, say `PTrustedData` with a method like 
```haskell
pfromTrustedData :: Term s PTrustedData -> Term s a
pfromTrustedData = punsafeCoerce
```?

File: Plutarch/TryFrom.hs [hunk 36]
```diff
@@ -0,0 +1,439 @@
...
+          the expected type.
+      - the operation `ptryFrom` proves equality between the less expressive `PType` `a` and
+        the more expressive `PType` `b`, hence the first element of the resulting Tuple
+        must always be wrapped in `PAsData` if the origin type was `PData` (see law 1)
+      - the result type `b` must always be safe than the origin type `a`, i.e. it must carry
+        more information
+-}
+class PTryFrom (a :: PType) (b :: PType) (s :: S) where
+  type PTryFromExcess a b :: HTree
+  ptryFrom :: Term s a -> ((Term s b, HSTree (PTryFromExcess a b) s) -> Term s r) -> Term s r
```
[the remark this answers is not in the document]
L-as: In the case where you don't do this, you'd just return `PTryFromExcess a b s`.

L-as:
I fixed a bunch of the issues but the commit is still WIP

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,439 @@
...
+
+----------------------- HTree and friends -----------------------------------------------
+
+data HTree
+  = HLeaf Symbol PType
+  | HNode Symbol [HTree]
+
+data HSTree (as :: HTree) (s :: S) where
```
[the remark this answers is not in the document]
MangoIV: this has been adjusted

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,439 @@
...
+        - all conversion are fallible, this happens if the representation doesn't match
+          the expected type.
+      - the operation `ptryFrom` proves equality between the less expressive `PType` `a` and
+        the more expressive `PType` `b`, hence the first element of the resulting Tuple
+        must always be wrapped in `PAsData` if the origin type was `PData` (see law 1)
+      - the result type `b` must always be safe than the origin type `a`, i.e. it must carry
+        more information
+-}
+class PTryFrom (a :: PType) (b :: PType) (s :: S) where
+  type PTryFromExcess a b :: HTree
```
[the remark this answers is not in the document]
MangoIV: this has been adjusted

plutarch-test/src/Plutarch/TryFromSpec.hs
MangoIV: I have added an example that utilizes a custom datatype and shows how deriving `PTryFrom` can work with those.

Plutarch/Builtin.hs
```diff
@@ -225,8 +227,18 @@ instance PConstant (PAsDataLifted a) where
...
+
+pforgetData :: forall s a. Term s (PAsData a) -> Term s PData
+pforgetData x = coerce $ pforgetData' Proxy (coerce x :: Term s (Helper1 (PAsData a)))
+
+-- | Like 'pforgetData', except it works for complex types.
+pforgetData' :: forall a (p :: PType -> PType) s. Proxy p -> Term s (p (PAsData a)) -> Term s (p PData)
+pforgetData' Proxy = punsafeCoerce
+
+-- | Like 'pforgetData', except it works for complex types.
+prememberData :: forall (p :: PType -> PType) s. Proxy p -> Term s (p PData) -> Term s (p (PAsData PData))
```
TotallyNotChase, line 240: these two functions are very nice (though this one's haddock is wrong), but it'd be nice to have this for all  `PIsDataRepr a` as well.

```hs
PIsDataRepr => Term s (p (PAsData a)) -> Term s (p a)
```
and such, perhaps not as a separate function though.

L-as:
I've removed the #344 stuff for now unfortunately because it's not clear what the nature of this is on-chain. There's also the issue with the `PConstant` instance for `ScriptContext` and everything in it. We can't put the sortedness check in the `pconstantToRepr` because it wouldn't be coherent wrt. `toData`, so the implementation that was in this PR was unfortunately incorrect and not trivially fixable. I added the laws to the documentation as I thought about this.

Plutarch/Api/V1.hs
```diff
@@ -25,7 +25,7 @@ module Plutarch.Api.V1 (
   type PStakeValidator,
 
   -- ** Value
-  Value.PValue (PValue),
```
L-as, line 28: I need to remember to revert this

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
```diff
@@ -23,6 +23,8 @@ import Plutarch.Builtin (pasConstr, pforgetData)
 import Plutarch.Prelude
 import Plutarch.Test
 
+-- import PlutusTx.AssocMap as PlutusMap
```
L-as, line 26: I need to remember to remove this

plutarch-test/plutarch-test.cabal
```diff
@@ -133,6 +134,7 @@ executable plutarch-test
     other-modules:
       Plutarch.FieldSpec
       Plutarch.MonadicSpec
+      Plutarch.TryFromSpec
```
L-as, line 137: I don't think we use any GHC 9 features here

MangoIV: I added it because something was broken that sid fixed and I left it like that to be able to demonstrate the `HasField` instance

Plutarch/Builtin.hs
```diff
@@ -170,12 +172,12 @@ instance PListLike PBuiltinList where
 instance (PLift a, PEq a) => PEq (PBuiltinList a) where
   (#==) xs ys = plistEquals # xs # ys
 
-data PData s
-  = PDataConstr (Term s (PBuiltinPair PInteger (PBuiltinList PData)))
-  | PDataMap (Term s (PBuiltinList (PBuiltinPair PData PData)))
-  | PDataList (Term s (PBuiltinList PData))
-  | PDataInteger (Term s PInteger)
-  | PDataByteString (Term s PByteString)
+data PData (s :: S) = PData (Term s PData)
```
TotallyNotChase, line 175: you could've kept this the way it was and implemented `PlutusType` with `ChooseData` or something. But this is fine too, if slightly confusing to people expecting isomorphism with the `Data` declaration.

Plutarch/Builtin.hs
```diff
@@ -240,6 +253,12 @@ instance PIsData a => PIsData (PBuiltinList (PAsData a)) where
   pfromData x = punsafeCoerce $ pasList # pforgetData x
   pdata x = punsafeBuiltin PLC.ListData # x
 
+newtype Helper2 f a s = Helper2 (PAsData (f a) s)
+
+instance PIsData (PBuiltinList PData) where
```
TotallyNotChase, line 258: what's up with the complexity here?

plutarch-test/conditional/Plutarch/TryFromSpec.hs
```diff
@@ -0,0 +1,506 @@
...
+
+checkDeep ::
+  forall (target :: PType) (actual :: PType).
+  ( PTryFrom PData (PAsData target)
+  , PIsData actual
+  , PIsData target
+  ) =>
+  ClosedTerm (PAsData actual) ->
+  ClosedTerm (PAsData target)
+checkDeep t = unTermCont $ fst <$> TermCont (ptryFrom @(PAsData target) $ pforgetData t)
```
TotallyNotChase, line 267: I was under the impression the checking is only for `PData`. Are we encouraging checking a correctly typed `PAsData` as well? If a `PAsData` is actually incorrect - doesn't that mean the typing has been compromised further up in a way it shouldn't have been?

plutarch-test/conditional/Plutarch/TryFromSpec.hs
```diff
@@ -0,0 +1,506 @@
...
+    (ter, exc) <- TermCont $ ptryFrom @(PAsData PInteger) opq
+    ver <- tcont $ plet $ pmkNatural #$ exc
+    pure $ (punsafeCoerce ter, ver)
+
+validator :: Term s PValidator
+validator = phoistAcyclic $
+  plam $ \dat red ctx -> unTermCont $ do
+    trustedRedeemer <- (\(snd -> red) -> red) <$> (TermCont $ ptryFrom @(PAsData (PBuiltinList (PAsData PNatural))) red)
+    let trustedDatum :: Term _ (PBuiltinList (PAsData PNatural))
+        trustedDatum = pfromData $ punsafeCoerce dat
```
TotallyNotChase, line 318: might be nice if we had a way to "verify" datums without `punsafeCoerce`. something that actually just does `punsafeCoerce` but is only restricted to datums. `PDatum` might actually come in handy for once.

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,367 @@
...
+----------------------- PDataRecord instances -------------------------------------------
+
+-- We could have a more advanced instance but it's not needed really.
+
+newtype ExcessForField (a :: PType) (s :: S) = ExcessForField (Term s (PAsData a), Reduce (PTryFromExcess PData (PAsData a) s))
+
+instance Reducible (PTryFromExcess PData (PAsData a) s) => Reducible (ExcessForField a s) where
+  type Reduce (ExcessForField a s) = (Term s (PAsData a), Reduce (PTryFromExcess PData (PAsData a) s))
+
+-- FIXME: Should we always succede? If we always succede, performance would increase a lot.
```
TotallyNotChase, line 234: outdated comment?

Plutarch/TryFrom.hs
```diff
@@ -0,0 +1,367 @@
...
+
+{- | if there is an instance to recover something that is unwrapped from something that is
+ unwrapped, then there is also the possibility to recover the whole thing but wrapped
+-}
+instance
+  ( PTryFrom a b
+  , PIsData a
+  , PIsData b
+  ) =>
+  PTryFrom (PAsData a) (PAsData b)
```
TotallyNotChase, line 328: I think it's better to encourage users to do these conversions on `PData` only. The only time I see a `PAsData a` -> `PAsData b` conversion being relevant is if the `a` is `PData`, in which case, you should probably just use `pforgetData`. Am I missing something?

TotallyNotChase (commented):
this is very nice, great work guys!

well, the `PMap` stuff is very concerning. I think the issues with `PConstant` could potentially be solved by having a separate method that normalizes the result if it makes sense for the type. For most types, it'd be the same as a regular `pconstantFromRepr` though. This starts getting complicated when users have maps/values in datums and redeemers.

The best outcome is if we can ensure that Plutip, at the very least, does indeed sort things. I do want to get to that this week, though I'd have to come up with a good enough test case.

Perhaps @t1lde could shed some light regarding whether or not this stuff is sorted on the actual chain, at least?

Plutarch/Builtin.hs
L-as: `PlutusType` wasn't implementable in an efficient  way due to the difference between `UnConstrData` and `ConstrData`. We could possibly do it with but that's Plutarch 2.0.

Plutarch/Builtin.hs
L-as: This is how you do type-level functions in Haskell lol. With dependent types this would have been an order of magnitude less complex, unfortunately we do not have dependent types. I still think this is worth it because the transformation describes here is safe and still zero-cost.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
L-as: `PTryFrom` is generalised. You can e.g. have `PTryFrom (PAsData Integer) (PAsData Natural)`.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
L-as: Not all datums are trusted.

L-as: Some times you don't need to verify the redeemer either.

Plutarch/Builtin.hs
TotallyNotChase: well yeah, but you could've also done this with the builtins yeah? I assume this is safer? In which case, all good I suppose.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
TotallyNotChase: ah, I see - we're overloading with any conversion, not just data ones. Cool

TotallyNotChase: although I do think the examples are a bit misleading for that purpose.
```hs
checkDeep
  @(PBuiltinPair (PAsData PInteger) (PAsData PByteString))
  @(PBuiltinPair (PAsData PByteString) (PAsData PByteString))
  (pdata $ ppairDataBuiltin # (pdata $ pconstant "foo") # (pdata $ pconstant "bar"))
```
I can't see this being idiomatic, but I might be missing something. The existence of such a check implies the `PBuiltinPair` typing is wrong and the types have been compromised somewhere.

Plutarch/TryFrom.hs
L-as: It's not outdated. It's about verifying whether the list is empty or not.

L-as:
@srid @TotallyNotChase it would be great if you could try out this in practice.

I don't think the abstraction is that ergonomic, and it would be possible to make it more ergonomic with HKDs I think, but until then, it should at least be usable. I am specifically worried about the usability of verification of `PDataRecord`, not to mention that verification for `PDataSum` doesn't actually give you the verified data, meaning it's very expensive.

I think one common issue we have here is that since the abstraction should ideally be zero-cost, or rather,
be no more expensive than doing it manually, we have to return both the `Term s a` and `Term s (PAsData a)` often.
If we return just one, that can be a lot of extra work necessary. We could fix this with rewrite rules, such that we can return
a special term `Term s a` that is in fact a pair of both, and `pdata`'ing it will be zero-cost.

We should probably do Plutarch 2.0 soon. Biggest issue is still PIR backend though.

Plutarch/Builtin.hs
L-as: I don't think there's any other way of doing this that is as efficient that isn't unsafe.

Plutarch/TryFrom.hs
TotallyNotChase: well yeah, I mean that there's no way to make this a "always succeeds" no-op right?

TotallyNotChase:
I like that pair idea, there are several places I've encountered during API design that could be optimized using that technique.

I'll definitely be trying it out. I suggest we get it merged and encourage more people to play with it even if it's not as ergonomic yet though, may as well complete this release cycle and get some more feedback from other people.

L-as:
OK, merging so we can work on this in tandem

Plutarch/Api/V1.hs
TotallyNotChase: did you forget to revert this

L-as: Yes, we should make another PR to fix this

L-as:
#400

PR #329, opened 2022-02-23
Start plutarch extra

Geometer1729:


TotallyNotChase:
ok so we definitely need to split this up. this PR should just contain the bare scaffold for plutarch-extra (as discussed). The utilities should follow steadily.

Also, the *fundamental* `PMap` (and `PValue`) should be in the main Plutarch library itself.

L-as:
Why does this have so many irrelevant commits?

Geometer1729:
> Why does this have so many irrelevant commits?

I based it on top of the list utils upstream branch because they have significant overlap.

L-as:
Please don't include anything but the *scaffold* for `plutus-extra`.

L-as:
i.e. the cabal files

MangoIV:
I've moved this to #351 and the list upstream part to #357

PR #333, opened 2022-02-25
Add `puncons`, `punsafeUncons` in `Plutarch.List`

Renegatto:


Plutarch/List.hs
```diff
@@ -133,6 +138,20 @@ pconvertLists = phoistAcyclic $
       (\x xs -> pcons # x #$ self # xs)
       pnil
 
+-- | Extract head and tail of the list, throws error if list is empty.
+punsafeUncons ::
```
L-as, line 142: Note, this isn't unsafe. This should be named `ptryUncons` or similar.

Plutarch/List.hs
```diff
@@ -133,6 +138,20 @@ pconvertLists = phoistAcyclic $
       (\x xs -> pcons # x #$ self # xs)
       pnil
 
+-- | Extract head and tail of the list, throws error if list is empty.
+punsafeUncons :: 
+  PIsListLike list a =>
+  Term s (list a) ->
+  Term s (PPair a (list a))
+punsafeUncons = pelimList (\x -> pcon . PPair x) perror
```
L-as, line 146: This should also be hoisted and a Plutarch-level function.

Plutarch/List.hs
```diff
@@ -67,6 +71,7 @@ import Plutarch.Bool (PBool (PFalse, PTrue), PEq, pif, (#&&), (#==), (#||))
 import Plutarch.Integer (PInteger)
 import Plutarch.Lift (pconstant)
 import Plutarch.Pair (PPair (PPair))
+import Plutarch.Maybe (PMaybe (PJust,PNothing))
```
L-as, line 74: `./bin/format`

L-as (changes requested):
Thanks for the contribution!

Plutarch/List.hs
```diff
@@ -133,6 +138,20 @@ pconvertLists = phoistAcyclic $
...
 
+-- | Extract head and tail of the list, throws error if list is empty.
+punsafeUncons :: 
+  PIsListLike list a =>
+  Term s (list a) ->
+  Term s (PPair a (list a))
+punsafeUncons = pelimList (\x -> pcon . PPair x) perror
+
+-- | Extract head and tail of the list, if list is not empty.
+puncons ::
```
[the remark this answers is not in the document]
Renegatto: What exactly to hoist here since it's not recursive?

Plutarch/List.hs
Renegatto: What exactly to hoist here?

L-as:
Hoisting has nothing to do with recursion. It's to do CSE. You hoist closed terms generally.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Revision history for plutarch
 
 # Unreleased
+- Add `puncons` and `punsafeUncons` functions for deconstructing list.
```
L-as, line 4: We should fix this in another PR for fixing the changelog

L-as (approved):
LGTM

srid:
@L-as What do you think of requiring new PRs like this one that adds new code to also add the corresponding examples to `plutarch-test` (so we get uplc/bench goldens, to test for any potential regression in the future)?

Issue #19, opened 2021-12-17
Detect infinite recursion (at compile-time) using stable names


L-as:
This will likely heavily slow down the compiler, but it's probably worth it.

TotallyNotChase:
An example that (probably) belongs here-
```
recurH :: Term s PUnit -> Term s PUnit
recurH _ = pif (pcon PTrue) (pcon PUnit) $ recurH $ pcon PUnit

-- >>> compile $ recurH £ pcon PUnit
```

L-as:
A much simpler example is just
```
bad :: Term s a
bad = bad
```

AriFordsham:
How do you do this without solving the halting problem?

L-as:
You can't. You can however catch many of the simple cases, just like Nix does.
If you do `let x = x; in x` in Nix, it's going to give you an error message.

L-as:
@TotallyNotChase Looking back on this almost three months later, has this ever been a problem for you in practice?

Also relevant for @srid and @Geometer1729

Geometer1729:
It hasn't been a problem for me yet.

TotallyNotChase:
Nope. It shouldn't be too common of a mistake to make in general. But I can definitely see the mistake *potentially* being made.

L-as:
That doesn't tell you where the problem is

Issue #260, opened 2022-02-09
Rewrite rules


L-as:
We want rewrite rules to support things like #128.

Rewrite rules should function like GHC's rewrite rules, albeit they should fire consistently. This is easier for us
to do because of the lack of complex optimisations.

delicious-lemon:
Yes! Is there any work toward this ongoing? 

My motivation is https://github.com/mlabs-haskell/apropos-tx/blob/26294dbaa3a1c9e794bdeab2af98d2f74624ab96/src/Apropos/Script/Iso/Arrow.hs#L49

These combinators introduce unnecessary construction/deconstruction of builtin pairs which could be reduced easily by a rewrite.

L-as:
I am working on this internally.

srid:
- Having a rewrite rule to "drop" traces (as part of #359) which incidentally resolves #290

blamario:
Just to be clear, are the rewrite rules supposed to operate on the compiled `Script`s like Shrinker does, or on a `Term`?

L-as:
@blamario They operate on the precompilation AST.

L-as:
I think I have a viable design FWIW

srid:
What sort of optimizations will rewrite rules provide? 

Specifically, if I use `pfield @"txInfo" ctx` in more than one place, will that get optimized (ie, hoisted at top-level and gets used via reference, instead of recomputing)?

L-as:
In some places however this means we have to do a `plet` and that's not free.

L-as:
They're supposed to operate on Plutarch terms.

Issue #259, opened 2022-02-09
Thread monad through `Term`


L-as:
This is what we want:

```haskell
type Term :: forall (s :: S). (Type -> Type) -> Type
type ClosedTerm :: (Type -> Type) -> Type

compile :: Monad m => ClosedTerm m a -> m Script
joinTerm :: Monad m => m (Term @s m a) -> Term @s m a
```

Reasoning:
Currently we have a bunch of functions that can err depending on how you use them, and we're using `error` for that.
This is not optimal, and makes it hard to e.g. catch (which we do in the tests).
There are also a lot of other use cases, for example,
one might want to have something like `x :: MonadIO m => m (Term s a)` currently, but this is very unergonomic, because
you can't e.g. do `x + x`, because `+` is defined for `Term s a` and not `m (Term s a)`. We could use `liftA2` here, but if we thread the monad through the term itself, then `x + x` would become completely legal.

PR #335, opened 2022-02-25
`PEq` generic deriving

### LINKED ISSUE #325: `PEq` generic derivation
`Eq` (from Haskell prelude) seems to be one of the most common classes derived for on chain types - we should have similar ergonomics on `PEq`.

I imagine it won't be *too hard* to do generic derivation for `PEq` - we're really only interested in scott encoding stuff. Non-data builtins are taken care of manually and `Data` encoding (`PIsDataRepr`) should be taken care of by #324

srid:
Resolves #325


L-as:
Wasn't @TotallyNotChase also working on this?

L-as:
It's really nice how we can have standard generic deriving for all the standard type classes

Plutarch/Bool.hs
```diff
@@ -117,3 +146,42 @@ por = phoistAcyclic $ plam $ \x y -> pif' # x # (phoistAcyclic $ pdelay $ pcon P
...
+  ccompare_NS (Proxy @(All PEq)) (pcon PFalse) eqProd (pcon PFalse) c1 c2
+  where
+    eqProd :: All PEq xs => NP (Term s) xs -> NP (Term s) xs -> Term s PBool
+    eqProd p1 p2 =
+      foldl' (#&&) (pcon PTrue) $
+        hcollapse $ hcliftA2 (Proxy :: Proxy PEq) eqTerm p1 p2
+      where
+        eqTerm :: forall a. PEq a => Term s a -> Term s a -> K (Term s PBool) a
+        eqTerm a b =
+          K $ a #== b
```
TotallyNotChase, line 169: you may have already found this out but be careful of polymorphism in the case of recursion! This will invoke the [infinite AST problem] if the data type is recursive! (e.g `PList`)

srid: I hadn't thought of it! But yes, this is indeed a problem: https://hercules-ci.com/github/Plutonomicon/plutarch/jobs/998

srid: Fixing #19 will be beyond the scope of this PR. Generics cannot help here. 

I hand-wrote a PEq instance for PPair:

```haskell
instance (PEq a, PEq b) => PEq (PPair a b) where
  p1 #== p2 = go # p1 # p2 
    where 
      go = phoistAcyclic $ trace "pair:plam" $ plam $ \x' y' ->
        pmatch x' $ \(PPair a b) -> pmatch y' $ \(PPair a' b') -> 
          a #== a' #&& b #== b'
```

And then defined a tree type in terms of it:

```haskell
data PTree a s
  = PLeaf (Term s a)
  | PBranch (Term s (PPair (PTree a) (PTree a)))
  deriving stock (GHC.Generic)
  deriving anyclass (Generic, PlutusType)
```

PTree's PEq instance:

```haskell
instance PEq a => PEq (PTree a) where 
  t1 #== t2 = go # t1 # t2
    where 
      go = phoistAcyclic $ trace "plam"  $ 
        plam $ \x' y' -> 
          pmatch x' $ \x -> pmatch y' $ \y -> 
            case (x, y) of 
              (PLeaf a, PLeaf b) -> 
                a #== b
              (PBranch a, PBranch b) ->
                a #== b
              _ -> 
                pcon PFalse
```

No generics is involved; but comparing two tree values causes a mutual recursive loop in the compiler, as evidenced by the flurry of interleaved `plam` and `ppair:plam` traces messages.

image

TotallyNotChase: No need to fix #19. This is more of an issue with the derivation itself. We *do* have to take care of it before allowing automatic `PEq` deriving at all since this is a big trap.

Fortunately, this is the part that I was experimenting on when working on `PEq` generic derivation myself. I thought of dispatching a call to either `self` (bound by a pfix #$ plam $ \self ... ->`) or `#==` depending on the outer type and inner field type. This has certain drawbacks that I haven't been able to solve. I originally wanted some way for the compiler to optimize out `pfix`s if the `self` argument was unused - but that is not currently possible.

srid: Dispatching a call to `self` would work for recursive types (and I'm curious how you managed to do this generically), but not for mutually recursive types like `PTree` above. 

I was thinking a compiler-level automatic `pfix` would be the solution for both these cases, including automatic use of Y* combinator as appropriate.

L-as: Interesting case to consider
```haskell
data A = A (Term s B) (Term s B) | ANil
data B = B (Term s A) (Term s A) | BNil
```

L-as: The equality functions for both would refer to each other, and hence be mutually recursive. This is a pretty big problem. Not sure how we can do this automatically in a good way. Perhaps we can define "mutually recursive groups" and automatically convert it into one `pfix` somehow such that for functions `fs : List Type`, we create `f : (n : Fin (length fs)) -> index fs n`.

L-as: We probably need to do this with #260 and #259 somehow

TotallyNotChase: Can letrec do this? Or did I completely misunderstand letrec?

L-as: `letrec` does this in an explicit way, we need implicit mutual recursion essentially. @srid we should just assume non recursive data types right now. A constructor should not contain the type being derived for for deriving to work right now. Fixing this would require a *lot* of work.

L-as:
FWIW I don't think handling the non-mutually recursive case is worth it. It's probably pretty rare in practice, and it adds a lot of extra code AFAICT, in addition to being inefficient because of the unnecessary `pfix` in simple cases, which can't be optimised away without #260 .

L-as:
Also, `punsafeHoistAcyclic` isn't really needed IMO. You only need `punsafeAsClosedTerm`.

Plutarch/Bool.hs
srid: I am finishing @TotallyNotChase 's original idea of always using `pfix` in the instance implementation, and thereby calling `self` for when checking equality of the inner type ... but it does come with the cost of slightly increased cpu/mem numbers (due to pfix being used regardless of data type being recursive). 

Do we keep that approach ... or just have generic deriving unconditionally fail on types like `PList` (where one of the constructors contains the type being derived for), if not for types like `A`/`B` above?

srid:
> FWIW I don't think handling the non-mutually recursive case is worth it. It's probably pretty rare in practice, and it adds a lot of extra code AFAICT, in addition to being inefficient because of the unnecessary `pfix` in simple cases, which can't be optimised away without #260 .

Okay ... removing it then.

TotallyNotChase:
Gotta make sure to put the pitfall in the guide then - will do that prior to 1.2 merge to master

srid:
I think this PR is ready. Is there anything I should do here?

Plutarch/Bool.hs
```diff
@@ -117,3 +139,37 @@ por = phoistAcyclic $ plam $ \x y -> pif' # x # (phoistAcyclic $ pdelay $ pcon P
...
+  ( PGeneric s t
+  , PlutusType t
+  , All2 PEq (PCode s t)
+  ) =>
+  Term s (t :--> t :--> PBool)
+gpeq =
+  phoistAcyclic $
+    punsafeAsClosedTerm @s $
+      plam $ \x y ->
+        pmatch x $ \x' ->
```
TotallyNotChase, line 155: since chained `pmatch`es like these are really inefficient (and since we have no other way to do this right now) - is it worth thinking about implementing a better method, in the future?

TotallyNotChase (approved):

Issue #253, opened 2022-02-08
`instance PlutusType (PDataSum as)`


L-as:
We're missing this

PR #345, opened 2022-03-02
Sean/#253

gnumonik:
First draft of a fix for #253. The changes should be adequately explained by the comments. General idea is that we need constructors to give PDataSum a PlutusType instance, so I gave it constructors and wrote the instance! 

This needs to be looked over by someone who is more familiar with the library than I am. Would greatly appreciate feedback / suggestions.


Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +186,161 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+    d@Dict -> go ixed d 
+ where 
+   go :: forall (n' :: Nat) a as . IndexedSum (n' + 1) (a:as) 
+      -> Dict (KnownNat n') 
+      -> Dict (KnownNat (n' + 1))
+   go (IxS _) Dict = mapDict plusNat (Dict :: Dict (KnownNat n', KnownNat 1))
+
+-- maybe don't need the constraints? 
+addLength :: forall k ks. (IndexableSum (k:ks), IndexableSum ks) => Length (k:ks) :~: Length ks + 1 
+addLength = UNSAFE.unsafeCoerce Refl
```
L-as, line 286: this is really bad....

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +186,161 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+
+4) PDataSum' is the helper type which allows us to express those invariants. The structure of the type has to be inductive, but 
+   we need four constructors (two Zero / two Suc) to cover each of the possible cases. The Boolean type argument indicates whether or not the 
+   sum contains an element, while the Natural type argument represents the position of the element in the list of choices. 
+
+      a) 'SumZE' represents a one element sum that *does not* contain its element.  
+
+      b) 'SumZF' represents a one element sum that *does* contain its element.
+
+      c) 'InjE' injects an element into a sum which does not contain any elements such that the sum will contain the injected element.
```
L-as, line 210: Inj_?

Plutarch/DataRepr/Internal.hs
gnumonik: Ugh, yeah. I'll find some way to get rid of it.

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +186,161 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+7) We need some helper utilities to reconstruct the Sum (which we have to do) in pmatch'. Those utilities are: 
+    
+    a) 'IndexedSum'. This is basically a singleton list a la SList from sop-core, except it can only represent nonempty lists, has an 
+       "explicit" inductive structure, and indicates the position of elements explicitly. 
+
+    b) A 'Length' type family. This doesn't work like 'length' from Prelude: 'Length '[a]' == 0, not 1. 
+
+    c) A type class 'IndexableSum' which is like SListI from sop-core, except for our 'IndexedSum' type. As far as I can tell it is impossible 
+       to reconstruct the sum without an inductive GADT that represents the structure of the sum. We need to conjure that structure out of thin air,
+       so we need a type class.
```
srid, line 232: I'm curious why we can't just reuse `NS` and the like from sop-core. Where is the non-empty property being required?

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +186,161 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+  Inj :: forall {n :: Natural} {x :: [PLabeledType]} {xss :: [[PLabeledType]]} {s :: S} 
+        . (KnownNat (n+1))
+        => (forall {out :: PType}. (Term s (PDataRecord x) -> Term s out) -> Term s out) 
+        -> PDataSum' n 'False xss s 
+        -> PDataSum' (n + 1) 'True (x ': xss) s 
+
+  Inj_ :: forall {n :: Natural} {b :: Bool} {x :: [PLabeledType]} {xss :: [[PLabeledType]]} {s :: S} 
+        .  KnownNat (n+1)
+        => PDataSum' n b xss s 
+        -> PDataSum' (n + 1) b (x ': xss) s
```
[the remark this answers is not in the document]
gnumonik: Yeah you should be able to infer it, I dunno what I was thinking there. That should be an easy fix. Thanks for pointing that out.

Plutarch/DataRepr/Internal.hs
gnumonik: NS enforces the nonempty property too, doesn't it? I guess that could work as a replacement for IndexedSum. 

I don't think it works as a replacement for the PDataSum' stuff though. I'm basing that on intuition + failing to get it to work (it was the first thing I tried for PDataSum).

Plutarch/DataRepr/Internal.hs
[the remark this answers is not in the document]
srid: I haven't looked at this PR in detail, but you can probably get rid of the `Bool` index as well if `PDataSum'`'s shape is made to be similar to that of [`NS`](https://hackage.haskell.org/package/generics-sop-0.5.1.2/docs/Generics-SOP.html#t:NS). Something like:

```haskell
data PDataSum' :: [[PLabeledType]] -> PType where 
  DSZ :: (forall {out :: PType}. (Term s (PDataRecord x) -> Term s out) -> Term s out) 
           -> PDataSum' (x ': xs) s
  DSS :: PDataSum' xs s -> PDataSum' (x ': xs) s
```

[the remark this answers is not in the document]
gnumonik: Mkay, I'll see if I can make something like that work. I agree that the `Bool` index is probably superfluous and can likely be discarded.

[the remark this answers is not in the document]
gnumonik: Ok I'm a moron and totally forgot how NS actually works. (In my defense I haven't used the SOP stuff for like a year and a half). For some reason I was assuming that the index of Z was a one-element list.  

Your suggested alternative appears to work, and makes `PDataSum'` obsolete.

gnumonik:
New commits should address all of the above suggestions/concerns. Code is significantly cleaner/more compact now.

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +184,55 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+  pcon' :: forall 
+           {s :: S} 
+           {b :: PType}
+         . PDataSum (def:defs) s 
+        -> Term s (PInner (PDataSum (def:defs)) b)
+  pcon'  = \case 
+        DSZ f ->  punsafeBuiltin PLC.ConstrData # (pconstant @PInteger $ adjustedLength $ Proxy @(def:defs)) # (f id) -- rewrite this with pconstrBuiltin?
+        DSS rest -> case rest of 
+          DSZ f -> punsafeBuiltin PLC.ConstrData # (pconstant @PInteger $ adjustedLength $ Proxy @(def:defs)) # (f id)
+          dss@(DSS _) -> pcon' dss
```
L-as, line 209: Isn't the index in this case incorrect?

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +184,55 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+          DSZ f -> punsafeBuiltin PLC.ConstrData # (pconstant @PInteger $ adjustedLength $ Proxy @(def:defs)) # (f id)
+          dss@(DSS _) -> pcon' dss  
+
+  pmatch' :: forall 
+             {s :: S} 
+             {b :: PType}
+           . Term s (PInner (PDataSum (def:defs)) b)
+          -> (PDataSum (def:defs) s -> Term s b) 
+          -> Term s b
+  pmatch' d f =
```
L-as, line 217: Can we use `pmatchDataSum` here somehow?

L-as (changes requested):
Looks mostly good!  Can you add some examples of how to use it too?

L-as:
Also, please target `staging`

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +184,55 @@ type family PUnLabel (a :: PLabeledType) :: PType where
  This is how most data structures are stored on-chain.
 -}
 type PDataSum :: [[PLabeledType]] -> PType
-data PDataSum (defs :: [[PLabeledType]]) (s :: S)
+data PDataSum (defs :: [[PLabeledType]]) (s :: S) where 
+  DSZ :: (forall {out :: PType}. (Term s (PDataRecord x) -> Term s out) -> Term s out) 
+      -> PDataSum (x ': xs) s
+
+  DSS :: PDataSum xs s -> PDataSum (x ': xs) s
```
srid, line 191: This is now isomorphic to `NS`, right? viz.:

```haskell
newtype DRec s x = DRec (forall {out :: PType}. (Term s (PDataRecord x) -> Term s out) )
type PDataSum defs s = NS (DRec s) defs
```

Plutarch/DataRepr/Internal.hs
```diff
@@ -180,7 +184,55 @@ type family PUnLabel (a :: PLabeledType) :: PType where
...
+
+-- the length functions in sop-core gives the actual length, we want length-1
+adjustedLength :: forall x xs proxy. SListI (x:xs) => proxy (x:xs) -> Integer 
+adjustedLength proxy = fromIntegral (lengthSList proxy) - 1  
+
+instance SListI (def:defs) => PlutusType (PDataSum (def:defs)) where
```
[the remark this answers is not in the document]
gnumonik: You're not familiar with its purpose in general, or here? 

In general, it's more or less the generics-sop version of `SingI` from singletons except specialized to lists (and way more annoying to work with). It's basically just a constraint that asserts that the typelevel list is known at compile time. 

Here, we need it for two things: 

1) It's necessary to calculate the length of the typelevel list so we can get the index of the correct variant. This is a consequence of the NS-like representation; we can't just count the constructors because, with the NS-like representation, we don't have a constructor for every element of the typelevel list. (This is why I thought it was initially impossible to use an NS-like representation - I forgot SListI existed. Under the old representation, you *could* count the constructors.) 

2) It's necessary to implement the helper function `go` in `pmatch'`. We need some kind of GADT from which we can recover information about the structure of the typelevel list. I don't think there's any way to make this typecheck without that while rebuilding the sum. I also don't think there's way to pass that information explicitly in the arguments to `pmatch'`, so we have to conjure it into existence, which requites a type class constraint. 

It looks like there *might have been some machinery at one point relating to `HList`s in the Plutarch.DataRepr.Internal.HList / Plutarch.DataRepr.Internal.HList.Utils modules that we could use instead of SListI, but either the actual HList stuff never got added or was replaced with HRec.  

I think we need something roughly equivalent to SListI/SList. Since we're already using Generics.SOP anyway, it seemed like the most parsimonious choice. If you have an idea of how to get around needing the constraint I'd be happy to implement it, but I can't think of anything.

Plutarch/DataRepr/Internal.hs
gnumonik: Ugh yeah, sorry about that. I made all the changes in a rush and missed that. I think it should be 

``` haskell 
        DSS rest -> case rest of 
          DSZ f -> punsafeBuiltin PLC.ConstrData # (pconstant @PInteger $ adjustedLength $ Proxy @(defs)) # (f id)
          dss@(DSS _) -> pcon' dss  
```

Plutarch/DataRepr/Internal.hs
gnumonik: I don't  _think_ we can. But I could be wrong and if it's important I can look into it more. 

If I remember right the problem is that we can't get the `Nat` we need to use `mkDataReprHandler` from the `PInteger` that indicates the index of the constructor, and we can't construct the handlers manually because there's no (straightforward) way (that I could think of) to get a `Term s (PDataRecord def) -> Term s out`
from a `PDataSum defs s -> Term s out`. I'm less confident that there's no way around the second problem  -- there could well be something in the rest of the library that I'm not aware of which would make that possible.  

Does that make sense? If my reasoning here is flawed let me know.

Plutarch/DataRepr/Internal.hs
gnumonik: Yeah, I realized that like 20 seconds after the last commit. Should be an easy change.

gnumonik: `PDataSum` has to be a newtype to get the instance to work (GHC restriction on partially applied type synonyms) and I think `DRec` has to be a little different: 

```haskell
newtype DRec s x = DRec (forall {out :: PType}. ((Term s (PDataRecord x) -> Term s out) -> Term s out))

type PDataSum :: [[PLabeledType]] -> PType
newtype PDataSum defs s = PDataSum (NS (DRec s) defs)
```
Morally I think those are close enough to what you suggested though.

gnumonik:
Ok I think I addressed the second round of comments with the exception of an example (wanted to wait until everything else was finalized). 

I'm not sure if I screwed something up by changing the target to staging here. Never used a draft PR before. Let me know if I should just close this + reopen a new PR targeting staging. 

I think the primary remaining issues depend on whether @L-as has any suggestions in response to my replies, so I'll wait on those before doing anything else.

Plutarch/DataRepr/Internal.hs
```diff
@@ -41,18 +41,22 @@ import Generics.SOP (
   AllZipN,
   Code,
   Generic,
-  I (I),
+  I (..),
   K (K),
   LiftedCoercible,
   NP (Nil, (:*)),
+  NS (..),
```
L-as, line 48: Please don't use `..` imports.

Plutarch/DataRepr/Internal.hs
L-as: I very much feel like we can use `pmatchDataSum` here.

We can have a type which internal representation is `PDataSum defs`, and implements `PlutusType` such that `pmatch` uses `pmatchDataSum`, so we should be able to do the same here. `pmatchDataSum` implements important optimisations. Look at the `PIsDataRepr` class and generic instance for `PlutusType` using the `deriving via` wrapper.

L-as (changes requested):
See comments, thanks!

Plutarch/DataRepr/Internal.hs
gnumonik: Ok, I will take another look at it when I have some time. I did not look very closely at `PIsDataRepr`, hopefully that opens up some possibility I wasn't seeing. 

Just a heads up that I might not be able to get to this for a day or two.

srid:
Let's also address CI failures before the merge; in particular, 9.2 only syntax like `forall {..}` should be removed as they won't work on GHC 8.10.

TotallyNotChase (commented):
The remaining comments should've been addressed by these changes.

Also, I need to make sure none of the existing goldens changed - I'll do this now.

TotallyNotChase:
interesting CI failure.

L-as (commented):
LGTM besides Chase's comments

TotallyNotChase:
I spun up a [meta generics-sop issue](https://github.com/Plutonomicon/plutarch/issues/426). We'll tackle those on a separate PR.

I also checked the goldens to make sure existing tests weren't changed.

@L-as This PR can be merged now, though the CI is still a problem.

roberth:
> interesting CI failure.

Fixed in hercules-ci-agent 0.9.3

PR #347, opened 2022-03-02
Add `psatisfyWithinBenchmark`

srid:
Helpful to test whether a given Plutarch program is within the specified CPU/mem/size budget.

image

It also adds by necessity the `@:->` operator to pass the benchmark value while keeping `@->` as-is for backwards compatibility and satisfying the majority of use cases checking the evaluated program rather than benchmarks.


Geometer1729 (approved):
LGTM

L-as:
evalScript already does this

L-as:
https://github.com/Plutonomicon/plutarch/blob/staging/Plutarch/Evaluate.hs#L41

srid:
Yes, but the idea here is to test that a Plutarch program (any program used in goldens) takes a _specified_ budget (not just the max budget used by `evalScript`) including script size. @Geometer1729 ran into this when he caught equality checks on `Value` taking more than usual memory usage on large lists. 

If you look at the numbers in the `fib` example I've posted above, notice that it also checks for script size ceiling (100 bytes, specifically).

L-as:
Makes sense. You probably want evalScript' though.

srid:
I'm now thinking we should perhaps modify `@:->` to take another argument, representing the evaluated UPLC. Thus it accepts a function taking 3 arguments, each corresponding to the 3 goldens being generated.

srid:
> I'm now thinking we should perhaps modify `@:->` to take another argument, representing the evaluated UPLC. Thus it accepts a function taking 3 arguments, each corresponding to the 3 goldens being generated.

I did that change ^ in a separate PR for review-friendly diff: #353 (/cc @Geometer1729)

plutarch-test/src/Plutarch/Test/Golden.hs
```diff
@@ -1,8 +1,11 @@
+{-# LANGUAGE ImpredicativeTypes #-}
```
L-as, line 1: Hard dependency on GHC 8?

L-as (changes requested):

plutarch-test/src/Plutarch/Test/Golden.hs
L-as: I meant to write GHC 9. If you don't care about GHC 8, then sure, but I think we still do a bit. Anyway, just wrap the inner `ClosedTerm` in a newtype.

L-as: `Identity (ClosedTerm a)` might be enough.

L-as: otherwise make `newtype ClosedTerm' a = ClosedTerm' { unClosedTerm' :: ClosedTerm a }`

srid: @L-as I'm not sure I understand. CI for GHC 8.10 passes, therefore this change is still supported in GHC 8, I'd think? 

Since this code already compiles in GHC 8.10, what would be the advantage of wrapping in a newtype?

L-as:
https://downloads.haskell.org/ghc/9.2.2/docs/html/users_guide/exts/impredicative_types.html?highlight=impredicativetypes#extension-ImpredicativeTypes it's basically a new feature.

Surprised that it still works with GHC 8.

L-as (approved):
LGTM

L-as:
please resolve conflicts

L-as (approved):
LGTM

PR #349, opened 2022-03-03
Replaced GHC.Exts.Any with Data.Void.Void

blamario:
It turns out that `plutus-tx-plugin` doesn't like `Any`:
```
GHC Core to PLC plugin: E042:Error: Unsupported feature: Kind: forall k. k
Context: Compiling kind: forall k. k
Context: Compiling type: GHC.Types.Any @Type
```
This PR puts `Void` in its place, which `plutus-tx-plugin` doesn't complain about.

### DIFF
--- a/Plutarch/FFI.hs
+++ b/Plutarch/FFI.hs
@@ -16,7 +16,7 @@ module Plutarch.FFI (
 import Data.ByteString (ByteString)
 import Data.Kind (Constraint, Type)
 import Data.Text (Text)
-import GHC.Exts (Any)
+import Data.Void (Void)
 import GHC.Generics (Generic)
 import GHC.TypeLits (TypeError)
 import qualified GHC.TypeLits as TypeLits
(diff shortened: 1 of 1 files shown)


L-as (changes requested):
This is unsafe FYI. `applyCode` can be used to give us a value of type `Void`, and I don't think this would even be usable in practice because of the unsoundness and GHC's assumptions about the code. We'll have to inspect the generated code. You could also try with `()`.

You could also just remove the function or leave it as-is, noting that plutus-tx currently doesn't support it.

blamario:
I still don't understand how it's unsafe, but it is usable:
```
testCase "Export an integer and ignore it" $
            printCode ($$(PlutusTx.compile [|| const (7 :: Integer) ||]) `applyCode` opaqueExport (4 :: ClosedTerm PInteger))
              @?= "(program 1.0.0 ((\\i0 -> 7) 4))"
```

Issue #341, opened 2022-03-01
Semigroup instance for PUnit leaves args of  `<>` unevaluated


delicious-lemon:
https://github.com/Plutonomicon/plutarch/blob/d753dc34dfc30b144e94d6493c837ebd0c99b588/Plutarch/Unit.hs#L33

TotallyNotChase:
hmm, I'm not sure the alternative is always better. This is one of those things that might be better handled with Plutarch 2.0's goals

TotallyNotChase:
Not evaluating could be potentially useful depending on the context maybe? Although it _would_ be kinda silly for `PUnit`.

delicious-lemon:
@TotallyNotChase my use case is I want to treat `PUnit` as a Truth value so it would be nice if  `<>` behaved like `AND`.

TotallyNotChase:
one example I thought of where this impl could be considered useful is folding over a list of units with `<>`. Every element (all of those units) will indeed be evaluated anyway. However, with forceful evaluation within the `<>` body, it's sort of a redundant lambda application for each element. Not a big cost, but something to consider.

L-as:
This has been fixed

PR #352, opened 2022-03-03
Add `PShow`

srid:
Things to do,

- [x] Derive instances for other types
- [x] Add `ptraceShowId` and the like
- [x] `PByteString` instance (as hexstring)
- [x] `PShow`: quote the output via bytestring using utf8 encoding builtins


TotallyNotChase:
if we're adding a `Show` like typeclass - it should probably only exist in development mode (due to the size and cost). Or am I misunderstanding its practical usecases?

emiflake:
> if we're adding a `Show` like typeclass - it should probably only exist in development mode

I think if we can eliminate all traces anyway, the use of `Show` in a trace would be eliminated anyway, right?

srid:
`pshow` is mainly useful in `ptrace*` - so yes, in non-development mode all of this should vanish (unless you are using `pshow` _outside_ of tracing for some reason).

I'd expect users to just use `ptraceShowId` and the like, in conjunction with `deriving anyclass PShow` on their data types, without using `pshow` directly.

L-as (changes requested):
nice

Plutarch/Trace.hs
```diff
@@ -34,6 +44,14 @@ ptrace s a = pforce $ ptrace' # s # pdelay a
 ptrace _ a = a
 #endif
 
+-- | Like Haskell's `traceShowId` but for Plutarch
+ptraceShowId :: PShow a => Term s a -> Term s a
+#ifdef Development
+ptraceShowId x = ptrace (pshow x) x
```
TotallyNotChase, line 50: maybe this function should be called `ptraceShow`

srid: Well, it is analogous to Haskell's `traceShowId` - I kept the naming for consistency.

Plutarch/Show.hs
```diff
@@ -0,0 +1,143 @@
...
+-- | Return the string representation of a Plutarch value
+pshow :: PShow a => Term s a -> Term s PString
+pshow = pshow' False
+
+instance PShow PString where
+  -- FIXME: We ideally want to wrap the string in quotes, and escape it
+  -- appropriately just like Haskell's Show instance. However, there appears to
+  -- be no way to do this using Plutus' builtins. So we will just return the
+  -- string as is.
+  pshow' _ = id
```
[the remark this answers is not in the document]
srid: Yes, I already use `<>` liberally - but `pshow' _ x = "\"" <> x "\""` would not be correct as it wouldn't escape the double quotes inside of `x`.

Plutarch/Show.hs
```diff
@@ -0,0 +1,143 @@
...
+
+-- | Generic version of `pshow`
+gpshow ::
+  forall a s.
+  (PGeneric s a, HasDatatypeInfo (a s), PlutusType a, All2 PShow (PCode s a)) =>
+  Bool ->
+  Term s (a :--> PString)
+gpshow wrap =
+  phoistAcyclic $
+    punsafeAsClosedTerm @s $
```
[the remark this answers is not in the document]
srid: @L-as See also this PR which does the same: https://github.com/Plutonomicon/plutarch/pull/335/files#diff-bdc1d03d75f8e4a1a716fbda46bf5b0fcf50f7a1ede7d7730430ce0e8bcf15b7R151-R154

Plutarch/Show.hs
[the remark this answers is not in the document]
TotallyNotChase: oh wow that's a neat idea

Plutarch/Show.hs
```diff
@@ -99,10 +101,40 @@ instance PShow PInteger where
           pcase perror digit $
             flip fmap [0 .. 9] $ \(x :: Integer) ->
               (pconstant x, pconstant (T.pack . show $ x))
-      pcase :: PEq a => Term s b -> Term s a -> [(Term s a, Term s b)] -> Term s b
-      pcase otherwise x = \case
-        [] -> otherwise
-        ((x', r) : cs) -> pif (x #== x') r $ pcase otherwise x cs
+
+instance PShow PByteString where
+  pshow' _ x = "0x" <> showByteString # x
```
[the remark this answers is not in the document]
L-as: the `<>`

Plutarch/ByteString.hs
```diff
@@ -74,6 +78,23 @@ pbyteStr = pconstant
 pconsBS :: Term s (PInteger :--> PByteString :--> PByteString)
 pconsBS = punsafeBuiltin PLC.ConsByteString
 
+-- | Case matching on bytestring, as if a list.
+pelimBS ::
```
L-as, line 82: I don't think we should export this function. It's not directly necessary for Plutarch, nor is it even a good abstraction to use.

Plutarch/Show.hs
```diff
@@ -65,11 +65,29 @@ pshow :: PShow a => Term s a -> Term s PString
 pshow = pshow' False
 
 instance PShow PString where
-  -- FIXME: We ideally want to wrap the string in quotes, and escape it
-  -- appropriately just like Haskell's Show instance. However, there appears to
-  -- be no way to do this using Plutus' builtins. So we will just return the
-  -- string as is.
-  pshow' _ = id
+  pshow' _ x = "\"" <> pshowStr # x <> "\""
```
L-as, line 68: You can hoist this

L-as (changes requested):
Looks mostly good to me

srid:
Hoisted PString, PByteString impl, as well as unexposed pelimBS.

Plutarch/Unit.hs
```diff
@@ -39,4 +38,4 @@ instance Monoid (Term s PUnit) where
   mempty = pcon PUnit
 
 instance PShow PUnit where
-  pshow' _ x = pmatch x $ \PUnit -> "()"
+  pshow' _ _ = "()"
```
L-as, line 41: this is actually an oversight in the other instances, and I don't think this is actually a "fix", see #341

L-as (approved):
LGTM, we should fix the PUnit stuff in another PR

L-as:
@srid please resolve conflicts

L-as (approved):
LGTM

PR #353, opened 2022-03-03
Avoid redundant evaluation in `@==`

srid:
Also make `@:->` take the evaluated script as 3rd argument, as noted in https://github.com/Plutonomicon/plutarch/pull/347#issuecomment-1058238380


L-as (approved):
LGTM

Issue #286, opened 2022-02-15
Polymorphically deriving PLift,PConstant without the headaches


t1lde:
`DerivePConstantViaNewtype` is pretty painful to use with polymorphic types - the Constraints required for the derived instances are leaky. An example is:

```haskell

instance 
  forall a. 
  ( a ~ (PConstanted (PLifted a))
  , PUnsafeLiftDecl a
  ) =>
  PUnsafeLiftDecl (PFoo a) where 
  type PLifted (PFoo a) = Foo tag (PLifted a)

deriving via
  (DerivePConstantViaNewtype 
    (Foo  a)
    (PFoo (PConstanted a))
    (PConstanted a))
  instance 
  (a ~ (PLifted (PConstanted a))
  , PUnsafeLiftDecl (PConstanted a)
  , PConstant a
  ) =>
  PConstant (Foo a) 
``` 

One idea to simplify is with Constraint synonyms which provide a shorthand for the constraints:

```haskell
type PConstantable :: Constraint
type PConstantable a =  (a ~ (PLifted (PConstant a)), PUnsafeLiftDecl (PConstanted a), PConstant a)

type PLiftable a :: Constraint
type PLiftable a = (a ~ (PConstanted (PLifted a)), PUnsafeLiftDecl a)
```

peter-mlabs:
As an example of using `Data` instead, we  can have

```haskell
data Component a = Component
  {cValue :: Plutus.Value,
   cDatum :: a 
  }

data PComponent (a :: PType)  (s :: S)
  = PComponent (Term s (PDataRecord '["_0" ':= PValue, "_1" ':= a]))
  deriving stock (GHC.Generic)
  deriving anyclass (SOP.Generic, PIsDataRepr)
  deriving
    (PMatch, PIsData)
    via PIsDataReprInstances (PComponent a)

instance 
  forall a . 
  ( a ~ (PConstanted (PLifted a))
  , PUnsafeLiftDecl a
  ,  Ledger.FromData (Component (PLifted a))
  , Ledger.ToData (Component (PLifted a))
  ) =>
  PUnsafeLiftDecl (PComponent a) where 
  type PLifted (PComponent a) = Component (PLifted a) 


deriving via
  (DerivePConstantViaData  
    (Component a)
    (PComponent (PConstanted a)))
  instance 
  (a ~ (PLifted (PConstanted a))
  , PUnsafeLiftDecl (PConstanted a)
  , PConstant a
[...]

PR #354, opened 2022-03-03
Peter mlabs/issue286

peter-mlabs:
Added the helper type synonyms `PConstantableData`, `PLiftableData`, and `PConstantable`, along with some TODO'd haddock docs (mostly cribbed from the guide).

Let me know if anything seems amiss.


TotallyNotChase (changes requested):
Looks great! I think `PLift p` should also just be an alias to `(p ~ (PConstanted (PLifted p)), PUnsafeLiftDecl p)`.

Also I believe you're missing a `./bin/format` run

peter-mlabs:
> Looks great! I think `PLift p` should also just be an alias to `(p ~ (PConstanted (PLifted p)), PUnsafeLiftDecl p)`.

Sorry, I'm not sure I understand.  In #286 Las suggested using `PLift p` to accomplish this. Are you saying that I should declare a 

```
type PLift :: Type -> Constraint
type PLift = (p ~ (PConstanted (PLifted p)), PUnsafeLiftDecl p)
```
instead of having it alias `PUnsafeLiftDecl`? 

Or that I should inline that [here](https://github.com/peter-mlabs/plutarch/blob/68701201c1491fb4533bb2bdc18cd6063fee439a/Plutarch/DataRepr/Internal.hs#L431)?

TotallyNotChase:
I think `PLift` should just be an alias to the 2 constraints together.

peter-mlabs:
Done. Thanks!

Plutarch/Lift.hs
```diff
@@ -57,7 +66,8 @@ The Haskell type is determined by `PLifted p`.
 
 This typeclass is closely tied with 'PConstant'.
 -}
-type PLift = PUnsafeLiftDecl
+type PLift :: PType -> Constraint
+type PLift p = (p ~ (PConstanted (PLifted p)), PUnsafeLiftDecl p)
```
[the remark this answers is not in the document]
peter-mlabs: @TotallyNotChase, thoughts?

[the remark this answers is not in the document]
TotallyNotChase: oh I didn't notice that constraint is actually [bundled with `PUnsafeLiftDecl` already](https://github.com/peter-mlabs/plutarch/blob/68701201c1491fb4533bb2bdc18cd6063fee439a/Plutarch/Lift.hs#L47). That's quite nice actually - yes, it's redundant then.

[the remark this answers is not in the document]
TotallyNotChase: @L-as Btw a linter would catch the redundant parens stuff. Maybe we should have a `lint.sh`

Plutarch/Lift.hs
```diff
@@ -112,8 +126,54 @@ instance
...
+- The lifted type of @a@ has a @PUnsafeLiftDecl@ instance.
+- There is type equality between @a@ and @PLifted (PConstanted a)@.
+
+These constraints are sufficient to derive a @PConstant@ instance for the newtype.
+
+For deriving @PConstant@ for a wrapped type represented in UPLC as @Data@, see
+@DerivePConstantViaData@.
+-}
+type PConstantable :: Type -> Constraint
+type PConstantable a = (a ~ (PLifted (PConstanted a)), PUnsafeLiftDecl (PConstanted a), PConstant a)
```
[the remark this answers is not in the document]
L-as: We can't.

L-as:
Your changes have broken some existing code

L-as:
I'm honestly not sure how, try to see if you can get the tests working

peter-mlabs:
Ah, I didn't know how to run the tests and forgot to ask :sweat_smile: 
I've updated the DEVGUIDE with a point to `plutarch-test` and will try to track down the issues.

Plutarch/Lift.hs
```diff
@@ -41,7 +50,7 @@ Laws:
    in UPLC (via the 'UntypedPlutusCore.Constant' constructor) is a valid @p@.
 -}
 class (PConstant (PLifted p), PConstanted (PLifted p) ~ p) => PUnsafeLiftDecl (p :: PType) where
-  type PLifted p :: Type
+  type PLifted p = (r :: Type) | r -> p
```
TotallyNotChase, line 53: The tests were erroring out since they had trouble figuring out the transitive `PLifted p` result in the newly added `PLift` type alias- `type PLift p = (p ~ PConstanted (PLifted p), PUnsafeLiftDecl p)` - only seemed to happen in `pconstant` usages.

This type family dependency fixes that and helps GHC immediately figure out the transitive result.

I honestly think we should sprinkle in type family deps everywhere, but especially across the `PConstant`/`PLift` architecture - I quite like type fam deps. It should be theoretically possible but our abstract derivers make it extremely complicated to _not_ violate the injectivity principles if adding type fam deps to the `PConstant` associated types.

TotallyNotChase:
This can be merged now.

Plutarch/Lift.hs
L-as: I hadn't realised this was possible for associated type families! This seems quite powerful.

Plutarch/Lift.hs
```diff
@@ -70,7 +79,8 @@ The Haskell type is determined by `PLifted p`.
 
 This typeclass is closely tied with 'PConstant'.
 -}
-type PLift = PUnsafeLiftDecl
+type PLift :: PType -> Constraint
+type PLift p = (p ~ PConstanted (PLifted p), PUnsafeLiftDecl p)
```
L-as, line 83: This is redundant as noted before

Plutarch/Lift.hs
```diff
@@ -125,8 +139,54 @@ instance
...
+- The lifted type of @a@ has a @PUnsafeLiftDecl@ instance.
+- There is type equality between @a@ and @PLifted (PConstanted a)@.
+
+These constraints are sufficient to derive a @PConstant@ instance for the newtype.
+
+For deriving @PConstant@ for a wrapped type represented in UPLC as @Data@, see
+@DerivePConstantViaData@.
+-}
+type PConstantable :: Type -> Constraint
+type PConstantable a = (a ~ PLifted (PConstanted a), PConstant a)
```
L-as, line 189: We should really name `PConstant` `PConstantDecl` and call this `PConstant`, otherwise people will accidentally use `PConstant` instead of `PConstantable`.

L-as (changes requested):
Thanks @TotallyNotChase !

Plutarch/Lift.hs
TotallyNotChase: that's a good idea - though it's kind of a big breaking change. I'm all for it though - will add it to the changelog.

Plutarch/Lift.hs
TotallyNotChase: whoop - removed

Plutarch/Lift.hs
TotallyNotChase: Done! And also simplified a bunch of usage sites with these type aliases - actually a huge improvement! I'm very happy with how it came out. Thanks again @peter-mlabs!

Plutarch/Lift.hs
```diff
@@ -70,7 +84,8 @@ The Haskell type is determined by `PLifted p`.
 
 This typeclass is closely tied with 'PConstant'.
 -}
-type PLift = PUnsafeLiftDecl
+type PLift :: PType -> Constraint
+type PLift p = PUnsafeLiftDecl p
```
L-as, line 88: You don't need this change anymore. You can keep the kind signature though I suppose.

L-as (changes requested):
Looks mostly great!

Plutarch/Lift.hs
TotallyNotChase: I like standalone type signatures :)
Removed the change in definition.

L-as:
Running CI

L-as (approved):
LGTM

peter-mlabs:
Thanks for wrapping this up @TotallyNotChase and @L-as <3

PR #351, opened 2022-03-03
Add plutarch-extra scaffold

MangoIV:
## Changes 

- set up plutarch-extra
- add a new folder for testing plutarch-extra specific stuff

PR #355, opened 2022-03-04
Add PThese, PTheseData

jhodgdev:
This PR implements Plutarch analogues for [Data.These](https://hackage.haskell.org/package/these). Would it be possible to make `PTheseData` an instance of `PUnsafeLiftDecl`? I assume it would be, as `plutus-tx` appears to support `These`. I am unsure as to how to find the underlying PLC implementation.

### DIFF
--- a/Plutarch/Api/V1/These.hs
+++ b/Plutarch/Api/V1/These.hs
@@ -1,0 +1,61 @@
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Plutarch.Api.V1.These (PTheseData (..)) where
+
+import qualified GHC.Generics as GHC
+import Generics.SOP
+import Plutarch.DataRepr (PIsDataReprInstances (PIsDataReprInstances))
+import Plutarch.Lift (
+  PConstantRepr,
+  PConstanted,
+  PLifted,
+  PUnsafeLiftDecl,
+  pconstantFromRepr, 
+  pconstantToRepr
+ )
+import qualified Plutus.V1.Ledger.Api as Plutus
+import qualified PlutusTx.These as PlutusThese
+import Plutarch.Prelude
+
+data PTheseData (a :: PType) (b :: PType) (s :: S)
+  = PDThis (Term s (PDataRecord '["_0" ':= a]))
+  | PDThat (Term s (PDataRecord '["_0" ':= b]))
+  | PDThese (Term s (PDataRecord '["_0" ':= a, "_1" ':= b]))
+  deriving stock (GHC.Generic)
+  deriving anyclass (Generic, PIsDataRepr)
+  deriving
+    (PlutusType, PIsData)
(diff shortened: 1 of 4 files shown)


emiflake:
> In Plutarch, we won't be going that route for value unions since it's rather inefficient.

I'm curious what's the approach that will be used instead? I know in the case of many Value functions the `These` can be eliminated with inline explicit branching, is this what you're envisioning?

TotallyNotChase:
AFAICT if `PValue`s (and `PMap`s) are required to be sorted at all times - there isn't a need for `These`.

emiflake:
But in the case of, say, `geq`, we would require traversing the full Value anyway, right? So in those cases the `These` overhead is a constant added cost, but not an algorithmic one.

TotallyNotChase:
Sorry, I'm not sure what you mean there. For `checkBinRel` in general, I also imagine iterating through both lists at the same time depending on the ordering of the keys would do:

```hs
checkBinRel :: Ord k => (Int -> Int -> Bool) -> [(k, Int)] -> [(k, Int)] -> Bool
checkBinRel f = inner
  where
    inner [] [] = True
    inner [] l = all (f 0 . snd ) l
    inner l [] = all (flip f 0 . snd) l
    inner a@((k1, v1):xs) b@((k2, v2):ys)
      | k1 < k2 = f v1 0 && inner xs b
      | k1 == k2 = f v1 v2 && inner xs ys
      | otherwise = f 0 v2 && inner a ys
```
I _imagine_ it'd have the same behavior assuming both the values/maps are sorted but I might be missing something here.

Plutarch/Api/V1/These.hs
```diff
@@ -0,0 +1,61 @@
...
+  , Plutus.FromData a
+  , PConstant a
+  , PLifted (PConstanted b) ~ b
+  , Plutus.FromData b
+  , Plutus.ToData b
+  , PConstant b
+  ) =>
+  PConstant (PlutusThese.These a b)
+  where
+  type PConstantRepr (PlutusThese.These a b) = [(Plutus.Data, Plutus.Data)]
```
t1lde, line 58: If the intention is to match the Plutus These `ToData/FromData` repr,  you can derive via something like 
```haskell
deriving 
  via PConstantViaData (These a b) (PTheseData (PConstanted a) (PConstanted b))
  instance (...) => PConstant (These a b)
```

t1lde:
> I know in the case of many Value functions the `These` can be eliminated with inline explicit branching, is this what you're envisioning?

For Value utils I'd say eliminating the intermediate data would be right approach. & if necessary, retaining the scott-encoded `PThese` to return in some cases. `PThese` could be used at the haskell-level to have Value util implementations that look a little more like the plutus-tx ones.

L-as (changes requested):
This could fit into `plutarch-extra`.

L-as:
Wait for that to be merged if you still want this. #351

Issue #344, opened 2022-03-02
Make invariants for `PMap` and `PValue` explicit


L-as:
- Map entries should be sorted.
- Value must not contain entries of zero amount.

L-as:
Related:
- #303
- #326

blamario:
No function in [PlutusTx.AssocMap](https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/src/PlutusTx.AssocMap.html#toList) enforces ordering nor requires the `Ord k` constraint. They also seem very inefficient for unrelated reasons, but that's beside the point. The point is we have to work with `Value`s as we get them, which is unsorted, not as we wish they were in Plutus.

L-as:
I'm not sure what you mean? If you mean putting `Value` into datums, then that would be covered by a `PTryFrom` implementation.

blamario:
No, I meant actual values in output UTxOs in the `ScriptContext` given to the validator. Keep in mind that the `ScriptContext` can be constructed by multiple testing libraries as well as the actual Cardano node. Do we trust all of them to normalize the script context? Just today there's this issue reported in Slack https://mlabs-corp.slack.com/archives/C02QQHLFB6Y/p1648459396901379 where `plutus-simple-model` doesn't filter out the zero token quantities. Speaking of which, the PlutusTx implementation of equality [doesn't expect](https://playground.plutus.iohkdev.io/doc/haddock/plutus-ledger-api/html/src/Plutus.V1.Ledger.Value.html#eq) the zeroes to be filtered out.

Furthermore, suppose in some future ~release~commit of `cardano-node` or wherever this logic is, the values in `ScriptContext` UTxOs are *not* properly normalized, either through oversight or by design. A PlutusTx-compiled validator won't be affected, Plutarch will be. I wouldn't expect IOHK to accept the blame for creating a new vulnerability in the Plutarch-compiled validator scripts that would have been *already present* on-chain. Their reference implementation is safe after all.

This situation sucks from the performance point of view, but unless we have a written-down guarantee that UTxO values are always going to be normalized, the only sane way to ensure Plutarch's forward-compatible safety on-chain is to replicate the behaviour of PlutusTx. Not necessarily the implementation, mind you. The linked PlutusTx equality comparison starts by converting the two maps into a list of `These`.

blamario:
BTW the  `instance Eq Map` (as opposed to `Value`) doesn't even sort:
```
[...]

PR #361, opened 2022-03-05
Issue #343: add instance for `PEq` for `PValue`

### LINKED ISSUE #343: `PValue` not an instance of `PEq`

MangoIV:
### Changes 
- this resolves #343  

### Tasks 
- [x] change the `PEq` instance of `PValue` to use the instance of `PData`


TotallyNotChase (changes requested):
I think we decided on using `EqualsData` rather than iterative equality.

This requires that we enforce certain other invariants as well. See #344

Plutarch/Api/V1/Value.hs
```diff
@@ -41,6 +42,9 @@ newtype PValue (s :: S) = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName
     (PlutusType, PIsData)
     via (DerivePNewtype PValue (PMap PCurrencySymbol (PMap PTokenName PInteger)))
 
+instance PEq PValue where
+  a #== b = pforgetData (pdata a) #== pforgetData (pdata b)
```
TotallyNotChase, line 46: `pdata a #== pdata b` will work here too btw

TotallyNotChase (approved):

Plutarch/Api/V1/Value.hs
L-as: I don't think you've fixed this

L-as (changes requested):
Thanks!

Plutarch/Api/V1/Value.hs
MangoIV: whoops

L-as:
Can you make this into one commit with a descriptive commit message?

L-as:
Specifically, one that isn't just `fix:`

MangoIV:
It is! It's just multiline, as always ;D

L-as:
@MangoIV the issue with multiline commit messages is that most tools don't expand them, and having a helpful first line is very nice

Plutarch/Api/V1/Value.hs
```diff
@@ -41,6 +41,9 @@ newtype PValue (s :: S) = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName
     (PlutusType, PIsData)
     via (DerivePNewtype PValue (PMap PCurrencySymbol (PMap PTokenName PInteger)))
 
+instance PEq PValue where
```
L-as, line 44: Now that I think about it, we sehould use the same `PEq` for `PBuiltinList (PAsData a)`, and then we can just use newtype derive here.

L-as (changes requested):
thanks

MangoIV:
This is ready afaict

TotallyNotChase:
it'd also be very nice to have a small golden test for the equality - making sure it is indeed using data equality and such.

Plutarch/Api/V1/Value.hs
MangoIV: Correction:
@L-as I don't think that this instance will help because 
```haskell
type PBuiltinMap = (PBuiltinList (PBuiltinPair (PAsData a) (PAsData b))```
and hence it picks the usual instance for `PEq` that isn't `PData`.

L-as (approved):
LGTM

L-as (changes requested):
Actually, can't you fix the instance for `PBuiltinMap` or is that not possible?

MangoIV:
> Actually, can't you fix the instance for `PBuiltinMap` or is that not possible?

Excuse me, I'm not sure what you mean?

L-as:
Depends on https://github.com/input-output-hk/plutus/issues/4476

TotallyNotChase:
@L-as I think you meant 4509?

L-as:
Indeed!

L-as:
@MangoIV We have `PEq` for `PValue` now, what we need:
- Make `PMap`'s `PEq` use `pdata`. 
- Remove `PBuiltinMap` if possible or make it a newtype, a type alias is confusing.

L-as:
This is no longer needed. Thanks.

PR #363, opened 2022-03-07
Add a few new sections to the guide + fixes

TotallyNotChase:
- Fix `por` and `#||` definition in "Hoisting operators" section
- Fix `pmatch'` and `pcon'` type signature in PlutusType section
- Fix a typo in PIsDataRepr section
- Explicitly note that `PDataRecord` fields should have a `PIsData` instance (and that relates to `PBuiltinList`).
- Add note about `newtype`s in the Isomorphism between PlutusTx and Plutarch data encoded types section
- Add information on `pto` in the PlutusType section
- Code example of hoisting in the primary "Hoisting" section.
- Add note on when it may be a good idea to `plet` fields of scott encoded types prior to construction with `pcon`.
- Add new subsection under "Tricks", encouraging users to pattern match on the `pmatch` result immediately.
- Add section on `plet`ing lambdas rather than hoisting them, when it's not possible to hoist them.

### DIFF
--- a/docs/Concepts/Hoisting.md
+++ b/docs/Concepts/Hoisting.md
@@ -34,10 +34,10 @@ You can factor out most of the logic to a Plutarch level function, and apply tha
 
 ```hs
 (#||) :: Term s PBool -> Term s PBool -> Term s PBool
-x #|| y = por # x # pdelay y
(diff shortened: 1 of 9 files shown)


L-as (approved):
LGTM

PR #364, opened 2022-03-07
add mising deriving to guide

peter-mlabs:


TotallyNotChase:
Actually, you can only derive `PDataFields` for newtypes to `PDataRecords`. That is, single constructor product types. It won't work for sum types since you can't access the fields directly without first determining the constructor.

peter-mlabs:
Ah okay. I think my confusion was because I didn't have a `newtype`, I had:

```
data Redeemer (s :: S)
  = Redeemer
      ( Term
          s
          ( PDataRecord
              '[ "aDatum" ':= PInteger
               , "cDatum" ':= PByteString
               ]
          )
      )
  deriving stock (GHC.Generic)
  deriving anyclass (SOP.Generic, PIsDataRepr)
  deriving
    (PlutusType, PIsData, PDataFields)
    via PIsDataReprInstances (Redeemer)
```

Which only allows field access if `PDataFields` is derived and throws a type error otherwise.

L-as:
Whether it's a newtype or not is not what's relevant here. You can't have field access when it's not a record.

PR #370, opened 2022-03-09
Fix staging CI after #352

srid:
I think this is the first time I see git merge doing the wrong thing.

#352


L-as (approved):
It happens all the time. Git is fundamentally broken.

PR #371, opened 2022-03-10
Optimized `PEq` and `POrd` derivation for data encoded types

### LINKED ISSUE #366: `POrd` for Data-encoded types
image

We can avoid doing `pmatch` and use an optimised implementation

TotallyNotChase:
Resolves #366 

I'm also fairly confident about a generalized `pmatchN` specifically for data encoded types that could potentially perform better than the current multiple `pmatch` case. That's not included in this PR though.

This PR intentionally doesn't hoist anything. I'm not totally sure about the best spots to hoist since introducing a `plam` may end up being a trade off. I do intend to hoist most of it - I'm just not sure which places are the best to hoist.

I have tests and benchmarks for these already, just need to figure out how to port them into this repo. Numbers on the optimized version vs normal `pmatch` version will be posted later today.

TODO:
- [x] Add tests
- [x] Update guide (need to merge `master` in `staging`)


Plutarch/DataRepr/Internal.hs
```diff
@@ -121,6 +124,35 @@ instance PlutusType (PDataRecord '[]) where
...
+  _ #<= _ = pconstant True
+  _ #< _ = pconstant False
+
+instance (POrd x, PIsData x) => POrd (PDataRecord '[label ':= x]) where
+  l1 #< l2 = unTermCont $ do
+    PDCons x _ <- tcont $ pmatch l1
+    PDCons y _ <- tcont $ pmatch l2
+
+    pure $ pfromData x #< pfromData y
+  l1 #<= l2 = pdToBuiltin l1 #== pdToBuiltin l1 #|| l1 #< l2
```
TotallyNotChase, line 143: I have yet to test whether this pattern of "`#==` or `#<`" is better for `#<=` impl or if having a separate explicit definition similar to `#<=` is better.

Plutarch/DataRepr/Internal.hs
```diff
@@ -139,6 +171,10 @@ pdcons = punsafeCoerce $ pcons @PBuiltinList @PData
 pdnil :: Term s (PDataRecord '[])
 pdnil = punsafeCoerce $ pnil @PBuiltinList @PData
 
+-- | Convert a 'PDataRecord' into a builtin list of data values, losing type information in the process.
+pdToBuiltin :: Term s (PDataRecord xs) -> Term s (PBuiltinList PData)
```
L-as, line 175: This is just `pto` AFAICT?

Plutarch/DataRepr/Internal.hs
L-as: An explicit implementation here can be more efficient.

Plutarch/DataRepr/Internal.hs
```diff
@@ -121,6 +124,35 @@ instance PlutusType (PDataRecord '[]) where
...
+instance (POrd x, PIsData x, POrd (PDataRecord (x' ': xs))) => POrd (PDataRecord ((label ':= x) ': x' ': xs)) where
+  l1 #< l2 = unTermCont $ do
+    PDCons x xs <- tcont $ pmatch l1
+    PDCons y ys <- tcont $ pmatch l2
+
+    a <- tcont . plet $ pfromData x
+    b <- tcont . plet $ pfromData y
+
+    pure $ pif (a #< b) (pconstant True) $ pif (a #== b) (xs #< ys) $ pconstant False
+  l1 #<= l2 = pdToBuiltin l1 #== pdToBuiltin l2 #|| l1 #< l2
```
L-as, line 154: Explicit implementation here too would also help.

Plutarch/DataRepr/Internal.hs
```diff
@@ -178,6 +214,16 @@ instance PIsData (PDataSum defs) where
   pdata = punsafeCoerce
   pfromData = punsafeCoerce
 
+instance PEq (PDataSum defs) where
+  x #== y = pdata x #== pdata y
+
+instance MkLtReprHandler defs => POrd (PDataSum defs) where
```
L-as, line 220: You can make a more efficient instance for sums with only one constructor.

Plutarch/DataRepr/Internal.hs
```diff
@@ -344,8 +385,77 @@ instance
...
+        -- Left arg's constructor id is greater, no need to continue.
+        $ pconstant False
+  where
+    applyHandlers :: Term s (PBuiltinList PData) -> Term s (PBuiltinList PData) -> LTReprHandlers defs s -> [Term s PBool]
+    applyHandlers _ _ LTRHNil = []
+    applyHandlers args1 args2 (LTRHCons handler rest) =
+      handler (punsafeCoerce args1) (punsafeCoerce args2) :
+      applyHandlers args1 args2 rest
+
+    go ::
```
L-as, line 426: You don't optimise out any shared cases. This can happen if a sum type has two identical constructors, e.g. `PMaybeData PData PData`. Not sure this is worth optimising for.

L-as: You can see how it's done in `pmatchDataSum`.

Plutarch/Bool.hs
```diff
@@ -83,13 +84,20 @@ class PEq t where
...
-class POrd t where
+class PEq t => POrd t where
   (#<=) :: Term s t -> Term s t -> Term s PBool
   (#<) :: Term s t -> Term s t -> Term s PBool
 
 infix 4 #<=
 infix 4 #<
 
+instance PEq PBool where
+  x #== y' = plet y' $ \y -> pif' # x # y #$ pnot # y
```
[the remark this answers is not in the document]
L-as: You should always `plet` something when you use something twice. If it's a variable reference it will be eliminated as an optimisation. There are few cases where it's better not to `let` in practice.

Plutarch/Bool.hs
```diff
@@ -83,13 +84,20 @@ class PEq t where
...
+
+instance POrd PBool where
+  x #< y = pif' # x # pconstant False # y
```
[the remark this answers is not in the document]
L-as: Booleans typically have lazy behavior, so maybe?

Plutarch/DataRepr/Internal.hs
```diff
@@ -121,6 +124,35 @@ instance PlutusType (PDataRecord '[]) where
...
   pmatch' _ f = f PDNil
 
+-- | This uses data equality. 'PEq' instances of elements don't make any difference.
+instance PEq (PDataRecord xs) where
+  x #== y = pdToBuiltin x #== pdToBuiltin y
+
+-- Lexicographic ordering based 'Ord' instances for 'PDataRecord'.
+
+instance POrd (PDataRecord '[]) where
+  _ #<= _ = pconstant True
```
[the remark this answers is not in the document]
L-as: TBH this is probably fine.

Plutarch/DataRepr/Internal.hs
```diff
@@ -121,6 +124,35 @@ instance PlutusType (PDataRecord '[]) where
...
+  _ #< _ = pconstant False
+
+instance (POrd x, PIsData x) => POrd (PDataRecord '[label ':= x]) where
+  l1 #< l2 = unTermCont $ do
```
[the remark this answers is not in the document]
L-as: What if you use `#<` on the same type a lot?

L-as (changes requested):
Looks mostly good

TotallyNotChase:
I'm wondering what to do about tracking performance. I've been using the property tests to eyeball the performance from multiple runs but I don't know if we can do something similar here. It'd be nice to generate inputs to benchmark and take an average or something of that sort - the generation done by quick check is surprisingly predictable tbh.

Thoughts @srid?

Plutarch/DataRepr/Internal.hs
TotallyNotChase: Yes, but for some reason I seem to be getting ambiguous type variable on that `b` in `PInner` whenever using `pto` in some of these places: `pto x == pto y`, `pdToBuiltin = pto` doesn't work either here. What am I missing here?

Plutarch/DataRepr/Internal.hs
TotallyNotChase: I think that case should be handled by [`pmatchLT`](https://github.com/Plutonomicon/plutarch/pull/371/files#diff-850ef084fc92b4778569ed7b35931834f5de4efd686a7d6fd242ec21d19c503fR397) regardless

Plutarch/Bool.hs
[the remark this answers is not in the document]
TotallyNotChase: Haskell's own `Ord` instances (including those for booleans) evaluate both arguments though, `False < undefined = undefined`. 

This is so hard to choose and be consistent over across the codebase.

TotallyNotChase:
There's an interesting case when you keep using `pmatch`, but compare `PDataRecord`s using the new `POrd` instance in this PR-
```hs
data PBar s
  = PBar (Term s (PDataRecord '[ "_0" ':= PInteger ]))
  | PBaz (Term s (PDataRecord '[ "_0" ':= PByteString ]))
  deriving stock (GHC.Generic)
  deriving anyclass (Generic, PIsDataRepr)
  deriving (PlutusType, PIsData) via (PIsDataReprInstances PBar)

instance POrd PBar where
  x #< y = unTermCont $ do
    a <- pmatchC x
    b <- pmatchC y
    pure $ case (a, b) of
      (PBar _, PBaz _)         -> pconstant True
      (PBar flds1, PBar flds2) -> flds1 #< flds2
      (PBaz _, PBar _)         -> pconstant False
      (PBaz flds1, PBaz flds2) -> flds1 #< flds2
```

For some reason, even this one has *a few cases* where it performs _better_ than the "no pmatch" approach (i.e `LTReprHandler` approach): `instance POrd PBar where x #< y = pto x #< pto y`

```hs
> sample testBar
(pmatch: CPU{3438586} Mem{7896} Size{130},optimized: CPU{2793856} Mem{6532} Size{112})
(pmatch: CPU{4576049} Mem{9225} Size{130},optimized: CPU{5024194} Mem{10127} Size{112})
(pmatch: CPU{4576049} Mem{9225} Size{130},optimized: CPU{5024194} Mem{10127} Size{112})
(pmatch: CPU{4576049} Mem{9225} Size{130},optimized: CPU{5024194} Mem{10127} Size{112})
(pmatch: CPU{4499709} Mem{9225} Size{130},optimized: CPU{4947854} Mem{10127} Size{112})
(pmatch: CPU{3438586} Mem{7896} Size{130},optimized: CPU{2793856} Mem{6532} Size{112})
(pmatch: CPU{4576049} Mem{9225} Size{130},optimized: CPU{5113513} Mem{10427} Size{115})
(pmatch: CPU{3438586} Mem{7896} Size{130},optimized: CPU{2298711} Mem{5330} Size{112})
(pmatch: CPU{3438586} Mem{7896} Size{130},optimized: CPU{2793856} Mem{6532} Size{112})
(pmatch: CPU{4576049} Mem{9225} Size{130},optimized: CPU{5024194} Mem{10127} Size{112})
(pmatch: CPU{3438586} Mem{7896} Size{130},optimized: CPU{2793856} Mem{6532} Size{112})
```

In all the cases where the performance is worse, the constructors on both match (and thus, it proceeds with field comparison).

Plutarch/DataRepr/Internal.hs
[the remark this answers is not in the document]
TotallyNotChase: there's a delicate balance here that I'm not totally sure how to handle. I think hoisting should be done at the `PDataSum` level since there won't be any recursion there. If I hoist it for `PDataRecord`s, for every single field a new `plam` would be introduced. Right now, it all just nicely unrolls into a bunch of field comparisons - as if you wrote it by hand.

At the same time, hoisting for larger `PDataRecord`s (say, 5 elements? yet to determine) might actually be good - it could potentially have a sort of "memoization" effect, where multiple comparisons can reference one function that compares 5 fields at once (no inner `plam`s for each field). This sort of batch hoisting seems good in my intuition - but all of this needs to be tested.

Plutarch/DataRepr/Internal.hs
TotallyNotChase: I'll give explicit impls a try and see how that fares. Do you think explicit impls are necessary for only `PDataRecord`s or could it be beneficial at the `PDataSum` level too?

The interesting thing to think about is that the current implementation has a happy path when things are equal - this won't be the case for an iterative, explicit impl. But it could still liberate perf in non-empty case.

L-as:
> I'm wondering what to do about tracking performance. I've been using the property tests to eyeball the performance from multiple runs but I don't know if we can do something similar here. It'd be nice to generate inputs to benchmark and take an average or something of that sort - the generation done by quick check is surprisingly predictable tbh.

We shouldn't be using quickcheck for this. Use the existing API but use a seed to pseudorandomly generate tests.

Plutarch/DataRepr/Internal.hs
L-as: You're completely duplicating parts of the logic in the same file.

TotallyNotChase:
The existing API already supports quickcheck quite nicely though. And we're already using some property tests here and there. Do we really have to handroll a random input generator?

L-as:
QuickCheck isn't deterministic and isn't fit for goldens really

L-as:
Though TBF QuickCheck can also use a seed I think

Plutarch/DataRepr/Internal.hs
```diff
@@ -344,8 +385,91 @@ instance
...
       mkSOP = SOP . mkSum @_ @n @pcode
 
-pasDataSum :: PIsDataRepr a => Term s a -> Term s (PDataSum (PIsDataReprRepr a))
-pasDataSum = punsafeCoerce
+data LTReprHandlers (defs :: [[PLabeledType]]) (s :: S) where
+  LTRHNil :: LTReprHandlers '[] s
+  LTRHCons ::
+    (Term s (PDataRecord def) -> Term s (PDataRecord def) -> Term s PBool) ->
+    LTReprHandlers defs s ->
+    LTReprHandlers (def : defs) s
```
srid, line 393: This is isomorphic to `NS` from sop-core; could we use that instead?

```haskell
type LTReprHandlers defs s = NS (LTRepr s) defs
newtype LTRepr s def = LTRepr (Term s (PDataRecord def) -> Term s (PDataRecord def) -> Term s PBool)
```
```

File: plutarch-test/plutarch-base/Plutarch/POrdSpec.hs [hunk 11]
```diff
@@ -0,0 +1,95 @@
...
+import Plutarch.Prelude
+import Plutarch.SpecTypes (PTriplet)
+
+spec :: Spec
+spec = do
+  describe "ord.pisdata" $ do
+    propertySet @PBool "PBool"
+    propertySet @(PMaybeData PInteger) "PMaybeData PInteger"
+    propertySet @(PTriplet PInteger) "PMaybeData PInteger"
+    propertySet @PAddress' "PAddress"
```
srid, line 37: Could we have some golden tests too?

srid:
> I'm wondering what to do about tracking performance. I've been using the property tests to eyeball the performance from multiple runs but I don't know if we can do something similar here. It'd be nice to generate inputs to benchmark and take an average or something of that sort - the generation done by quick check is surprisingly predictable tbh.
> 
> Thoughts @srid?

Sorry I missed this mention. The approach I've been taking is to hand-write the "examples" in `pgoldenSpec`. See [`ShowSpec.hs`](https://github.com/Plutonomicon/plutarch/blob/843bc9d4de55e1c58900ce143302fcea80e8bea5/plutarch-test/plutarch-base/Plutarch/ShowSpec.hs) for an example. This is actually beneficial in one way: we end up automatically providing examples for Plutarch functionality without intending to write them.

I don't see why we should mix property and golden tests, instead of keeping them separate.

TotallyNotChase:
I agree, I realized later that it's better to just have some golden tests alongside the property tests rather than mix them up.

Plutarch/DataRepr/Internal.hs
TotallyNotChase: something like
```hs
foo :: Term s (PDataRecord l) -> Term s (PBuiltinList PData)
foo x = pto x
```
doesn't work - particularly, it fails with an "type variable b0 is ambiguous" error (the `b0` is the `b` for `PInner`). I think GHC is refusing to expand the `PInner` type family at the presence of that `l` type variable. When `l` is replaced with a concrete type rather than a type variable-
```hs
foo :: Term s (PDataRecord '[]) -> Term s (PBuiltinList PData)
foo x = pto x
```
it works fine.

It looks like `pto` is not going to work with types that take in type variables? Potentially worth making an issue and tackle separately?

L-as: Yep, this is a bug in the definition of the PlutusType instance. We should fix this if possible.

TotallyNotChase: I actually have no idea how, hints?

Plutarch/DataRepr/Internal.hs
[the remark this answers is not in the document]
TotallyNotChase: decided to use `phoistAcyclic` at the `PDataSum` level - it should be optimal for multi-use cases. Slightly worse for single worse case due to the lambda application

Plutarch/DataRepr/Internal.hs
L-as: we need to have an instance for all PDataRecords somehow

TotallyNotChase: yeah, the wall I've been hitting on that is essentially:

```hs
Couldn't match type ‘labeledX’
               with ‘PLabelOf labeledX ':= PUnLabel labeledX’
```

Minimal repro:

```hs
type PUnLabel :: PLabeledType -> PType
type family PUnLabel a where
  PUnLabel (name ':= a) = a

type PLabelOf :: PLabeledType -> Symbol
type family PLabelOf a where
  PLabelOf (name ':= a) = name
```

GHC doesn't seem to realize `labeledX` will always be equal to `PLabelOf labeledX ':= PUnLabel labeledX` - is there a way to assert that? (other than having it as a constraint in the instance head)

Plutarch/DataRepr/Internal.hs
srid: Oops, I meant `NP`.

Plutarch/DataRepr/Internal.hs
blamario: This constraint meets your stated requirements:

    instance (labeledX ~ PLabelOf labeledX ':= PUnLabel labeledX) => Thingy (labeledX ': xs) where

TotallyNotChase: yeah but I believe I need this to work for all `x ': xs`, or even all `xs` (both `'[]` and `x ': xs`), without any constraints in the instance head. The constraint will need to be propagated all the way and it just seems a bit unergonomic.

L-as (approved):
LGTM

PR #372, opened 2022-03-10
Some `plam` reductions

TotallyNotChase:
This should benefit #371 

I think there may be more places to optimize. Need to check.


Plutarch/Bool.hs
```diff
@@ -111,7 +111,7 @@ pif b case_true case_false = pmatch b $ \case
 
 -- | Boolean negation for 'PBool' terms.
 pnot :: Term s (PBool :--> PBool)
-pnot = phoistAcyclic $ plam $ \x -> pif x (pcon PFalse) $ pcon PTrue
+pnot = phoistAcyclic $ plam $ \x -> pif' # x # pconstant False # pconstant True
```
L-as, line 114: Why use `pconstant` when it's unnecessarily longer? Just use `pcon PFalse` and such

TotallyNotChase: Habit, but yeah - changed.

TotallyNotChase:
Seems like those were all the reductions I was able to spot.

L-as (approved):
LGTM

PR #378, opened 2022-03-15
Added FFI test suite

blamario:


plutarch-test/plutarch-base/Plutarch/FFISpec.hs
```diff
@@ -0,0 +1,382 @@
...
+spec :: Spec
+spec = do
+  describe "Test.PlutarchFFI" $ do
+    describe "Simple types" $ do
+      it "integer literal" $
+        printCode $$(PlutusTx.compile [||42 :: Integer||]) @?= "(program 1.0.0 42)"
+      it "PlutusTx integer function" $
+        printCode double @?= "(program 1.0.0 (\\i0 -> multiplyInteger 2 i1))"
+      it "Plutarch integer function" $
+        printTerm (plam $ \(x :: Term _ PInteger) -> 2 Prelude.* x) @?= "(program 1.0.0 (\\i0 -> multiplyInteger 2 i1))"
```
srid, line 233: You want to use`pgoldenSpec` here. See https://github.com/Plutonomicon/plutarch/blob/f5327118b2bb52e0ef34fc514615a76404b5bb5c/plutarch-test/plutarch-base/Plutarch/RecSpec.hs#L212

blamario: I wanted to, but the majority of the tests don't fit.  The `@|` operator expects a `ClosedTerm`, not `CompiledCode`. The tests here also apply Shrinker. We could look into generalizing `pgoldenSpec` but it's probably not worth it for this module alone.

L-as (approved):
LGTM

blamario:
Yeah, that's what I said in Discourse:
> The major problem now is that `sydtest-discover` is totalitarian. The current hacky workaround only helps the CI, doesn't work inside `nix develop`.

PR #383, opened 2022-03-16
Optimize `PEq` generic instances

### LINKED ISSUE #368: Remove redundant `#&&` in `PEq` generic instance
https://github.com/Plutonomicon/plutarch/blob/5ce1d3dce163818af4e98a301f09478cc7bf0f6f/Plutarch/Bool.hs#L170-L171

For example, a single product constructor will generate `pcon PTrue #&& p1` instead of just `p1`.

srid:
... by removing the redundant `#&&` application in n-ary products where n>0

Resolves #368


L-as (approved):
LGTM

Issue #382, opened 2022-03-16
`ghcid` is broken


srid:
After #378, the ghcid commands no longer work:

```sh-session
$ ghcid -c 'cabal repl plutarch-test:exe:plutarch-test' -T Main.main
...
Preprocessing executable 'plutarch-test' for plutarch-test-1.1.0..
GHCi, version 9.2.1: https://www.haskell.org/ghc/  :? for help

<no location info>: error: [-Wmissing-home-modules, -Werror=missing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules: 
        Plutarch.FFISpec
Failed, no modules loaded.

No files loaded, GHCi is not working properly.
Command: cabal repl plutarch-test:exe:plutarch-test
```

blamario:
That's what I said on Discord:
> The major problem now is that sydtest-discover is totalitarian. The current hacky workaround only helps the CI, doesn't work inside nix develop.

https://github.com/NorfairKing/sydtest/issues/40

srid:
It does look like we ought to get rid of `sydtest-discover`, but I cannot confirm if that will fix the problem in this case.

blamario:
Fixee by PR #387

PR #387, opened 2022-03-17
Mario/issue382

blamario:


L-as:
#382

L-as (approved):
LGTM, albeit honestly I'd prefer that we stop using sydtest-discover entirely. This "in between" (AFAICT at least) is very odd.

PR #390, opened 2022-03-18
plutarch-test: Look for unused goldens and fail.

### LINKED ISSUE #389: Check for unused goldens in CI

srid:
Resolves #389 

Also,

- Normalize hand-written goldens in `ScriptsSpec` to use the existing golden framework so that they are tracked for the unused checks. This uses `pconstant` to shoehorn the Haskell string into the current golden format. 
- `ScriptsSpec` now produces a single golden set, thus reducing the number of goldens.

User projects can perform this check by using `Plutarch.Test.Run.runPlutarchTest`


srid:
Demo https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2Fqbbx6dm5hmswv82i6mrydf2w2jc6sp7c-test-ghc810-nodev.drv/log?via-job=fde8f7ac-b7f1-467b-97ba-8429cc280bb8

image

plutarch-test/Main.hs
```diff
@@ -8,24 +8,36 @@ import qualified ExtraSpec
...
 main :: IO ()
 main = do
   setLocaleEncoding utf8
-  sydTest BaseSpec.spec
-  putStrLn "\n\n\n--------------------------------------------\n\n\n"
-  sydTest ExtraSpec.spec
 
-  putStrLn "\n\n\n--------------------------------------------\n\n\n"
+-- We use 'runPlutarchSpec' to test for unused goldens, but do so only in GHC
+-- 9. Because, under GHC 8 certain modules are disabled (see the CPP below)
```
L-as, line 25: This assumption doesn't hold after #378

L-as: in the sense that the same is true of GHC 9

srid: That PR hasn't changed anything in that regard. 

`Plutarch.FieldSpec` and `Plutarch.MonadicSpec` are still unused in GHC 8. In GHC 8 their associated goldens (see screenshot posted above) are seen incorrectly as unused goldens unless we do this CPP switch.

srid: > in the sense that the same is true of GHC 9

In practice, however, although we disable `Plutarch.FFISpec` in GHC 9 - that module is not generating any goldens, so it doesn't impact the unused goldens check. 

Would you instead go for a more general approach? We can use a known file path prefix for GHC 8 only goldens (Eg: `goldens/ghc8.field.data.bench.golden`) and automatically ignore them when checking under GHC 9.

L-as: Honestly, I'd say don't bother. This is fine. It's not worth the effort.

L-as: The proper solution would be to somehow accumulate state and run the process twice compiled with difference GHCs and then check that in CI.

L-as (approved):
LGTM

PR #391, opened 2022-03-21
Port guide overhaul to staging

TotallyNotChase:
There have not been any changes to the guide itself compared to what's on `master`, other than in the final commit: [Migrate PPrelude information](https://github.com/Plutonomicon/plutarch/commit/67a0b022b45b876e58405053eb4fea2930adb346)


L-as (approved):
Honestly haven't checked the contents but this seems infeasible given the size

TotallyNotChase:
A trick that might be useful for reviewing cases like these is by checking "Files changed" against both `staging` and `master`. The `master` diff shouldn't have any changes to `docs/` and the staging diff shouldn't have any changes to anything other than `docs/`

blamario:
That was brave. It looks like only docs were affected, but I don't entirely trust Git to figure out the staging -> master merge.

PR #357, opened 2022-03-04
Upstream @Geometer1729's list utilities

MangoIV:
### Changes 

- merged Geometers `list-upstream` branch with my `staging` branch

### Notes

- **This needs #351** 

### Open tasks

- [x] fix spelling error
- [x] factor out methods that ought to be in `plutarch-extra`
- [x] refactor methods as reviewed by @TotallyNotChase 
- [ ] ~~remove `hedgehog` dep~~ We're using something that isn't in `QuickCheck`
- [ ] task by @srid: maybe simplify `Util.hs`

Issue #331, opened 2022-02-23
Switch to tasty and plutus-golden


L-as:
sydtest is under a non-commercial license: https://github.com/NorfairKing/sydtest/blob/master/sydtest/LICENSE.md

This is quite a big problem.

@NorfairKing
@srid 

We can try converting to `tasty`, but we need to vendor the golden file logic probably. See https://github.com/Liqwid-Labs/plutus-extra/tree/master/plutus-golden

srid:
> https://github.com/NorfairKing/sydtest/blob/master/sydtest/LICENSE.md

I *think* the license Plutarch (and its users) has for sydtest is different to this one. I will let @NorfairKing clarify.

L-as:
![image](https://user-images.githubusercontent.com/22075344/155519096-f5d0b1f5-1c38-4820-948a-fdda1af3782b.png)

Seems like we'll have to switch to something else if others are to use `plutarch-test`.

L-as:
@kozross might be interested in this

kozross:
Yeah, I would say a switch-away is essential here. I'm happy to help with this effort.

L-as:
That would be really helpful. The biggest potential issue seems to be that tasty does not tell us where we are in the test hierarchy we are for figuring out the names of the golden files.

kozross:
`tasty` (and any stuff built atop it) does golden tests quite badly in my experience.

kozross:
`tasty` is significantly better than `hspec` in my experience, from the perspective of documentation, usability and extensibility.

NorfairKing:
> tasty is significantly better than hspec in my experience, from the perspective of documentation, usability and extensibility.

My experience contradicts this, fwiw. See the comparison table here:
https://github.com/NorfairKing/sydtest#features--comparison-to-similar-projects

srid:
@kozross, I'd be interested to hear which test runner you plan to adopt and how it will impact the goldens tree syntax we have now (it relies on sydtest's monadic test spec hierarchy to compute the filename).

Ideally, we want to keep changes to the `Plutarch.API` exports minimal, and just swap out the internals.

L-as:
https://github.com/NorfairKing/sydtest/pull/39

PR #395, opened 2022-03-22
Cleanup `Plutarch.Test.Property` (and export?)

srid:
This builds on top of the prior upstreaming in #357 to prepare the property testing library for use in other projects.

- Replace `Util.hs:EquivalenceMethod` with `Equality`/`Totality`, because the former can allow the library user to generate redundant tests.
- Replace `haskPlutEquiv` with `prop_haskEquiv`; the latter uses -
  - The new Equality/Totality types
  - TypeApplications instead of proxy helpers (thus avoids large exports)
  - `NP` from sop-core to build arguments.
- Move `Marshal` to separate module
- Remove unused generators

As well as some upstreaming:

- Use a large budget for evaluating property tests by necessity
- Upstream `prop_dataRoundTrip` in Extra module, which will contain properties based on `prop_haskEquiv`
- `Marshal`: add default implementation using `PLifted`


L-as:
> This builds on top of the prior upstreaming in https://github.com/Plutonomicon/plutarch/pull/357 to prepare the property testing library for use in other projects.

This doesn't make sense when you consider that it's `plutarch-test` which is unusable by downstream projects. You could rename `plutarch-test` to `plutarch-internal-test` and make a new `plutarch-test` that doesn't depend on `sydtest`.

L-as:
We could also remove the `sydtest` dependency from `plutarch-test`.

L-as:
The third option is not to merge the code at all and remove all the unused property testing code.

srid:
Ardana still can use it, though. 

I am assuming that other projects refrain from using `plutarch-test` because of sydtest licensing restrictions. That said, even though `Plutarch.Test.Property` can _currently_ only be used by Ardana, I expect it will _eventually_ become usable for other projects after #331 is done. 

I'd offer 2 more options:

- We can move this code to a separate package, `plutarch-test-property`, if other projects want to use it right away without waiting for #331, but this extra structure is not necessary and becomes redundant after tasty migration.
- Alternatively, we un-export `Plutarch.Test.Property`, remove the unused `Extra.hs` - and merge the PR. Ardana projects would patch the .cabal in Nix to export the module (until #331 is done that is).

L-as:
I don't see why we would do something specific to Ardana.

srid:
We don't have to -- hence that 2nd option.

TotallyNotChase:
well, in any case, are we working on that sydtest switch? What would be some of the challenges/blockers you can foresee?

L-as (approved):
Merging as this is still a good change

srid:
> well, in any case, are we working on that sydtest switch? What would be some of the challenges/blockers you can foresee?

I finally got to work on it. It probably will be hspec + hedgehog. Main question was about accessing test hierarchy in monadic specs (for determining golden file paths); and one approach to that (could be upstreamed) is: https://github.com/Plutonomicon/plutarch/commit/fd84c3d6bf6d3035aaec84aab5ea0ebea62d7a45

PR #396, opened 2022-03-24
Mario/value api

blamario:


Plutarch/Api/V1/AssocMap.hs
```diff
@@ -56,3 +63,37 @@ instance
...
+    lookupDataWith
+      # (phoistAcyclic $ plam $ \pair -> pcon $ PJust $ pfromData $ psndBuiltin # pair)
+      # pdata key
+
+-- | Look up the given key data in a 'PMap'.
+lookupData :: (PIsData k, PIsData v) => Term (s :: S) (PAsData k :--> PMap k v :--> PMaybe (PAsData v))
+lookupData = lookupDataWith # (phoistAcyclic $ plam $ \pair -> pcon $ PJust $ psndBuiltin # pair)
+
+-- | Look up the given key data in a 'PMap', applying the given function to the found key-value pair.
+lookupDataWith ::
```
TotallyNotChase, line 80: I think it might be more efficient to have an explicit impl here, something of this sort perhaps
```hs
plookup :: (PIsData k, POrd k) => Term s (PAsData k :--> PMap k v :--> PMaybe (PAsData v))
plookup = phoistAcyclic $
  plam $ \kdat m ->
    plet (pfromData kdat) $ \k ->
      precList
        ( \self x xs -> unTermCont $ do
            k' <- pletC $ pfstBuiltin # x
            pure $
              pif
                (pfromData k' #< k)
                (self # xs)
                $ pif
                  (kdat #== k')
                  (pcon . PJust $ psndBuiltin # x)
                  $ pcon PNothing
        )
        (const $ pcon PNothing)
        # pto m
```

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -56,3 +63,37 @@ instance
...
+    )
+lookupDataWith = phoistAcyclic $
+  plam $ \unwrap key map ->
+    pmaybe # pcon PNothing # unwrap
+      #$ plet (plam $ \pair -> pfstBuiltin # pair #== key)
+      $ \predicate ->
+        pfind # predicate # pto map
+
+-- | Construct a singleton 'PMap' with the given key and value.
+singleton :: (PIsData k, PIsData v) => Term (s :: S) (k :--> v :--> PMap k v)
```
TotallyNotChase, line 97: potentially better to take in `PAsData k` and `PAsData v` to avoid the extraneous `pdata` calls within.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
```diff
@@ -47,6 +48,10 @@ spec = do
             $ \case
               PMinting c -> popaque c
               _ -> perror
+    describe "value" $ do
+      pgoldenSpec $ do
+        "singleton" @| PValue.singleton # pconstant "c0" # pconstant "sometoken" # 1 @-> \p ->
+          plift p @?= mint
```
srid, line 54: Can we add goldens for all the new functions in this PR, not just `singleton`?

blamario: Of course, that's why this is a draft PR.

Plutarch/Api/V1/AssocMap.hs
blamario: Good point, but I've made that a separate function `singletonData`.

Plutarch/Api/V1/AssocMap.hs
blamario: That does improve performance:

      Actual:   
      valueOf {"exBudgetCPU":12794940,"exBudgetMemory":28740,"scriptSizeBytes":222}
      Expected: 
      valueOf {"exBudgetCPU":12106088,"exBudgetMemory":29008,"scriptSizeBytes":229}

However PlutusTx's [lookup](https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx-AssocMap.html#v:lookup) does not require `Ord k` and I think we should mirror the semantics of PlutusTx API.

I've used your `precList` idea to get

      valueOf {"exBudgetCPU":10676984,"exBudgetMemory":24208,"scriptSizeBytes":200}

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
srid: Ah, yes, I hadn't seen the draft status.

Plutarch/Api/V1/AssocMap.hs
TotallyNotChase: I think we're gonna have to decide whether we go for the invariant assumptions labeled by #344 anyway. In which case, we probably are going to have `POrd` in everything. I do imagine the optimization of finding a key through comparison along the way will be pretty bad for smaller maps and values though.

Also in general, I don't think mirroring PlutusTx's API is a good idea FWIW. There's a lot of inefficiency warts that don't work well for UPLC. But maybe we can afford to not have peak efficiency everywhere, that's upto others to decide.

L-as:
Haven't looked at the code too much but #344, we need to figure out whether these assumptions are correct or not.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
```diff
@@ -46,6 +48,68 @@ spec = do
...
+            plift p @?= mint <> mintOtherSymbol
+    describe "map" $ do
+      pgoldenSpec $ do
+        let pmap, pdmap, emptyMap, doubleMap, otherMap :: Term _ (AssocMap.PMap PByteString PInteger)
+            pmap = AssocMap.singleton # pconstant "key" # 42
+            pdmap = AssocMap.singletonData # pdata (pconstant "key") # pdata 42
+            emptyMap = AssocMap.empty
+            doubleMap = AssocMap.singleton # pconstant "key" # 84
+            otherMap = AssocMap.singleton # pconstant "newkey" # 6
+        "lookup" @| AssocMap.lookup # pconstant "key" # pmap #== pcon (PJust 42) @-> passert
```
srid, line 77: Ok, so the general advice when writing golden programs is to keep assertion checks out of the golden itself. So this should instead look like:

```haskell
"lookup" 
  @| AssocMap.lookup # pconstant "key" # pmap
  @-> \result -> passert $ result #== pcon (PJust 42)
```

The idea is to store the `lookup` function application _only_ in the golden file. Only the expression being tested is made a golden value. The comparing of its evaluated result happens outside, specifically in the `@->` assertion. 

(Likewise for other golden below).

plutarch-test/goldens/api.map.uplc.eval.golden
```diff
@@ -0,0 +1,19 @@
+lookup (program 1.0.0 True)
+lookupData (program 1.0.0 True)
```
srid, line 2: Hence [this advice](https://github.com/Plutonomicon/plutarch/pull/396/files#r845580960), because otherwise `eval.golden` will be filled with `True`, which is not very useful.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
```diff
@@ -202,3 +272,6 @@ getFields = phoistAcyclic $ plam $ \addr -> psndBuiltin #$ pasConstr # addr
 
 dummyCurrency :: CurrencySymbol
 dummyCurrency = Value.currencySymbol "\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11"
+
+pshouldReallyBe :: ClosedTerm a -> ClosedTerm a -> Expectation
+pshouldReallyBe a b = pshouldBe b a
```
srid, line 277: Hmm, what is this for?

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
srid: Basically typically wherever `@-> passertNot` or `@-> passert` is used.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
blamario: Use of `pshouldBe` was giving me ambiguous type errors, so I decided to tighten the constraints. I also swapped the arguments to make it more convenient.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
blamario: Done.

TotallyNotChase:
So uhh, are we going with the normalized assumptions for Plutarch 1.2?

At this point, I'm honestly ok with going without that assumption and sacrificing a bit of performance for consistency and ease of Plutarch 1.2 rollout.

Anyway, I'd like to benchmark these a bit - I'll do that as soon as I can. They look quite neat though.

blamario:
From the IOHK answers it seems that we shouldn't count on any normalization, but I'll expand on this thought in issue #344. I have a few notes regarding the benchmarks.

First, I'm quite confident this implementation is *far* better than what `Plutus.V1.Ledger.Value` provides. Leaving aside the general horror of any code generated by `plutus-tx-plugin`, the algorithms in the Plutus module tend to generate intermediate data structures.

Second, the question of optimality is unfortunately a matter of tradeoffs to some extent. For example, commit https://github.com/Plutonomicon/plutarch/pull/396/commits/e7639104b113384b1a7d334c7652f1f9c6971846 slightly increased the cost of both `insert` and `delete`, but because it made them share most of their code any script that uses both will now see the total code size decrease.

Third, I've intentionally left a couple of minor optimizations on the floor. For example I added `AssocMap.mapEitherWithKey :: Term (s :: S) ((k :--> a :--> PEither b c) :--> PMap k a :--> PPair (PMap k b) (PMap k c))` because there's a function of the same name and function in `Data.Map`, its signature makes it clear what it does, and because I needed it to implement `unionWith`. The optimal helper function would actually be `AssocMap.mapMaybeWithKey :: Term (s :: S) ((k :--> a :--> PMaybe b) :--> PMap k a :--> PPair (PMap k a) (PMap k b))` but it's less general, there's no equivalent in `Data.Map`, and the optimization would be really minor.

L-as:
Either we assume sortedness or enforce it.

L-as:
PUnsortedValue, PValue, ptryFrom

L-as:
I think the direction this is going towards is that we should enforce normalisation and make a CIP for enforcing it.

L-as:
https://github.com/input-output-hk/plutus/issues/4509#issuecomment-1093472123

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -18,9 +47,45 @@ import Plutarch.Lift (
...
+  (#),
+  (#$),
+  type (:-->),
+ )
+import Plutarch.Unsafe (punsafeFrom)
+
+import Prelude hiding (all, lookup)
 
-newtype PMap (k :: PType) (v :: PType) (s :: S) = PMap (Term s (PBuiltinMap k v))
+newtype PMap (k :: PType) (v :: PType) (s :: S) = PMap (Term (s :: S) (PBuiltinMap k v))
```
L-as, line 88: ?

Plutarch/Api/V1/AssocMap.hs
L-as: There is no reason to mirror what PlutusTx does.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -3,12 +3,41 @@
 
 module Plutarch.Api.V1.AssocMap (
   PMap (PMap),
+
+  -- * Creation
+  empty,
```
L-as, line 8: Naming

Plutarch/Maybe.hs
```diff
@@ -34,3 +36,10 @@ pfromJust = phoistAcyclic $
   plam $ \maybe -> pmatch maybe $ \case
     PNothing -> perror
     PJust a -> a
+
+-- | 'PMaybe' destructor equivalent to 'maybe'
+pmaybe :: Term (s :: S) (b :--> (a :--> b) :--> PMaybe a :--> b)
```
L-as, line 41: This is likely not efficient.

Plutarch/Either.hs
```diff
@@ -11,3 +21,10 @@ data PEither (a :: PType) (b :: PType) (s :: S)
   | PRight (Term s b)
   deriving stock (GHC.Generic)
   deriving anyclass (Generic, HasDatatypeInfo, PlutusType, PEq, PShow)
+
+-- | 'PEither' destructor equivalent to 'either'
+peither :: Term (s :: S) ((a :--> c) :--> (b :--> c) :--> PEither a b :--> c)
```
L-as, line 26: This is likely not efficient.

flake.nix
```diff
@@ -615,7 +615,7 @@
           type = "app";
           program = checkedShellScript system "plutarch-test-${name}"
             ''
-              cd ${self}/plutarch-test
+              { touch ./plutarch-test/.crumb 2> /dev/null && cd ./plutarch-test && rm .crumb ; } || cd ${self}/plutarch-test
```
L-as, line 618: What is this?

Plutarch/Api/V1/AssocMap.hs
blamario: You mean you want `pempty` instead? Any other functions to put p on?

flake.nix
blamario: It's a workaround for a permission problem with `nix run`:
https://discord.com/channels/922576618424250418/922577570766143498/962077469765165076

Plutarch/Either.hs
blamario: Relative to what? It's slightly less efficient than `pmatch`:
```
peither {"exBudgetCPU":119192,"exBudgetMemory":500,"scriptSizeBytes":23}
pmatch {"exBudgetCPU":119192,"exBudgetMemory":500,"scriptSizeBytes":18}
```

where

```
    "peither" @| peither # psucc # psucc @-> \p -> passert (p # pcon (PLeft 42) #== 43)
    "pmatch"
      @| plam (\p-> pmatch p $ \case {PLeft n -> psucc # n; PRight n -> psucc # n})
      @-> \p -> passert (p # pcon (PLeft 42) #== 43)
```

Plutarch/Maybe.hs
blamario: I'm not using it any more, so it can be removed if you find it dangerous.

Plutarch/Either.hs
blamario: It's still used by `AssocMap.mapEitherWithKeyData` but this is not used any more, so it can all be removed I suppose.

Plutarch/Api/V1/AssocMap.hs
blamario: I don't know what I was thinking. Reverted.

blamario:
Note that this PR depends on PR #453.

blamario:
The current state of the PR is that `PMap` and `PValue` are kept sorted by all operations. All insertions, merges, and equality comparisons make use of this, but lookups don't. There are functions `AssocMap.assertSorted` and `Value.assertSorted` to make sure the inputs are already sorted and normalized, and also `Value.normalize` that takes a sorted value potentially containing zeroes or empty submaps and removes them. The latter is to provide a way to clean up after relatively rare operations like `Value.unionWith # (plam (-))` that may produce a non-normalized `PValue`.

TotallyNotChase:
Do you think it's a good idea to bake in the normalization check inside `unionWith` so that its output is _never not normalized_? Effectively by checking the union result before cons-ing it on.

Also, should `pnormalize` also sort?

Plutarch/Api/V1/Value.hs
```diff
@@ -48,3 +57,102 @@ deriving via
...
+  a #== b = pto a #== pto b
+
+instance Semigroup (Term s PValue) where
+  a <> b = unionWith # plam (+) # a # b
+
+instance Monoid (Term s PValue) where
+  mempty = pcon (PValue AssocMap.empty)
+
+-- | Construct a singleton 'PValue' containing only the given quantity of the given currency.
+singleton :: Term (s :: S) (PCurrencySymbol :--> PTokenName :--> PInteger :--> PValue)
```
TotallyNotChase, line 72: I'd prefer a version that takes in `PAsData` values instead, either as a separate function or a modified version of this one.

The same goes for any other function that could skip a `pdata`/`pfromData` within if provided the more focused type.

Plutarch/Api/V1/Value.hs
```diff
@@ -48,3 +57,102 @@ deriving via
...
+valueOf = phoistAcyclic $
+  plam $ \value symbol token ->
+    AssocMap.foldAt
+      # symbol
+      # 0
+      # plam (\map -> AssocMap.foldAt # token # 0 # plam pfromData # pfromData map)
+      # pto value
+
+-- | Check if the value is zero.
+isZero :: Term (s :: S) (PValue :--> PBool)
```
TotallyNotChase, line 88: Aren't we assuming this will never happen? (assuming `unionWith` is changed to ensure it never returns values with 0s in them)

L-as:
`Term n PValue` should *always* be normalised. You can add `PBadValue` if you want to have unnormalised values.

Plutarch/Api/V1/Value.hs
blamario: There already is `singletonData`.

blamario: Not for `Value` though, but how often do you have `Term s (PAsData PInteger)` lying around? It seems unlikely.

blamario:
@TotallyNotChase 
> Do you think it's a good idea to bake in the normalization check inside `unionWith` so that its output is never not normalized?

No, because it would be a waste most of the time. There's no need to check after `unionWith (plam const)` or `unionWith (plam +))` for example.

> Also, should `pnormalize` also sort?

None of the operations, including `unionWith`, is ever going to un-sort a value. The present design is to apply `assertSorted`  once to the input values, then to manipulate them using the provided operations using `normalize` only when necessary.

@L-as
> `Term s PValue` should *always* be normalised. You can add `PBadValue` if you want to have unnormalised values.

I don't want to replicate every operation to operate on `PBadValue`s. How about adding a phantom type parameter instead:

```
newtype PValue (state :: ValueState) (s :: S) = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName PInteger)))

data ValueState = Sorted Normalization | Unverified
data Normalization = Normalized | Unnormalized

assertSorted :: Term s ((PValue Unverified) :--> PValue (Sorted Normalized))
normalize :: Term s (PValue (Sorted _) :--> PValue (Sorted Normalized))

unionWith :: Term s (PValue (Sorted _) :--> PValue (Sorted _) :--> PValue (Sorted Unnormalized))

instance {-# OVERLAPS #-} PEq (PValue (Sorted Normalized))
instance PEq (PValue _)

(#==) :: Term s (PValue (Sorted Normalized)) -> Term s (PValue (Sorted Normalized)) -> Term s PBool
isZero :: Term s (PValue (Sorted Normalized)) -> Term s PBool
singleton :: Term (s :: S) (PCurrencySymbol :--> PTokenName :--> PInteger :--> PValue (Sorted Normalized))
valueOf :: Term s (PValue _ :--> PCurrencySymbol :--> PTokenName :--> PInteger)
```

Plutarch/Api/V1/Value.hs
L-as: Very often if it's in your datum or redeemer?

blamario: I've added the function.

L-as:
What do you mean by `Sorted Unnormalised`?

blamario:
> I don't think there's any point in parameterising `PValue` since we don't have any functions that can operate on any `PValue x`.

`assertSorted` operates on any `PValue x`,  `Semigroup` and `Monoid` operate on any `Sorted` value,  `#==` has two different instances for `Sorted Normalized` and `Sorted NotNormalized`.

> What do you mean by `Sorted Unnormalised`?
It's for values with sorted keys that may contain zero amounts or empty submaps. Such as returned by `unionWith # (plam (-))`.

blamario:
Yesterday's API was simpler, but less safe. That's when it was possible to forget to normalize after a `unionWith`, today's API prevents that.

To p or not to p, that is the question. I don't mind if that's the consensus.

blamario:
The plot thickens. I realized I didn't account for values with negative amounts, which are legal in Plutus AFAIU. The `Semigroup` instance thus can't assume that zeroes won't appear in the result. So I introduced the `Positive` subtype, and I separated `ValueKeyGuarantees = Unsorted | Sorted` from `ValueAmountGuarantees = NoGuarantees | NonZero | Positive` as two orthogonal `PValue` parameters which seems clearer.

I've also added `pconstantSingleton` and `pconstantPositiveSingleton` because they can cheaply inspect the amount on the Haskell side and generate an optimal term of the right type. I'm not sure if they should be taking `ClosedTerm PInteger` like now or just `Integer`, I'd appreciate some feedback there.

TotallyNotChase:
I think we should just introduce a check for "zero"-ness for each `unionWith` resultant element at this point, that'd trivialize the API at the cost of some efficiency.

I wonder how that'd compare to just not assuming sortedness/normalization in `unionWith`

Plutarch/Rec/TH.hs
```diff
@@ -15,10 +16,13 @@ deriveAll name = (<>) <$> deriveScottEncoded name <*> Rank2.TH.deriveAll name
 -- | Use as a TH splice for @type instance ScottEncoded@ declarations.
 deriveScottEncoded :: TH.Name -> Q [TH.Dec]
 deriveScottEncoded name = do
-  con <- reifyConstructor name
+  (con, tyVars) <- reifyConstructor name
```
L-as, line 19: ?

flake.nix
L-as: I'm honestly not sure what this workaround is supposed to do? Elaborate?

Plutarch/Rec/TH.hs
```diff
@@ -37,10 +41,20 @@ bare :: TH.Type -> Q TH.Type
...
+bindingName (TH.KindedTV name _) = name
+#endif
+
+reifyConstructor :: TH.Name -> Q (TH.Con, [TH.Name])
 reifyConstructor ty = do
   (TH.TyConI tyCon) <- TH.reify ty
   case tyCon of
-    TH.DataD _ _nm _tyVars _kind [c] _ -> return c
-    TH.NewtypeD _ _nm _tyVars _kind c _ -> return c
+    TH.DataD _ _nm tyVars _kind [c] _ -> return (c, bindingName <$> tyVars)
```
L-as, line 58: Separate PR?

Plutarch/Maybe.hs
L-as: I think we should remove this unfortunately, yeah. Let's do this in Plutarch 2.0 efficiently.

Plutarch/Either.hs
L-as: Plutarch 2.0 todo as with `pmaybe`

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -1,14 +1,49 @@
+{-# LANGUAGE TemplateHaskell #-}
```
L-as, line 1: Where are we using template haskell?

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -18,10 +53,50 @@ import Plutarch.Lift (
...
+import Plutarch.Prelude (
+  DerivePNewtype (..),
+  PAsData,
+  PBool (PFalse),
+  PBuiltinPair,
+  PCon (pcon),
+  PConstantData,
+  PEither (..),
+  PEq ((#==)),
+  PIsData (..),
```
L-as, line 66: Please no `..` imports, it's not clear what will be imported. In addition, import methods as if they were normal definitions.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -47,3 +122,339 @@ instance
...
+-- | Look up the given key in a 'PMap'.
+plookup :: (PIsData k, PIsData v) => Term (s :: S) (k :--> PMap k v :--> PMaybe v)
+plookup = phoistAcyclic $
+  plam $ \key ->
+    plookupDataWith
+      # (phoistAcyclic $ plam $ \pair -> pcon $ PJust $ pfromData $ psndBuiltin # pair)
+      # pdata key
+
+-- | Look up the given key data in a 'PMap'.
+plookupData :: (PIsData k, PIsData v) => Term (s :: S) (PAsData k :--> PMap k v :--> PMaybe (PAsData v))
```
L-as, line 139: Please don't add these unnecessary kind signatures at the use sites in definition types.
`forall (s :: S).` is OK, but `A (s :: S)` isn't at the use-site. It's essentially the same as doing `f (x :: X)` unnecessarily at the term-level.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -47,3 +122,339 @@ instance
...
+psingletonData :: (PIsData k, PIsData v) => Term (s :: S) (PAsData k :--> PAsData v :--> PMap k v)
+psingletonData = phoistAcyclic $
+  plam $ \key value -> punsafeFrom (pcons # (ppairDataBuiltin # key # value) # pnil)
+
+-- | Construct a 'PMap' from a list of key-value pairs, sorted by ascending key data.
+pfromAscList :: (POrd k, PIsData k, PIsData v) => Term (s :: S) (PBuiltinMap k v :--> PMap k v)
+pfromAscList = plam $ (passertSorted #) . pcon . PMap
+
+-- | Assert the map is properly sorted
+passertSorted :: (POrd k, PIsData k, PIsData v) => Term (s :: S) (PMap k v :--> PMap k v)
```
L-as, line 261: ? Shouldn't it always be sorted?

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -47,3 +122,339 @@ instance
...
+  , mergeInsert :: f (PBuiltinPair (PAsData k) (PAsData v) :--> PBuiltinMap k v :--> PBuiltinMap k v :--> PBuiltinMap k v)
+  }
+
+type instance
+  ScottEncoded (MapUnion k v) a =
+    (PBuiltinMap k v :--> PBuiltinMap k v :--> PBuiltinMap k v)
+      :--> (PBuiltinPair (PAsData k) (PAsData v) :--> PBuiltinMap k v :--> PBuiltinMap k v :--> PBuiltinMap k v)
+      :--> a
+
+$(Rank2.TH.deriveAll ''MapUnion)
```
L-as, line 288: Please no TH in Plutarch. TH is in general a PITA.

L-as: You can do this in other ways, by e.g. having `(x : Bool) -> if x then A else B` as the "type" (you'd have to use `POpaque` since we don't have dependent types).

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -47,3 +122,339 @@ instance
...
+instance (POrd k, PIsData k, PIsData v, Semigroup (Term s v)) => Monoid (Term s (PMap k v)) where
+  mempty = pempty
+
+{- | Combine two 'PMap's applying the given function to any two values that
+ share the same key.
+-}
+punionWith ::
+  (POrd k, PIsData k, PIsData v) =>
+  Term (s :: S) ((v :--> v :--> v) :--> PMap k v :--> PMap k v :--> PMap k v)
+punionWith = phoistAcyclic $
```
L-as, line 302: This doesn't seem efficient to me honestly, but we can look at that later.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -47,3 +122,339 @@ instance
...
+  letrec $ \MapUnion {merge, mergeInsert} ->
+    MapUnion
+      { merge = plam $ \xs ys -> pmatch xs $ \case
+          PNil -> ys
+          PCons x xs' -> mergeInsert # x # xs' # ys
+      , mergeInsert = plam $ \x xs ys -> pmatch ys $ \case
+          PNil -> pcons # x # xs
+          PCons y ys' ->
+            plet (pfstBuiltin # x) $ \xk ->
+              plet (pfstBuiltin # y) $ \yk ->
```
L-as, line 336: I think you want to `plet` `x` and `y` too. Can't remember whether they already are, but an extra `plet` is free.

Plutarch/Api/V1/Contexts.hs
```diff
@@ -43,8 +43,8 @@ newtype PTxInfo (s :: S)
           ( PDataRecord
               '[ "inputs" ':= PBuiltinList (PAsData PTxInInfo) -- Transaction inputs
                , "outputs" ':= PBuiltinList (PAsData PTxOut) -- Transaction outputs
-               , "fee" ':= PValue -- The fee paid by this transaction.
-               , "mint" ':= PValue -- The value minted by the transaction.
+               , "fee" ':= PValue 'Sorted 'NonZero -- The fee paid by this transaction.
```
L-as, line 46: We probably want some human-readable aliases.

L-as: `type PValue = PValue' 'Sorted 'NonZero`.

L-as: Also, this can not be negative

Plutarch/Api/V1/Value.hs
```diff
@@ -38,13 +54,200 @@ deriving via
...
+    pif
+      ( AssocMap.pall
+          # (plam $ \submap -> AssocMap.pall # plam (0 #<) # submap)
+          # pto value
+      )
+      (punsafeFrom $ pto value)
+      (ptraceError "Negative amount in Value")
+
+-- | Forget the knowledge of value's positivity.
+pforgetPositive :: Term s (PValue 'Sorted 'Positive) -> Term s (PValue 'Sorted 'NonZero)
```
L-as, line 249: seems like this could be more general? the return could be `PValue a b` I think.

Plutarch/Api/V1/Value.hs
```diff
@@ -38,13 +54,200 @@ deriving via
   instance
     PConstantDecl Plutus.CurrencySymbol
 
-newtype PValue (s :: S) = PValue (Term s (PMap PCurrencySymbol (PMap PTokenName PInteger)))
+data ValueKeyGuarantees = Sorted | Unsorted
+data ValueAmountGuarantees = NoGuarantees | NonZero | Positive
+
+newtype PValue (keys :: ValueKeyGuarantees) (amounts :: ValueAmountGuarantees) (s :: S)
```
L-as, line 60: So we have some subtyping relations here, I wonder whether we should add `PSubtype` instances. `PSubtype` essentially gives us one-way coercions.

L-as (commented):
Thanks!

Plutarch/Api/V1/Tx.hs
```diff
@@ -90,7 +90,7 @@ newtype PTxOut (s :: S)
           s
           ( PDataRecord
               '[ "address" ':= PAddress
-               , "value" ':= PValue
+               , "value" ':= PValue 'Sorted 'NonZero
```
L-as, line 93: This should be `Positive`, but you may want to add a warning wrt. the babel stuff, because then you can have negative value in the output.

L-as:
Maybe you want to rebase and clean up the history BTW?

L-as:
See also #465

Plutarch/Rec/TH.hs
blamario: There was a bug in TH generation for types with parameters additional to `(s :: S)`, such as the `PValue (keys :: ValueKeyGuarantees) (amounts :: ValueAmountGuarantees) (s :: S)`.

Plutarch/Api/V1/Value.hs
blamario: I've removed the function to simplify the API, it's no better than `(== mempty)` anyway.

flake.nix
blamario: The workaround code tests if directory `./plutarch-test/` is writable, and if so continues from there; if it doesn't it goes to `${self}/plutarch-test/` instead. I forget how exactly I ran into the problem, but `nix run .#test-ghc810-nodev` was failing because it didn't have permissions on the `${self}/plutarch-test` directory.

Plutarch/Rec/TH.hs
blamario: Okay, I'll cherry-pick this into a separate PR.

Plutarch/Api/V1/AssocMap.hs
blamario: In order to generate the instances for `MapUnion`, so I could use `letrec`.

Plutarch/Rec/TH.hs
L-as: This should be a separate PR really

Plutarch/Rec/TH.hs
blamario: I forgot I already put this in the merged PR #453, I just need to rebase this PR on top of staging.

Plutarch/Rec/TH.hs
blamario: It was, PR #453.

Plutarch/Api/V1/AssocMap.hs
blamario: Do you mean `plet (psndBuiltin # x) $ \xv ->`?

Plutarch/Api/V1/AssocMap.hs
blamario: This is meant for user-supplied values in datums and such.

Plutarch/Api/V1/AssocMap.hs
blamario: I've inlined and simplified the instances.

Plutarch/Api/V1/Value.hs
blamario: Do you mean you want to change the current definition of `PSubtypeDecl`? It's not overridable as it stands.

blamario:
I've cleaned up all the small stuff. What's left:

- [ ] synonym `type PValue = PValue' 'Sorted 'Positive`;
- [ ] a new module `Plutarch.Api.V1.Value.Simple` (?) that's less performant but makes all values `'Sorted 'NonZero` all the time and gets rid of type parameters;
- [ ] `PSubtype` instances would seem to require changes to `TryFrom`; and
- [ ] more instances (like `Group`) and utility functions.

The first two kinda solve the same problem of type-level noise in two different ways, so I don't know if I'd do both.

Plutarch/Api/V1/AssocMap.hs
L-as: Nope, `x` itself

Plutarch/Api/V1/AssocMap.hs
L-as: Well then it should be represented in the type, no? When would you ever have an unsorted map? Isn't this what `PTryFrom` is for?

Plutarch/Api/V1/Value.hs
L-as: Yep.

Plutarch/Api/V1/AssocMap.hs
blamario: I didn't want to add a phantom type parameter to `AssocMap` as well, but that is an option. I need this function as a helper for `PValue.passertSorted`.

Plutarch/Api/V1/AssocMap.hs
blamario: I tried; `plet x $ \x->` made no difference but `plet y $ \y->` did to my surprise. I've pushed the commit, but note that while the script size and CPU cost went down the memory cost went up slightly. It still seems to be an improvement.

Plutarch/Api/V1/AssocMap.hs
L-as: Well, shouldn't all maps be sorted? You don't need to support unsorted maps.

blamario: I've added a `KeyGuarantees = Sorted | Unsorted` type parameter to `PMap`, shared with `PValue`. This makes for a cleaner and safer interface overall. The downside is that `PLifted PMap` is now restricted to `Unsorted` because of the functional dependency.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -91,17 +93,20 @@ import qualified Rank2
 
 import Prelude hiding (all, any, filter, lookup, null)
 
-newtype PMap (k :: PType) (v :: PType) (s :: S) = PMap (Term s (PBuiltinMap k v))
-  deriving (PlutusType, PIsData, PEq, PShow) via (DerivePNewtype (PMap k v) (PBuiltinMap k v))
+data KeyGuarantees = Sorted | Unsorted
+
+newtype PMap (keysort :: KeyGuarantees) (k :: PType) (v :: PType) (s :: S) = PMap (Term s (PBuiltinMap k v))
+  deriving (PlutusType, PIsData, PEq, PShow) via (DerivePNewtype (PMap keysort k v) (PBuiltinMap k v))
+type role PMap nominal nominal representational phantom
```
L-as, line 100: They should all be `nominal`. Anything else is unsound.

L-as:
Should we just assume `ScriptContext` and related are correct?

Plutarch/Api/V1/AssocMap.hs
L-as: Do note that we have to do this for all other data types too, and I am in the process of doing that.

blamario: Okay. I just followed the existing `type role Term phantom representational`, you're saying that's going to change to `type role Term nominal nominal`?

L-as:
> > Should we just assume ScriptContext and related are correct?
> 
> I thought we already assumed (and verified, sort of) that? Otherwise, there's not much point in caring about sortedness/normalization.

I mean, when the user crafts them manually.

Plutarch/Api/V1/AssocMap.hs
L-as: Yep. It's broken.

L-as (approved):
LGTM

PR #398, opened 2022-03-24
Nix'ify docs website, and check for broken links in CI

srid:
Adds `nix run .#docs` to run a live web view of docs locally. As well as `nix run .#website` to build the static website. This is the same as what we do in https://github.com/Plutonomicon/plutonomicon - but using a newer version of Emanote containing support for the [full range of relative links](https://github.com/srid/emanote/pull/268).

The latter command also checks for broken links (the CI for this PR succeeds because those broken links have been fixed here).

Eventually, we can include Haddock generated HTML in the generated site of `nix run .#website`, before deploying it statically to GH pages using #393.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,6 +57,9 @@
   inputs.autodocodec.url = "github:srid/autodocodec/ghc921";
   inputs.autodocodec.flake = false;
 
+  inputs.emanote.url = "github:srid/emanote/rel-link-and-anchors";
+  # inputs.emanote.inputs.ema.inputs.nixpkgs.follows = "nixpkgs";
+
   outputs = inputs@{ self, nixpkgs, iohk-nix, haskell-nix, plutus, flake-compat, flake-compat-ci, hercules-ci-effects, ... }:
     let
       extraSources = [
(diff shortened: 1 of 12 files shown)


srid:
As expected CI fails due to broken links:

image

https://hercules-ci.com/accounts/github/Plutonomicon/derivations/%2Fnix%2Fstore%2F11rfc1njln7s6sbn584fnbch2vhnhchr-plutarch-docs-html.drv/log?via-job=f8e10c21-6471-4932-9a66-a0c076e8715a

TotallyNotChase (approved):
The docs look good, I think you fixed all the remaining broken links too?

Also, I was curious, can we have a search feature somehow?

TotallyNotChase:
actually could you also fix the [hyperlink here please](https://github.com/Plutonomicon/plutarch/blob/website/docs/Introduction/Plutarch%20Types.md?plain=1#L5) - I accidentally made it `\[[3](./../Concepts.md#references)]`, should actually point to `Introduction.md#references`

srid:
> The docs look good, I think you fixed all the remaining broken links too?

Yup.

> Also, I was curious, can we have a search feature somehow?

Not right now, but eventually[^now] I'd like to integrate something like https://stork-search.net/ into Emanote. 

[^now]: However, for now, it may be possible to create the Stork index and inject the JS all in Nix as part of `nix build .#website` to get it working on static site alone (if not the live server).

TotallyNotChase:
One small thing, what's up with the github issue hyperlinks getting skewed?

image

Check the troubleshooting section for example. It seems to generate two links from it.

TotallyNotChase (approved):
beautiful

flake.nix
```diff
@@ -582,6 +584,35 @@
...
+            mkdir $out
+            ${inputs.emanote.defaultPackage.${system}}/bin/emanote \
+              --layers "${self}/docs;${configDir}" \
+              gen $out
+          '';
+      plutarchWebsiteApp = system:
+        rec {
+          type = "app";
+          # '' is required for escaping ${} in nix
+          script = (nixpkgsFor system).writers.writeBash "emanoteLiveReload.sh" ''
```
L-as, line 609: How is this different from `writeShellScriptBin`?

srid: It's a copy-paste from [plutonomicon](https://github.com/Plutonomicon/plutonomicon/blob/f31ce5405777558ae9c7eb83c739266ca579ee98/flake.nix#L26), but now that you mention it I realize we should instead be using [`checkedShellScript`](https://github.com/Plutonomicon/plutarch/blob/aecc2050eb63ff0041576473aa3193070fe91314/flake.nix#L533) from the flake.

Issue #385, opened 2022-03-16
Fix kind of HRec


L-as:
it should be `type HRec :: [(Symbol, Type)] -> Type`

L-as:
I'm thinking maybe we don't want to change the kind but instead fix the constructor.

L-as:
your idea is good

TotallyNotChase:
I'm a bit confused here. I think `Labeled Symbol` wouldn't work as the environment here, since in the current use cases, `Labeled` is being treated as if it had the kind signature-
```hs
type Labeled :: Symbol -> Type -> Type
```
(though in reality, the current kind signature of `Labeled` is `forall k. k -> Type -> Type`)

That is, the first parameter of `Labeled` must be a type level symbol. `Labeled Symbol` wouldn't fit this description right? Since that'd mean `Labeled` has kind `Type -> Type -> Type`, and you can't actually specify `Labeled name` where `name :: Symbol`.

Anyway, the environment argument shouldn't actually take the `Symbol` literal I believe, since it's supposed to be different for each type in `as`.

I do think a lot of this `HRec` stuff and the machinery in `TryFrom` can be simplified via `sop-core` but it might be a bit involved.

Issue #392, opened 2022-03-21
Fix `pto` for `PDataRecord xs`


L-as:
Currently it doesn't work because there is no catch-all `PlutusType (PDataRecord xs)` instance

TotallyNotChase:
is it even possible to safely define a `pmatch'` for `PDataRecord xs`?

```hs
pmatch' :: Term s (PBuiltinList PData) -> (PDataRecord xs s -> Term s b) -> Term s b
pmatch' l f = pelimList
  (\x xs ->
    let
        dl :: ???
        dl = PDCons (punsafeCoerce x) (punsafeCoerce xs)
      in f dl -- type mismatch
  )
  (f PDNil)  -- type mismatch
  l
```

PR #415, opened 2022-04-04
Scattered cleanups

### LINKED ISSUE #369: Can't use `OverloadedRecordDot` when extracting the `"data"` field from `PTxInfo`
### Problem
`data` is a reserved word in Haskell and hence you can't do something like `txInfo.data` without the Parser complaining. 

### Solution
Perhaps rename to `data'` or `dataMap`?

### LINKED ISSUE #400: Fix remaining issues in #326

### LINKED ISSUE #402: Unexport `hrecField`
We already have `getField` which does the same thing.

TotallyNotChase:
* Unexport `hrecField`. Resolves #402 
* Revert removal of `PValue` and `PMap` constructors. Resolves #400 
* Rename `PTxInfo`'s `data` field to `datums`. Resolves #369 
* Add `Num` instance for `PPOSIXTime` and export its constructor.
* Change `PIsDataRepr` superclass to be `PlutusType`, not `PMatch`.

Also trying to fix a few more small things that will be pushed up later.


L-as:
Merge in staging to fix CI

TotallyNotChase:
* Change `HRec`'s kind to `[(Symbol, Type)] -> Type` - Resolves #385 

I've been thinking of refining this one further. It's possible to do-
```hs
type Labeled :: (Symbol, Type) -> Type
newtype Labeled labeledX = Labeled {unLabeled :: Snd labeledX}

type HRec :: [(Symbol, Type)] -> Type
type HRec as = NP Labeled as
```
which is beautiful! But I don't know if a type alias is the right thing to do here since we have instances for `HRec`.

More importantly, I _believe_ the most elegant type for `HRec` would involve `[(Symbol, PType)]` as the type level list parameter - but I think that'll make quite a few classes and families a fair bit more complicated. We could even have the type level list be `[(Symbol, k)]` and make `HRec` (and everything else) take a `f :: k -> Type` (generics-sop style). I don't know if any of these refinements are worth the trouble though.

L-as (approved):
Looks good to me for now.

L-as:
> * Change `HRec`'s kind to `[(Symbol, Type)] -> Type` - Resolves [Fix kind of HRec #385](https://github.com/Plutonomicon/plutarch/issues/385)
> 
> 
> I've been thinking of refining this one further. It's possible to do-
> 
> ```haskell
> type Labeled :: (Symbol, Type) -> Type
> newtype Labeled labeledX = Labeled {unLabeled :: Snd labeledX}
> 
> type HRec :: [(Symbol, Type)] -> Type
> type HRec as = NP Labeled as
> ```
> 
> which is beautiful! But I don't know if a type alias is the right thing to do here since we have instances for `HRec`.
> 
> More importantly, I _believe_ the most elegant type for `HRec` would involve `[(Symbol, PType)]` as the type level list parameter - but I think that'll make quite a few classes and families a fair bit more complicated. We could even have the type level list be `[(Symbol, k)]` and make `HRec` (and everything else) take a `f :: k -> Type` (generics-sop style). I don't know if any of these refinements are worth the trouble though.

You could just make it a newtype, but that would ruin the point a bit and would complicate constructing and destructing.
I think it's fine the way you've done it.

TotallyNotChase:
The last thing I tried to work on is #392. There's a WIP of it in the latest commit but I can't see how the `pmatch` could be implemented in a useful way - does anyone have any ideas?

Plutarch/DataRepr/Internal.hs
```diff
@@ -108,6 +107,12 @@ data PDataRecord (as :: [PLabeledType]) (s :: S) where
     PDataRecord ((name ':= x) ': xs) s
   PDNil :: PDataRecord '[] s
 
+instance {-# OVERLAPPABLE #-} PlutusType (PDataRecord l) where
+  type PInner (PDataRecord l) _ = PBuiltinList PData
+  pcon' (PDCons x xs) = pcons # pforgetData x # pto xs
+  pcon' PDNil         = pnil
+  pmatch' _ _ = undefined -- FIXME
```
L-as, line 114: You could just do `error "you need a more specific PDataRecord"` lol

TotallyNotChase: that's exactly what I've been leaning towards after trying out a bunch of things that didn't work....

that might honestly be the best thing to do right now. This instance is only really useful for the `pto` anyway.

Plutarch/DataRepr/Internal.hs
```diff
@@ -108,6 +107,12 @@ data PDataRecord (as :: [PLabeledType]) (s :: S) where
     PDataRecord ((name ':= x) ': xs) s
   PDNil :: PDataRecord '[] s
 
+instance {-# OVERLAPPABLE #-} PlutusType (PDataRecord l) where
+  type PInner (PDataRecord l) _ = PBuiltinList PData
+  pcon' (PDCons x xs) = pcons # pforgetData x # pto xs
+  pcon' PDNil         = pnil
```
L-as, line 113: these should probably also err

L-as: or maybe not, doesn't seem like they are any less efficient? anyway, I think you can actually do `pcon' (PDCons x xs) = pcon' (PDCons x xs)` and have it work

L-as (changes requested):

Plutarch/DataRepr/Internal.hs
L-as: You should probably add type annotations to make it clear what's happening

Plutarch/DataRepr/Internal.hs
L-as: in a dependently typed language we could pattern match on the `l` probably

TotallyNotChase: it's ok I heard GHC is landing `DependentHaskell` tomorrow

Plutarch/DataRepr/Internal.hs
TotallyNotChase: added type annotations, also made it use the underlying `pcon'` instead - good idea!

TotallyNotChase:
'tis ready

L-as (approved):
LGTM!

PR #422, opened 2022-04-06
minor error correction in Plutarch Lambdas

pb99u069:


srid:
@TotallyNotChase How are changes to `master` docs integrated to that of `staging`? Is that a straightforward merge? Because, IIRC, we changed `staging` docs substantially.

TotallyNotChase (approved):
Thanks!

PR #424, opened 2022-04-07
plutarch-extra: `findOwnInput`, `getContinuingOutputs`, `findDatum`

srid:
Upstream these functions from `plutus-ledger-api`, along with golden and assertion tests for them. It also adds a `Plutarch.Extra.Monad` as these functions use it; should they be put in TermCont.hs instead?

Note: we are coming to the point of "stealing" mock `ScriptContext`s from nearby specs, but I think it would be useful to create a separate module for it exporting a cleaner API, possibly with a well-populated `ScriptContext`.  It could happen in another PR. I also wonder if we can reuse Haskell tests from the Plutus side via [`prop_haskEquiv`](https://github.com/Plutonomicon/plutarch/blob/4052b285eb890799332c0cbe19cb08c1070f267a/plutarch-test/common/Plutarch/Test/Property/HaskEquiv.hs#L106-L123).

- [x] Optimize the original implementation of these functions
- [x] Naming: `findOwnInput` or `pfindOwnInput`?


TotallyNotChase (commented):
A general note: I don't know if we have a policy on this - but efficient utility functions should have their input types as focused as possible. For example, `findDatum` should take in the datums field, not the full `PTxInfo` to deconstruct etc.

I generally believe mirroring the Plutus ledger api is a bad idea for idiomatic UPLC generation, but this is completely upto what we end up deciding for `plutarch-extra` - i.e whether or not we really need to focus on efficiency.

srid:
That's a good point. But there is another approach to this problem. Since a `ScriptContext` is read-only and never changes during validation, could we optimize "lookups" in Plutarch? Sort of like hoisting lambdas, but now we hoist these "lookups" - and reference them once. cf. https://github.com/Plutonomicon/plutarch/issues/260#issuecomment-1090501474

L-as:
That is just CSE. We'd have to find a way to do the analysis while preserving semantics.

srid:
The PR is ready. Let me know what the naming should be. 

I'll do the improvements to the tests (stated in the PR description) separately.

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,90 @@
...
+  PTxInfo,
+  PTxOut,
+  PTxOutRef,
+ )
+import Plutarch.Prelude
+
+import Plutarch.Extra.Monad (tmatch)
+
+-- | Find the output txns corresponding to the input being validated.
+getContinuingOutputs :: Term s (PScriptContext :--> PBuiltinList PTxOut)
```
TotallyNotChase, line 26: If going for the "maximally focused input" route, this function should take the OutRef from `Spending _` and the `outputs` field instead of the whole context.

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,90 @@
...
+    matches = phoistAcyclic $
+      plam $ \adr txOut ->
+        adr #== pfield @"address" # txOut
+
+{- | Find the input currently being validated.
+
+  Tries to finds the transaction's input by looking for a `PTxInInfo` in the inputs
+  coresponding to the `PTxOutRef` which the script purpose is spending
+-}
+findOwnInput :: Term s (PScriptContext :--> PMaybe PTxInInfo)
```
TotallyNotChase, line 49: If going for the "maximally focused input" route, this should take the inputs field and the outref

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,90 @@
...
+        pure $ pcon PNothing
+  where
+    matches :: Term s (PTxOutRef :--> PTxInInfo :--> PBool)
+    matches = phoistAcyclic $
+      plam $ \outref txininfo ->
+        pfield @"id" # outref
+          #== pfield @"id" # (pfield @"outRef" # txininfo)
+
+-- | Find the data corresponding to a data hash, if there is one
+findDatum :: Term s (PDatumHash :--> PTxInfo :--> PMaybe PDatum)
```
TotallyNotChase, line 71: If going for the "maximally focused input" route, this should take in the datum assoc list instead of the whole txInfo

plutarch-extra/Plutarch/Extra/Monad.hs
```diff
@@ -0,0 +1,18 @@
...
+-- | TermCont-related adapters for Plutarch functions.
+module Plutarch.Extra.Monad (
+  tlet,
+  tmatch,
+) where
+
+import Plutarch.Prelude
+
+-- | Like `plet` but works in a `TermCont` monad
+tlet :: Term s a -> TermCont s (Term s a)
```
TotallyNotChase, line 13: I think we should keep using the `p` as prefix for Plutarch stuff. I tend to call termcont utils like `pletC`, `pletFieldsC`, `pmatchC` etc. But feel free to choose something else - as long as it starts with `p`.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
```diff
@@ -202,3 +213,54 @@ getFields = phoistAcyclic $ plam $ \addr -> psndBuiltin #$ pasConstr # addr
 
 dummyCurrency :: CurrencySymbol
 dummyCurrency = Value.currencySymbol "\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11\x11"
+
+------------------- Mocking a ScriptContext ----------------------------------------
+
+validContext0 :: Term s PScriptContext
```
TotallyNotChase, line 219: I think these mocks should be put into a common module that's not a `*Spec`

TotallyNotChase (changes requested):
Mostly concerned about efficiency. But I'm fine if we decide to not care much about that for `plutarch-extra`.

Also yes the naming should use a `p` prefix I'd suggest.

plutarch-test/plutarch-base/Plutarch/ApiSpec.hs
srid: Yes, I will do this next, but I want to do it in a separate PR (#433) so as to keep this one small and scoped.

srid:
After the brief discussion on Discord, we decided to go with the "maximally focused input" approach. See 864e325.

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,107 @@
...
+
+  Takes as argument the datum assoc list from a `PTxInfo`.
+
+  __Example:__
+
+  @
+  pfindDatum # datumHash #$ pfield @"datums" # txinfo
+  @
+-}
+pfindDatum :: Term s (PDatumHash :--> PBuiltinList (PAsData (PTuple PDatumHash PDatum)) :--> PMaybe PDatum)
```
TotallyNotChase, line 91: I wonder if it'd be more practically useful to have something like `pparseDatum :: Term s (PAsData PDatumHash :--> PBuiltinList (PAsData (PTuple PDatumHash PDatum)) :--> PAsData a)` instead. I imagine users will generally want `ptryFrom` the returned `PDatum` immediately.

Also making that `PDatumHash` a `PAsData` and using data equality could reduce a few `pfromData` calls.

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,107 @@
...
+  @
+-}
+pgetContinuingOutputs :: Term s (PBuiltinList (PAsData PTxInInfo) :--> PBuiltinList (PAsData PTxOut) :--> PTxOutRef :--> PBuiltinList PTxOut)
+pgetContinuingOutputs = phoistAcyclic $
+  plam $ \inputs outputs outRef ->
+    pmatch (pfindOwnInput # inputs # outRef) $ \case
+      PJust tx -> do
+        let resolved = pfield @"resolved" # tx
+            outAddr = pfield @"address" # resolved
+        pfilter # (matches # outAddr) #$ pmap # plam pfromData # outputs
```
TotallyNotChase, line 44: this should be a singular `pfilter` yeah? that `pmap` is morally a no-op.

plutarch-extra/Plutarch/Extra/Api.hs
```diff
@@ -0,0 +1,107 @@
...
+          inputs = pfield @"inputs" # (getField @"txInfo" ctx)
+      pure $ pfindOwnInput # inputs # outRef
+    _ ->
+      pure $ ptraceError "not a spending tx"
+  @
+-}
+pfindOwnInput :: Term s (PBuiltinList (PAsData PTxInInfo) :--> PTxOutRef :--> PMaybe PTxInInfo)
+pfindOwnInput = phoistAcyclic $
+  plam $ \inputs outRef ->
+    pfind # (matches # outRef) #$ pmap # plam pfromData #$ inputs
```
TotallyNotChase, line 73: I think this should be a singular `pfind`

plutarch-extra/Plutarch/Extra/Api.hs
TotallyNotChase, line 91: I wonder if it'd be more practically useful to have something like `pparseDatum :: Term s (PAsData PDatumHash :--> PBuiltinList (PAsData (PTuple PDatumHash PDatum)) :--> PAsData a)` instead. I imagine users will generally want `ptryFrom` the returned `PDatum` immediately.

Also making that `PDatumHash` a `PAsData` and using data equality could reduce a few `pfromData` calls.

TotallyNotChase (approved):
Looks quite nice in general - just a few optimization concerns that don't necessarily need to be resolved in this PR.

plutarch-extra/Plutarch/Extra/Api.hs
srid: @TotallyNotChase How would that be a no-op? Where else would you `pfromData` the elements of `outputs`?

Or do you mean making this function return ` PBuiltinList (PAsData PTxOut)`?

plutarch-extra/Plutarch/Extra/Api.hs
srid: > I imagine users will generally ptryFrom the returned PDatum immediately.

Do you have an example code for this?

plutarch-extra/Plutarch/Extra/Api.hs
TotallyNotChase: > Or do you mean making this function return PBuiltinList (PAsData PTxOut)?
yep!

plutarch-extra/Plutarch/Extra/Api.hs
TotallyNotChase: For the usage? I'd imagine something like (rough writeup)-

```hs
PJust x' <- pmatchC $ pfindDatum # h # l
x <- fst <$> ptryFrom @SomeType $ pto x'
```

For the implementation of a `pparseDatum`-like function, perhaps something like (rough writeup)-

```hs
pfindDatum :: forall a s. PTryFrom PData (PAsData a) => Term s (PAsData PDatumHash :--> PBuiltinList (PAsData (PTuple PDatumHash PDatum)) :--> PMaybe (PAsData a))
pfindDatum = phoistAcyclic $
  plam $ \datumHash datumMap -> unTermCont $ do
    datmMaybe <- tcont . pmatch $ pfind # plam (\x -> pfield @"_0" # x #== datumHash) # datumMap
    case datmMaybe of
      PNothing -> pure $ pcon PNothing
      PJust tupl ->
        let
          g :: Term _ PData
          g = pto . pfromData $ pfield @"_1" # pfromData tupl
        in pcon . PJust . fst <$> tcont (ptryFrom @(PAsData a) $ pto . pfromData $ pfield @"_1" # pfromData tupl)
```

plutarch-extra/Plutarch/Extra/Monad.hs
```diff
@@ -0,0 +1,18 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | TermCont-related adapters for Plutarch functions.
+module Plutarch.Extra.Monad (
```
TotallyNotChase, line 5: Sorry, I forgot to comment this previously - but I think this module should be named `Plutarch.Extra.TermCont`.

We should definitely make note so as to add more termcont utils here later: `pletFieldsC`, `ptraceC` and some more come to mind.

TotallyNotChase: btw what's up with the `UndecidableInstances` pragma?

TotallyNotChase (dismissed):
beautiful!

plutarch-extra/Plutarch/Extra/Monad.hs
srid: From previous code I guess. Removed.

L-as (approved):
LGTM

PR #299, opened 2022-02-16
Throw error on division by 0 for rationals

Geometer1729:
I noticed down stream that rationals weren't throwing an error on division by 0 like I expected, so I added some tests and fixed it so they do.

PR #427, opened 2022-04-08
Throw error on division by 0 for rationals (#299) - continuation

TotallyNotChase:
This is a continuation of #299. It should be ready to merge now.


plutarch-test/plutarch-base/Plutarch/ShowSpec.hs
```diff
@@ -11,69 +11,71 @@ import Test.Hspec
 
 spec :: Spec
 spec = do
-  describe "show" . pgoldenSpec $ do
```
TotallyNotChase, line 14: the only thing that was changed here is that the `let str x = pconstant @PString x` got hoisted upwards and `describe "show"` now houses 2 separate `pgoldenSpec`s, one has `plutarchDevFlagDescribe`

L-as (dismissed):
LGTM

PR #430, opened 2022-04-09
Explicit type applications for local binds

TotallyNotChase:
I'm surprised GHC 8 managed to fail due to this even though we already have `MonoLocalBinds` turned on. I guess the type inference didn't "spread" as well as GHC 9.

Also I'm pretty sure type family deps can fix this altogether.

### DIFF
--- a/plutarch-test/plutarch-base/Plutarch/PIsDataSpec.hs
+++ b/plutarch-test/plutarch-base/Plutarch/PIsDataSpec.hs
@@ -81,10 +81,10 @@ spec = do
         "4wheeler" @\ do
           let datrec =
                 pdcons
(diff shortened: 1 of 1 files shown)


L-as:
Why wasn't this necessary before?

TotallyNotChase:
> Why wasn't this necessary before?

These local binds were added in the `PDataSum` PlutusType PR. Before, they were simply inlined at the use site and GHC had no trouble figuring it out. Now, they are hoisted out to local bindings because they are used in 2 different places.

I'm sure these wouldn't error if they weren't hoisted out to be local bindings and just inlined instead but code duplication.

L-as:
Makes sense.

L-as (approved):

Issue #82, opened 2022-01-07
Support emitting PIR


L-as:
Some users want to inspect the generated code, and that would be much easier if types were emitted.

Copy-pasted from internal slack:

Let's say we have the identity function on integers:
pid' :: Term s (PInteger :--> PInteger)
pid' = plam $ \x -> x
This currently does not store any information about the type of x in the generated code.
With the current design, it's impossible since Haskell types are erased at runtime. We don't know whether it's PInteger or PString.  We could add a type class to plam to put that type into the generated code, but this solution stops working in the polymorphic case:
pid :: Term s (a :--> a)
pid = plam $ \x -> x
This wouldn't work because we don't have an instance for e.g. PType a . We could add that constraint, but now we have another problem:
Each time you instantiate this function, you will instantiate it with a different concrete type, i.e. monomorphisation will happen. This will mean that in the generated code, we will have many pids, each for a specific type.

What we really want is to generate a forall in the generated code. We could then have pforall as such:
pid :: Term s (PForall $ \a -> a :--> a)
pid = pforall $ \a -> plam $ \(x :: Term s a) -> x
There's however a huge issue here: Haskell doesn't have dependent types. We can not refer to the a in the definition of pid in the type of x. We can likely work around this with singletons and other madness, but it would probably turn into something so unergonomic that no one would use it.

TotallyNotChase:
I think the goal of preserving types may be a detriment to what Plutarch aims to be - an efficient plutus core generator. In the end, this might just lead to recreating Plutus Tx - with all its deficiencies.

Saizan:
Regarding the problem of binding `a`, in principle it can be done without singletons (with `ScopedTypeVariables`), though not in a very ergonomic way either:
```haskell
pid :: forall s. Term s (PForall $ \a -> a :--> a)
pid = pforall body
 where
   body :: forall a. PType a => Term s (a :--> a)
   body = plam $ \(x :: Term s a) -> x
```
However I would expect that if we give top level signatures, like the one for `pid`, we will not need to provide annotations like the one for `x` very often at all, so this might not be a big issue.

However Haskell does not have type lambdas either, so it's not clear how to write `PForall $ \a -> a :--> a`

Regarding retaining efficient untyped code I believe we would have to consider PIR and untyped plutus as different "compilation" targets. As in maybe under the hood Term is implemented as a pair of a typed and an untyped representation, with the same functional correctness semantics, but the untyped one is optimized.

[...]

PR #434, opened 2022-04-12
`PSubtype`

L-as:


L-as:
Now that I think about it, maybe this should be split into two classes.
We want `PSubtype PData (PAsData a)`, but that can't happen here because `pdowncast'` isn't implementable.
If we had a separate `PDowncast`, we could implement `pupcast` and `pdowncast'` separate essentially.

Plutarch/TryFrom.hs
```diff
@@ -318,25 +329,6 @@ instance {-# OVERLAPPING #-} SumValidation n '[] where
...
-  , PIsData b
-  ) =>
-  PTryFrom (PAsData a) (PAsData b)
-  where
-  type PTryFromExcess (PAsData a) (PAsData b) = PTryFromExcess a b
-  ptryFrom' opq = runTermCont $ do
-    ver' <- snd <$> TermCont (ptryFrom @b @a (pfromData opq))
-    pure $ (punsafeCoerce opq, ver')
-
-instance PTryFrom PData PData where
```
L-as, line 336: This instance doesn't make sense.

Plutarch/TryFrom.hs
```diff
@@ -318,25 +329,6 @@ instance {-# OVERLAPPING #-} SumValidation n '[] where
...
 
-{- | if there is an instance to recover something that is unwrapped from something that is
- unwrapped, then there is also the possibility to recover the whole thing but wrapped
--}
-instance
-  ( PTryFrom a b
-  , PIsData a
-  , PIsData b
-  ) =>
-  PTryFrom (PAsData a) (PAsData b)
```
L-as, line 329: This was illegal. Even if it's true that `PSubtype a b`, this does **not** imply `PSubtype (PAsData a) (PAsData b)`, and was also a bug before this change!

L-as: Notably, you could have `PInteger` and `PNatural` have completely different `PIsData` implementations.

L-as:
This should be ready.

L-as:
I didn't remove `pto`, because `pto` is still more powerful than `pupcast` in some cases unfortunately. Currently, `PSubtype` doesn't allow expressing the type polymorphism that `PInner` allows, hence `pto` is still necessary.
I could have made the subtyping relation just as powerful, but it would be annoying for the general case and be quite complex.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
```diff
@@ -487,7 +489,9 @@ theField = unTermCont $ do
...
 
-newtype PWrapInt (s :: S) = PMkWrapInt (Term s PInteger)
-  deriving (PlutusType, PIsData, PEq, POrd) via (DerivePNewtype PWrapInt PInteger)
+newtype PWrapInt (s :: S) = PWrapInt (PInteger s)
+  deriving newtype (PIsData, PEq, POrd)
 
-deriving via DerivePNewtype (PAsData PWrapInt) (PAsData PInteger) instance PTryFrom PData (PAsData PWrapInt)
+instance PTryFrom PData (PAsData PWrapInt) where
+  type PTryFromExcess PData (PAsData PWrapInt) = PTryFromExcess PData (PAsData PInteger)
+  ptryFrom' t f = ptryFrom' t $ \(t', exc) -> f (coerce (t' :: Term _ (PAsData PInteger)), exc)
```
MangoIV, line 497: Why do we not want deriving via `DerivePNewtype` anymore?

Plutarch/TryFrom.hs
MangoIV: We don't need to remove it completely, we should just not punsafeCoerce the opq but pdata the uwrapped version. it's probably helpful in many cases anyway because semantically it's correct, isn't it?

Plutarch/TryFrom.hs
```diff
@@ -1,368 +1,77 @@
...
-          the expected type.
-      - the operation `ptryFrom @b @a` proves equality between the less expressive `PType` `a` and
-        the more expressive `PType` `b`, hence the first element of the resulting Tuple
-        must always be wrapped in `PAsData` if the origin type was `PData` (see law 1)
-      - the result type `b` must always be safe than the origin type `a`, i.e. it must carry
-        more information
+@PTryFrom a b@ represents a subtyping relationship between @a@ and @b@,
+and a way to go from @a@ to @b@.
+Laws:
+- @(punsafeCoerce . fst) <$> tcont (pdowncast x) ≡ pure x@
```
MangoIV, line 55: this states the law wrt pdowncast, is that intended?

plutarch-test/conditional/Plutarch/TryFromSpec.hs
L-as: It didn't work and was in fact illegal.

Plutarch/TryFrom.hs
L-as: nope

Plutarch/TryFrom.hs
L-as: That would be just as illegal as it violates the laws of `PTryFrom`.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
MangoIV: I think we should rather just add a warning that this leads to issues if we instantiate DerivePNewtype incorrectly, I think that this instance is really helpful in practise 🤷

Plutarch/TryFrom.hs
MangoIV: That's a pity, it shouldn't really break the laws of PTryFrom tbh, I think many people would semantically expect that 
PSubType (PAsData PInteger) (PAsData PNatural). But I guess they can just add this instance themselves.

L-as: You can have an instance in specifically that case, but it's not really possible to have a general purpose reasonable law that allows such a general instance. The return of `ptryFrom` *must* be the same as the input, with only the type changed.

L-as: And anyway, it would still violate the `PSubtype` constraint.

Plutarch/TryFrom.hs
```diff
@@ -1,368 +1,77 @@
...
+ This concretely means that `\(x :: Term s b) -> punsafeCoerce x :: Term s a`
+ is legal and sound.
 
-import Data.Functor.Const (Const)
+ You can not make an instance for this yourself.
+ You must use the 'PInner' type family of 'PlutusType' to get this instance.
 
-import Data.Kind (Type)
+ Caveat: Only @PInner a POpaque@ is considered unfortunately, as otherwise
+ getting GHC to figure out the relation with multiple supertypes is quite hard.
```
TotallyNotChase, line 44: The diamond problem? I feel like we shouldn't allow multiple supertypes but then again a fundep would be too restrictive as well. I think Arc was actually talking about this exact issue and he managed to solve it with the dual typeclasses AFAIK - but that didn't work in this case without putting GHC into a loop?

I'm concerned about the type inference and potential complexity.

Plutarch/DataRepr/Internal.hs
```diff
@@ -627,3 +646,122 @@ class IsBuiltinList a where
...
+type HRecPApply :: [(Symbol, PType)] -> S -> [(Symbol, Type)]
+type family HRecPApply as s where
+  HRecPApply ('(name, ty) ': rest) s = '(name, Reduce (ty s)) ': HRecPApply rest s
+  HRecPApply '[] s = '[]
+
+newtype HRecP (as :: [(Symbol, PType)]) (s :: S) = HRecP (HRecGeneric (HRecPApply as s))
+
+instance Reducible (HRecP as s) where type Reduce (HRecP as s) = HRecGeneric (HRecPApply as s)
+
+newtype Flip f a b = Flip (f b a)
```
TotallyNotChase, line 661: We have these helper types in `.Builtin` as well right?

plutarch-test/conditional/Plutarch/TryFromSpec.hs
```diff
@@ -487,7 +489,9 @@ theField = unTermCont $ do
 
 ------------------- Sample usage DerivePNewType ------------------------------------
 
-newtype PWrapInt (s :: S) = PMkWrapInt (Term s PInteger)
-  deriving (PlutusType, PIsData, PEq, POrd) via (DerivePNewtype PWrapInt PInteger)
+newtype PWrapInt (s :: S) = PWrapInt (PInteger s)
```
TotallyNotChase, line 492: different newtyping style?

TotallyNotChase (commented):
I was under the impression `PTryFrom` was going to be replaced with `PSubtype` but we ended up having both. I guess there's no way to merge them?

More importantly, it'd be nice to have some examples and a topic in the guide for this.

Plutarch/TryFrom.hs
TotallyNotChase: Wait, so is there no "easy" way for users to get a `PTryFrom` instance for types like `CurrencySymbol` (and other "variant-less" newtypes)?

Plutarch/TryFrom.hs
L-as: I've decided to drop multiple supertypes entirely for the sake of #82

Plutarch/DataRepr/Internal.hs
L-as: Yeah, they're quite small, and more
similar to a lambda expression than a function definition. Not sure whether it's the right choice to share them.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
L-as: I tried to do this to keep the derive intact but it didn't work.

Plutarch/TryFrom.hs
L-as: See comment by Emily at #438

TotallyNotChase: Yeah that looks good - we should have that instance if we don't already.

plutarch-test/conditional/Plutarch/TryFromSpec.hs
TotallyNotChase: Could it be changed back then? To something like this potentially:-

```hs
newtype PWrapInt (s :: S) = PWrapInt (Term s PInteger)
  deriving (PlutusType, PIsData, PEq, POrd) via (DerivePNewtype PWrapInt PInteger)

instance PTryFrom PData (PAsData PWrapInt) where
  type PTryFromExcess PData (PAsData PWrapInt) = PTryFromExcess PData (PAsData PInteger)
  ptryFrom' t f = ptryFrom' t $ \(t', exc) -> f (pouterData t', exc)

pouterData :: (PIsData a, PIsData (PInner a POpaque)) => Term s (PAsData (PInner a POpaque)) -> Term s (PAsData a)
pouterData = pdata . punsafeDowncast . pfromData
```

Though there should really be an easy deriver for cases like this. I'm fine it naming it "UnsafeNewtypeDeriver" or something but it should be there.

L-as: Given that we will have to redo it in #441 anyway, I propose we leave it to another PR.

L-as: wrt. the other part, why change it back?

L-as: also pouterData there is unsafe

TotallyNotChase: If a new style is unnecessary, it should just use the same newtyping style as other places in the codebase - for the sake of consistency.

`pouterData` is indeed unsafe in the general sense, which is why I suggested "UnsafeNewtypeDeriver". It's morally safe in this invariant-less case though.

L-as: I think it's better to do it this way since we don't need punsafeCoerce that way.

L-as: actually wait...

L-as: currently `type role PAsData representational phantom`, so if `Coercible a b`, you have `Coercible (PAsData a) (PAsData b)`, which is in fact completely incorrect! lol

L-as: I will fix this and then I think we can merge so we can work on #446 ?

TotallyNotChase: Time for a release cycle sprint within a release cycle sprint. I hope we come out the other side with across-the-board improvements.

L-as:
Merging as the `PWrapInt` issue can't be fixed properly anyway

PR #435, opened 2022-04-13
Delegate `Rational` invariant to `PNonZero` + optimizations

TotallyNotChase:
The "fail on zero" invariant should be forced in the `PlutusType` instance - otherwise `pcon` is a trap for users.

The golden changes are purely within dev mode and only exist due to a slight change in the error trace.


Geometer1729 (dismissed):
LGTM

L-as:
I'm not sure this is the best solution. Shouldn't we just not export the constructors, or otherwise make the divisor (or dividend?) a Positive?

TotallyNotChase:
I think `pcon` is a pretty good notion of smart constructor that already exists. Unless we require `pcon` to never error - I don't see a reason to not utilize it for this purpose.

We could indeed just not export the constructor but then the `PlutusType` instance becomes a bit useless - unless you go on to introduce "safe" helper functions to help with `pmatch`'s handler and `pcon`'s argument, which seems like unnecessary extra work.

L-as:
The `PlutusType` instance would still be used internally.

in general, in Haskell, people expect constructors to be total. I don't think we should add partial constructors.

TotallyNotChase:
Not wanting `pcon` to be partial is a valid concern - though I don't know how applicable it is for Plutarch usecases. Do you think we should unexport the constructor and just have 2 functions for building them (one returning `PMaybe`, the other being partial)?

And for deconstruction, I suppose the `pnominator` and `pdenominator` will do? I feel like this might be a blocker for achieving maximal performance in userland (using `pmatch` directly in a utility function that is then hoisted will be more efficient) - but oh well.

L-as:
Well that's why I suggested Positive.

TotallyNotChase:
I assume you mean `NonZero`? Hmm, I could try that and see what it entails.

Plutarch/Rational.hs
```diff
@@ -21,121 +22,131 @@ import Plutarch (
...
 data PRational s
-  = PRational (Term s PInteger) (Term s PInteger)
+  = PRational (Term s PInteger) (Term s PNonZero)
   deriving stock (GHC.Generic)
-  deriving anyclass (Generic, PlutusType)
+  deriving anyclass (Generic, PEq)
+
+instance PlutusType PRational where
+  type PInner PRational b = (PInteger :--> PNonZero :--> b) :--> b
+  pcon' (PRational x y') = plet y' $ \y -> plam $ \f -> f # x # y
```
TotallyNotChase, line 57: I had to make sure the denominator was evaluated prior to building the lambda. Otherwise, you could make a `n / 0` and the error would only happen upon actually matching on the rational.

It's inefficient sometimes - so there's also `pratLazy` below.

L-as:
Shouldn't it be `Positive` and not `NonZero`? I don't see why it would be negative.

TotallyNotChase:
> Shouldn't it be `Positive` and not `NonZero`? I don't see why it would be negative.

The original invariant was `NonZero` so far so I preserved that behavior. I don't see anything technically wrong with allowing negative denominators but we can definitely change it to `Positive` if necessary.

Plutarch/NonZero.hs
```diff
@@ -0,0 +1,44 @@
...
+pnonZero :: Term s (PInteger :--> PMaybe PNonZero)
+pnonZero = phoistAcyclic $
+  plam $ \i ->
+    pif
+      (i #== 0)
+      (pcon PNothing)
+      $ pcon . PJust . pcon $ PNonZero i
+
+-- | Partial version of 'pnonZero'. Errors if argument is zero.
+pnonZero' :: Term s (PInteger :--> PNonZero)
```
L-as, line 38: We have the not-so-followed practice of using `ptry` for partial functions.

L-as (changes requested):
We need to find a general solution to this.

Plutarch/Rational.hs
TotallyNotChase: For this PR, should I just remove the `PlutusType` instance and have 2 construction functions? (one without the invariant check, and one with it)

Plutarch/NonZero.hs
TotallyNotChase: updated

TotallyNotChase:
I removed the constructor export and kept the `PlutusType` instance "normal" (no forced invariant check), and added supplementary functions for construction and deconstruction

Plutarch/Reducible.hs
```diff
@@ -15,3 +15,8 @@ instance Reducible a => Reducible (Const a b) where
 
 instance Reducible a => Reducible (Identity a) where
   type Reduce (Identity a) = Reduce a
+
+newtype Flip f a b = Flip (f b a)
```
L-as, line 19: I didn't do this on purpose. `Plutarch.Reducible` shouldn't export any helper types, since they're very often ad-hoc, and aren't relevant to the end-user anyway. They're completely internal and are not part of the interface of Plutarch. The `Reducible` instance is generically derived on my local machine, I'll push that later.

TotallyNotChase: It seems weird to have to copy/paste the same block of boilerplate code in 3 separate files in the same project. Though derivation helpers would indeed make it a bit better.

L-as: it would only be one line then

Plutarch/Rational.hs
```diff
@@ -38,106 +41,128 @@ import Plutarch.Builtin (
...
   deriving stock (GHC.Generic)
-  deriving anyclass (Generic, PlutusType)
+  deriving anyclass (Generic, PEq)
+
+{- | NOTE: This instance may delegate the nonzero invariant check during construction.
+
+As such, the constructor is not exported and users are expected to use 'prational', 'prationalLazy', and
+'punrational' instead.
+-}
+instance PlutusType PRational where
```
L-as, line 63: This should not have a manual implementation

Plutarch/Rational.hs
```diff
@@ -38,106 +41,128 @@ import Plutarch.Builtin (
...
+  type PInner PRational b = (PInteger :--> PNonZero :--> b) :--> b
+  pcon' (PRational x y) = plam $ \f -> f # x # y
+  pmatch' ratF f = ratF #$ plam $ \a b -> f $ PRational a b
+
+{- | Builds a rational with lazy non-zero denominator verification.
+
+i.e, if the denominator was zero, it'll error only after 'pmatch'ing on the resulting rational.
+-}
+prationalLazy :: Term s PInteger -> Term s PNonZero -> Term s PRational
```
L-as, line 72: This isn't lazy anymore

Plutarch/Rational.hs
```diff
@@ -38,106 +41,128 @@ import Plutarch.Builtin (
...
+      plistToRat = plam $ \x -> prationalLazy (pasInt #$ phead # x) $ punsafeDowncast $ pasInt #$ phead #$ ptail # x
   pdataImpl x' =
     phoistAcyclic
-      ( plam $ \x ->
-          punsafeCoerce (pdata $ pRatToList # x :: Term _ (PAsData (PBuiltinList (PAsData PInteger))))
+      ( plam $ \x -> unTermCont $ do
+          PRational a b <- tcont $ pmatch x
+          let res :: Term _ (PBuiltinList (PAsData PInteger))
+              res = pcons # pdata a #$ pcons # pdata (pto b) #$ pnil
+          pure $ punsafeCoerce $ pdata res
```
L-as, line 101: Why are we using `punsafeCoerce` here

Plutarch/Rational.hs
```diff
@@ -169,60 +194,48 @@ instance Num (Term s PRational) where
...
+  fromInteger n = prationalLazy (pconstant n) $ punsafeDowncast 1
 
 instance Fractional (Term s PRational) where
   recip x' =
     phoistAcyclic
       ( plam $ \x ->
-          pmatch x $ \(PRational xn xd) ->
-            pfailOnZero # xn # (pcon (PRational xd xn))
+          pmatch x $ \(PRational xn xd) -> plet (ptryNonZero # xn) $ \denm ->
+            prationalLazy (pto xd) denm
```
L-as, line 206: Bad code?

L-as:
@TotallyNotChase Can you make this PR up to date?

TotallyNotChase:
FYI this isn't ready yet - just syncing with staging.

Plutarch/NonZero.hs
```diff
@@ -0,0 +1,77 @@
...
+
+newtype PNonZero s = PNonZero (Term s PInteger)
+  deriving stock (Generic)
+  deriving anyclass (PlutusType, PIsData, PEq, POrd, PIntegral, PShow)
+instance DerivePlutusType PNonZero where type DPTStrat _ = PlutusTypeNewtype
+
+instance PNum PNonZero where
+  x #- y = ptryNonZero #$ pto x #- pto y
+
+  pfromInteger 0 = error "pnonZero.pfromInteger: encountered 0"
```
TotallyNotChase, line 44: This currently errors at Plutarch compile time - I feel like there should be a way to produce `Term`s that are just descriptive `CompilerError`s (raised by `compile` - eventually).

Plutarch/NonZero.hs
```diff
@@ -0,0 +1,77 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Plutarch.NonZero (PNonZero, pnonZero, ptryNonZero) where
```
TotallyNotChase, line 4: I assume the current convention to enforce invariants for types is to make sure the constructor is not exported? You can't have manual invariant checks in `PlutusType` anymore.

If so, `PValue`'s constructor needs to be unexported.

Plutarch/NonZero.hs
L-as: pthrow

Plutarch/NonZero.hs
L-as: yep

TotallyNotChase:
I'm trying to figure out optimization stops - where invariant checks should be skipped.

L-as:
Just put them in everywhere, then we'll make it efficient later.

Plutarch/NonZero.hs
TotallyNotChase: updated

Plutarch/Rational.hs
TotallyNotChase: updated to use `pdataImpl` now

Plutarch/Rational.hs
TotallyNotChase: removed, will figure out optimizations later as you said

TotallyNotChase:
Alright, will do optimizations later then

TotallyNotChase:
> Hmm, why did the `PNonZero` -> `PPositive` change make the benchmarks perform worse? There wasn't really any changes other than `#==` becoming `#<=` - should investigate.

looks like the benchmarks are performed on scripts compiled with tracing mode on - that's not correct.

Plutarch/Rational.hs
```diff
@@ -44,120 +47,142 @@ import Plutarch.Builtin (
...
   (#/) :: Term s a -> Term s a -> Term s a
   precip :: Term s (a :--> a)
   pfromRational :: Term s (PRational :--> a)
 
 data PRational s
-  = PRational (Term s PInteger) (Term s PInteger)
+  = PRational (Term s PInteger) (Term s PPositive)
   deriving stock (Generic)
-  deriving anyclass (PlutusType)
+  deriving anyclass (PlutusType, PEq)
```
L-as, line 70: Why are we deriving `PEq`? This doesn't seem correct.

Plutarch/Rational.hs
```diff
@@ -44,120 +47,142 @@ import Plutarch.Builtin (
...
 
 instance (PNum a, PFractional a) => Fractional (Term s a) where
   (/) = (#/)
   recip x = precip # x
-  fromRational x = pfromRational #$ pcon $ PRational (pconstant (numerator x)) (pconstant (denominator x))
+  fromRational x =
+    pfromRational #$ pcon $
+      PRational
+        (pconstant $ numerator x)
+        (punsafeDowncast . pconstant $ denominator x)
```
L-as, line 81: Are we sure `denominator` is always positive?

Plutarch/Rational.hs
```diff
@@ -44,120 +47,142 @@ import Plutarch.Builtin (
...
 
-pRatToList :: Term s (PRational :--> PBuiltinList (PAsData PInteger))
-pRatToList = plam $ \x -> pmatch x $ \(PRational a b) ->
-  pcons # pdata a #$ pcons # pdata b #$ punsafeCoerce (pnil :: Term s (PBuiltinList PData))
+newtype Flip f a b = Flip (f b a) deriving stock (Generic)
 
-pListToRat :: Term s (PBuiltinList PInteger :--> PRational)
-pListToRat = plam $ \x -> pcon $ PRational (phead # x) (phead #$ ptail # x)
-
-instance PEq PRational where
```
L-as, line 97: I think we still need this.

Plutarch/Rational.hs
```diff
@@ -44,120 +47,142 @@ import Plutarch.Builtin (
...
-              rd * ln #== rn * ld
-      )
-      # l'
-      # r'
+-- | NOTE: This instance produces a verified 'PPositive' as the excess output.
+instance PTryFrom PData (PAsData PRational) where
+  type PTryFromExcess PData (PAsData PRational) = Flip Term PPositive
+  ptryFrom' opq = runTermCont $ do
+    (_, ld) <- tcont $ ptryFrom @(PAsData (PBuiltinList PData)) opq
+    tcont $ \f -> pif (plength # ld #== 2) (f ()) (ptraceError "ptryFrom(PRational): data list length should be 2")
```
L-as, line 117: This isn't efficient, but we can leave it for later.

L-as: Add a TODO

Plutarch/Rational.hs
```diff
@@ -44,120 +47,142 @@ import Plutarch.Builtin (
...
+          pure $
+            preduce
+              #$ pcon
+              $ PRational (xn * pto yd + yn * pto xd) $
+                punsafeDowncast $ pto xd * pto yd
       )
       # x'
       # y'
 
   x' #- y' =
```
L-as, line 160: Couldn't we implement this efficiently in terms of `#+`? Add a TODO.

Plutarch/Rational.hs
```diff
@@ -192,48 +217,36 @@ instance PFractional PRational where
     phoistAcyclic $
       plam $ \x ->
         pmatch x $ \(PRational xn xd) ->
-          pfailOnZero # xn # pcon (PRational xd xn)
+          pcon $ PRational (pto xd) $ ptryPositive # xn
 
   x' #/ y' =
```
L-as, line 222: We can implement this in terms of `#*`. Add a TODO please.

Plutarch/Rational.hs
```diff
@@ -192,48 +217,36 @@ instance PFractional PRational where
...
-  plam $ \x ->
-    pmatch x $ \(PRational xn xd) ->
-      plet (pgcd # xn # xd) $ \r ->
-        plet (signum xd) $ \s ->
-          pcon $ PRational (s * pdiv # xn # r) (s * pdiv # xd # r)
+  plam $ \x -> unTermCont $ do
+    PRational xn xd' <- tcont $ pmatch x
+    xd <- tcont . plet $ pto xd'
+    r <- tcont . plet $ pgcd # xn # xd
+    s <- tcont . plet . signum $ xd
```
L-as, line 241: ...signum? Not `psignum`?

Plutarch/Rational.hs
```diff
@@ -192,48 +217,36 @@ instance PFractional PRational where
...
 pgcd :: Term s (PInteger :--> PInteger :--> PInteger)
 pgcd = phoistAcyclic $
-  plam $ \x' y' ->
-    plet (abs x') $ \x ->
-      plet (abs y') $ \y ->
-        plet (pmax # x # y) $ \a ->
-          plet (pmin # x # y) $ \b ->
-            pgcd' # a # b
+  plam $ \x' y' -> unTermCont $ do
+    x <- tcont . plet $ abs x'
```
L-as, line 247: `pabs`?

Plutarch/Rational.hs
```diff
@@ -254,37 +267,42 @@ pmax = phoistAcyclic $ plam $ \a b -> pif (a #<= b) b a
 pnumerator :: Term s (PRational :--> PInteger)
 pnumerator = phoistAcyclic $ plam $ \x -> pmatch x $ \(PRational n _) -> n
 
-pdenominator :: Term s (PRational :--> PInteger)
+pdenominator :: Term s (PRational :--> PPositive)
 pdenominator = phoistAcyclic $ plam $ \x -> pmatch x $ \(PRational _ d) -> d
 
 pfromInteger :: Term s (PInteger :--> PRational)
-pfromInteger = phoistAcyclic $ plam $ \n -> pcon $ PRational n 1
+pfromInteger = phoistAcyclic $ plam $ \n -> pcon $ PRational n $ punsafeDowncast 1
```
L-as, line 274: Why do you have `punsafeDowncast`?

L-as: Literals for `PPositive` should work fine.

L-as (changes requested):
We mostly just need TODOs, great work, thanks!

Plutarch/Rational.hs
TotallyNotChase: [it should be](https://hackage.haskell.org/package/base-4.16.2.0/docs/src/GHC.Real.html#%25) - but we can have further haskell level logic here to invert the sign of the numerator if denominator is negative. Probably unnecessary though.

Plutarch/Rational.hs
TotallyNotChase: modified

Plutarch/Rational.hs
TotallyNotChase: updated - are we preferring the Plutarch typeclass functions in general?

Plutarch/Rational.hs
TotallyNotChase: added

Plutarch/Rational.hs
TotallyNotChase: I optimized it

Plutarch/Rational.hs
TotallyNotChase: oh yes certainly, oversight - added back.

Plutarch/Rational.hs
L-as: Smaller code

Plutarch/Rational.hs
L-as: Good question.

L-as (approved):
LGTM

PR #444, opened 2022-04-19
materialize project

brainrake:
Use [https://input-output-hk.github.io/haskell.nix/tutorials/materialization.html](materialization) to speed up build and avoid IFD.
- need to check if materialized files are the same on other systems. `darwin` anyone?


flake.nix
```diff
@@ -455,8 +455,15 @@
...
           else source; in
+        let flavor = "${ghcName}-${system}"; in
         let pkgSet = (nixpkgsFor system).haskell-nix.cabalProject' ({
           src = ./.;
+          index-state = "2022-02-15T00:00:00Z";
+          plan-sha256 = {
+            "ghc921-x86_64-linux" = "1qx40r75wy3b3aqx0jx1ajgqhz5vcksnki31gn55rzbhvv03jk4n";
+            "ghc8107-x86_64-linux" = "0xc9wf1v7k8rvvb3a728idnsph9ryl2y57kxqh7gz22lwd5aj8ga";
+          }.${flavor};
+          materialized = ./. + "/.materialized-${flavor}";
```
L-as, line 466: Check materialised and command to update materialised files automagically? @brainrake

brainrake: It checks materialized when building the package (not devShell iirc). However, updateMaterialized script is a bit complicated, the suggestion in the trace tries to update the wrong path (inside /nix/store because of flake). I'll look into making it more accessible.

L-as:
https://github.com/input-output-hk/plutus/pull/4548 seems like it might not be worth it after all lol

L-as:
OTOH we still want to avoid IFD if possible, I wonder whether there's an in-between solution that doesn't involve checking in huge amounts of code.

brainrake:
this works now, at least on linux. need to check wether the same files are materialized on eg. darwin

L-as:
So, I'm thinking what we do is we don't check in the code, don't use a FOD, use a standard derivation (possibly with ca-references sprinkled in if it helps), and build the plan *only* on `x86_64-linux`. The `x86_64-darwin` etc. attributes will also import a `x86_64-linux` derivation to figure out the attributes.
This will mean that `nix flake check` will work flawlessly on `x86_64-linux`, and also on other platforms *if* the plan has already been generated by CI.
Considering we don't really support other platforms, this is an OK compromise for now.
As I've said to others before, the solution is to support WASI derivations in Nix, however, this would require a lot of work, and can be done in the future.

L-as:
Honestly we should just support only `x86_64-linux`.

brainrake:
could we define a platform-independent system for the IFDs?

L-as:
That's WASI

PR #449, opened 2022-04-19
Fix PPrelude package and add PRelude

MangoIV:
- I made a mistake when adding `PPRelude` that made it impossible to actually import it into a project and fixed this in this PR
- I added Relude which now seems to be compatible with ghc921

- I will test whether this works in another project this time, should've done this last time as well, I guess


MangoIV:
seems to work as expected.

srid:
@MangoIV I seem to be out of the loop here, but what purpose does `plutarch-preludes` serve?

I approve the use of `relude` as a prelude, but can't you use that directly in the smart contract, rather than creating a proxy package for it?

MangoIV:
It's supposed to be used with cabal mixins to avoid having to import the Plutarch Prelude without explicit import list. 
There's no overlap in the libraries between Prelude/Relude and Plutarch Prelude so it works out nicely. Like this you can avoid having to switch off -Wmissing-import-lists

srid:
> It's supposed to be used with cabal mixins to avoid having to import the Plutarch Prelude without explicit import list.

Okay. Could you create `plutarch-preludes/README.md` introducing its purpose, as well as instructing how this package should be used (ie., the `.cabal` changes necessary in downstream project for including the mixin and anything else relevant)?

L-as:
I'm not actually sure we need this, because this will likely be specific to each project. Seems like to should be part of the downstream project.

srid:
Yea, I do have a feeling it is better to "inline" these mixins directly in the downstream cabal. We can document how to do this (eg: use _both_ Plutarch's prelude and Relude's prelude) in Plutarch guide.

MangoIV:
> "inline" these mixins directly in the downstream cabal

I don't know about this. I feel like it's quite awkward to create an own library just to have a local prelude that contains the mixed prelude that is contained in the same package so you can fix warnings.

L-as:
well we've done it in our own private MLabs plutus-scaffold

MangoIV:
I have used it in a project and like this it's quite convenient, you just import the library and you can 

```haskell
  mixins:
    base hiding (Prelude),
    plutarch-preludes (PRelude as Prelude)
  ``` (or `PPrelude as Prelude` respectively)

MangoIV:
> well we've done it in our own private MLabs plutus-scaffold
that's actually the reason why I think we should have it in `plutarch-extra`, it just seemed awkward to me. But yeah, your decision.

MangoIV:
Another reason this makes sense is that it prevents ambiguities for projects with only one library (cabal will always ask you to specify whether you want to load the prelude package or the other package which is annoying, you could just always load the one that isn't the prelude package)

samuelWilliams99:
Any update here?

L-as:
I don't think it belongs in Plutarch honestly at all. We should just remove the prelude package.

MangoIV:
Why not? It's also just part of `plutus-extra`. This adds a whole lot of convenience,　in actual `plutarch` projects... What is the issue with it?

MangoIV:
Bumping this; I still think this is pretty useful for repos using plutarch as it saves a substantial amount of boilerplate and I don't know why we shouldn't include it,really :)

L-as:
This isn't a priority for 1.2

MangoIV:
relude doesn't build anymore with our current package set anyway :( 
I guess I'll ramp this up again as soon as we have other issues resolved. Feel free to close this for now :)

PR #456, opened 2022-04-23
Golden configuration to choose which goldens are tracked and golden storage path

### LINKED ISSUE #455: Let users choose which golden tests are generated in plutarch-test
I'm using `plutarch-test` in a project where I don't care about tracking the pre eval UPLC, I only care about the benchmark and the resulting UPLC. I think it'd be nice if we could have a config to choose which goldens are actually generated.

The change would be quite minimal I'd assume, something like-
```hs
data GoldenConf = GoldenConf
  { trackPreEval :: Bool
  -- ^ Whether to track UPLC golden.
  , trackPostEval :: Bool
  -- ^ Whether to track evaluated UPLC golden.
  , trackBench :: Bool
  -- ^ Whether to track benchmark golden.
  }

-- | Like 'pgoldenSpec' but takes a 'GoldenConf' to determine which goldens to track.
pgoldenSpec' :: HasCallStack => GoldenConf -> PlutarchGoldens -> Spec
pgoldenSpec' (GoldenConf {trackPreEval, trackPostEval, trackBench}) map = do
  base <- currentGoldenKey
  let bs = runListSyntax map
      goldenKeys = [(trackPreEval, UPLCPreEval), (trackPostEval, UPLCPostEval), (trackBench, Bench)]
      goldenTests = [ t | (True, t) <- goldenKeys ]
  -- Golden tests
  describe "golden" $ do
    goldenTestSpec base bs `mapM_` goldenTests
  -- Assertion tests (if any)
  let asserts = flip mapMaybe bs $ \(k, v) -> do
        (k,) . (\f -> f (goldenValueEvaluated v) $ goldenValueBenchVal v) <$> goldenValueExpectation v
  unless (null asserts) $ do
    forM_ asserts $ \(k, v) ->
      it (goldenKeyString $ "<golden>" <> k <> "assert") v

```
and `pgoldenSpec` would turn into `pgoldenSpec' $ GoldenConf True True True`

cc @srid

TotallyNotChase:
Resolves #455

### DIFF
--- a/plutarch-test/common/Plutarch/Test/Golden.hs
+++ b/plutarch-test/common/Plutarch/Test/Golden.hs
@@ -3,6 +3,7 @@
 module Plutarch.Test.Golden (
   -- * DSL
   pgoldenSpec,
+  pgoldenSpec',
   (@|),
   (@\),
   (@->),
@@ -18,6 +19,10 @@ module Plutarch.Test.Golden (
   defaultGoldenBasePath,
   goldenTestPath,
 
+  -- * Golden config
+  GoldenConf (..),
+  goldenTestsFromConf,
+
   -- * Evaluation
   evalScriptAlwaysWithBenchmark,
   compileD,
@@ -25,6 +30,7 @@ module Plutarch.Test.Golden (
 
 import Control.Monad (forM_, unless)
 import qualified Data.Aeson.Text as Aeson
+import Data.Default (Default (def))
 import Data.List.NonEmpty (nonEmpty)
 import Data.Maybe (mapMaybe)
 import Data.Semigroup (sconcat)
@@ -71,7 +77,13 @@ data GoldenConf = GoldenConf
   -- ^ Whether to track evaluated UPLC golden.
   , trackBench :: Bool
   -- ^ Whether to track benchmark golden.
+  , goldenBasePath :: FilePath
+  -- ^ Directory to put the goldens in.
   }
+  deriving stock (Eq, Show)
+
+instance Default GoldenConf where
+  def = GoldenConf True True True defaultGoldenBasePath
 
 {- | Class of types that represent `GoldenValue`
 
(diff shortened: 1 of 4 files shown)


plutarch-test/common/Plutarch/Test/Golden.hs
```diff
@@ -201,6 +229,11 @@ data GoldenTest
     Bench
   deriving stock (Eq, Show, Ord, Enum, Bounded)
 
+goldenTestsFromConf :: GoldenConf -> [GoldenTest]
+goldenTestsFromConf (GoldenConf {trackPreEval, trackPostEval, trackBench}) =
+  let goldenKeys = [(trackPreEval, UPLCPreEval), (trackPostEval, UPLCPostEval), (trackBench, Bench)]
+   in [t | (True, t) <- goldenKeys]
```
L-as, line 235: Is this more readable than `do { (True, t) <- goldenKeys ; pure t }` or `goldenKeys >>= \case { (True, t) -> [t] ; _ -> [] }`?

TotallyNotChase: The list comprehension looks more natural to me but I don't really care either way. Let me know if you want this changed to bind/do.

plutarch-test/common/Plutarch/Test/Golden.hs
```diff
@@ -64,6 +70,21 @@ data GoldenValue = GoldenValue
   -- ^ User test's expectation function
   }
 
+data GoldenConf = GoldenConf
+  { trackPreEval :: Bool
+  -- ^ Whether to track UPLC golden.
+  , trackPostEval :: Bool
+  -- ^ Whether to track evaluated UPLC golden.
+  , trackBench :: Bool
+  -- ^ Whether to track benchmark golden.
```
srid, line 79: Any reason to have these 3 separate fields instead of a single `Set GoldenTest` field?

plutarch-test/common/Plutarch/Test/Run.hs
```diff
@@ -20,11 +27,15 @@ import Test.Hspec.Core.Spec (SpecTree, Tree (Leaf, Node, NodeWithCleanup), runSp
   untracked golden files.
 -}
 noUnusedGoldens :: Spec -> IO ()
-noUnusedGoldens spec = do
+noUnusedGoldens = noUnusedGoldens' def
+
+-- | Like 'noUnusedGoldens' but takes a custom path to the golden storage.
+noUnusedGoldens' :: GoldenConf -> Spec -> IO ()
```
srid, line 33: How does this function behave when the test spec uses `pgoldenSpec'` multiple times but with _different_ config arguments?

srid (commented):
LGTM, aside from the fact that I'm not fully confident that `noUnusedGoldens'` works. Has this been tested anywhere?

plutarch-test/common/Plutarch/Test/Run.hs
TotallyNotChase: It would break. I did test it on another project when the same conf was used and it did work.

Unsure whether there's a solution to the different config issue.

plutarch-test/common/Plutarch/Test/Golden.hs
srid: Yup, I'd say that goes without saying.

Maybe: `data GoldenTest = GoldenTest_Bench | ...`.

srid (approved):
LGTM, with the proviso that both `pgoldenSpec'` and `noUnusedGoldens'` will be deprecated following #458

PR #461, opened 2022-04-27
Pretty printer for Plutarch terms

TotallyNotChase:
Stuff to do
- [x] All unique variable names.
- [x] Use `ST` monad (cleanup)
- [x] Better `punDataSum` detection
- ~~Corpus for good variable name generation~~
- [x] Better constants pretty printing
- [x] if/then/else pretty printing
- ~~Ability to choose names for lambdas~~


Plutarch/Pretty.hs
```diff
@@ -0,0 +1,218 @@
...
+                     in (newName : l, Set.insert newName existingNames, newG)
+                )
+                ([], Set.fromList $ Map.elems nameMap, g)
+                [0 .. depth]
+         in PP.hang indentWidth $
+              PP.sep
+                [ "\\" <> PP.hsep (reverse $ map PP.pretty names) <+> "->"
+                , go False (Map.mapKeys (+ (depth + 1)) nameMap <> Map.fromList (zip [0 .. depth] names)) finalG t
+                ]
+    go _ nameMap g (Apply _ (LamAbs _ _ t) arg) =
```
TotallyNotChase, line 78: I currently only do prettified let bindings for alternating `Apply` + `LamAbs` chains (i.e `Apply (LamAbs (Apply (LamAbs ...)))`). This is the kind of structure produced by `plet`.

Something like `Apply (Apply (Apply (LamAbs ...)))` will not produce let bindings, just a lambda and an application.

I think this conveys the intent better, and makes it easier to identify some optimizations (inlined hoisted funcs etc). But I understand that having let bindings for the other structure as well might make things easier to the eye.

TotallyNotChase:
@L-as do you know where I can find a corpus for "variable names". Essentially just a list of words that are about 1-7 characters long I suppose. 

Good name generation remains the only remaining task. Although many ASTs are identified and given a "smart name" (a mechanism that is intentionally left open to extension in the future), most ASTs are currently impossible to identify and a fresh name is generated. Most names right now are just a jumble of characters, which is a subpar solution.

L-as:
> @L-as do you know where I can find a corpus for "variable names". Essentially just a list of words that are about 1-7 characters long I suppose.
> 
> Good name generation remains the only remaining task. Although many ASTs are identified and given a "smart name" (a mechanism that is intentionally left open to extension in the future), most ASTs are currently impossible to identify and a fresh name is generated. Most names right now are just a jumble of characters, which is a subpar solution.

I don't know of any such thing, but single character and two-character variable names seems fine to me

L-as:
Well it can be done automatically with some magic! I'd say don't worry about it, I'll try doing it later.

TotallyNotChase:
hmm, alright - I'll mark this ready for now. I do think we should have that naming solution soon since it's crucial for leveraging the pretty printer to debug bigger scripts.

A quick and dirty solution could be to just use the annotations.

TotallyNotChase:
Should `printScript`/`printTerm` usages in goldens be replaced with this? Prettification will fundamentally produce much larger outputs, but it'll at least be legible.

L-as:
Let's not do that later.

L-as:
After 1.2

L-as (approved):
LGTM,

PR #467, opened 2022-05-09
Obviate transitive dependency `lint-utils`

srid:
Update the 'emanote' flake input, so that we no longer rely on `lint-utils`. Potentially fixes https://github.com/Liqwid-Labs/plutarch-quickcheck/issues/2

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -190,11 +190,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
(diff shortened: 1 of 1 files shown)


emiflake (approved):
Can confirm it fixes https://github.com/Liqwid-Labs/plutarch-quickcheck/issues/2.

L-as (approved):

PR #471, opened 2022-05-17
Actually run `TryFromSpec` tests; Also amend `PConstantDecl` law

TotallyNotChase:
Turns out, we weren't even running the tests in `TryFromSpec`.....?


L-as (approved):
LGTM

PR #475, opened 2022-05-31
replace haskell.nix fourmolu with nixpkgs fourmolu

MangoIV:
idk if the builds fails due to fourmolu, in case it actually does, i think there's an issue on fourmolu gh iirc

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -100,7 +100,6 @@
         "terminfo"
       ];
 
-      tools.fourmolu = { };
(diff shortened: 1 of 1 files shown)


L-as:
I fixed that in #443

L-as (approved):
LGTM, errors will be fixed later

L-as:
ha nvm this is merging inte my branch

L-as:
will have to do merge locally

PR #477, opened 2022-05-31
Mario/value ada

blamario:


Plutarch/Api/V1/Value.hs
```diff
@@ -192,6 +223,60 @@ pvalueOf = phoistAcyclic $
       # plam (\map -> AssocMap.pfoldAt # token # 0 # plam pfromData # pfromData map)
       # pto value
 
+-- | The 'PCurrencySymbol' of the Ada currency.
+padaSymbol :: Term s PCurrencySymbol
+padaSymbol = phoistAcyclic (foreignImport $ PlutusTx.liftCode Plutus.adaSymbol)
```
L-as, line 228: Why are we using the FFI here..? Seems like you want `pconstant`.

blamario: They generate the same code, but here you are.

L-as (approved):
LGTM

PR #480, opened 2022-06-08
Adjusted the `docs/Run.md` file to work with current api 

MangoIV:
The functions in `docs/Run.md` were broken with the current api. I fixed that and added a utility function that you can use to serialize your script


L-as (approved):
LGTM

PR #493, opened 2022-06-28
Rewrite scott encoding docs

peter-mlabs:
Rewrote and expanded the Scott encoding docs in a manner I find more intuitive. As with previous contributions, this should be checked for both technical accuracy (because this revision is part of my own effort to understand Scott encoding) as well as the usual grammar, spelling, formatting and so forth.

I'd like to augment this with:

- A description and tangible examples of _why_ Scott encoding is more performant that data encoding (I can't recall whether this exists elsewhere in the guide, but if so, I'm not sure where)
- Backlinks to this page in the haddocks of the relevant typeclasses
- A description of how scott and data encoding are derived (as in, the underlying implementation of the generic deriving mechanism), if it's not too complex.

### DIFF
--- a/docs/Concepts/Data and Scott encoding.md
+++ b/docs/Concepts/Data and Scott encoding.md
@@ -1,77 +1,329 @@
 # Data encoding and Scott encoding
 
-In Plutus Core, there are really two (conflicting) ways to represent non-trivial ADTs: [`Constr`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx.html#t:Data) data encoding, or Scott encoding. You should use only one of these representations for your non-trivial types.
+In Plutus Core, there are really two (conflicting) ways to represent non-trivial ADTs: [`Constr`](https://playground.plutus.iohkdev.io/doc/haddock/plutus-tx/html/PlutusTx.html#t:Data) data encoding or Scott encoding. You should use only one of these representations for your non-trivial types.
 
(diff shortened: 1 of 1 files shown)


peter-mlabs:
Some comments from @seungheonoh:

- First four paragraphs are wordy and would be easier to follow if this wasn't the case
- ULPC part could use some elaboration

TotallyNotChase (commented):
I think a lot of this is quite good - but it's rather verbose and more related to scott encoding itself than to Plutarch/UPLC. It'd be nice if the in-depth explanation of scott-encoding was just hyperlinked instead of being put in Plutarch docs. Since, although I do agree that understanding scott encoding is quite convenient for Plutarch - I personally don't think an in-depth explanation belongs in the Plutarch docs. I personally just read the wikipedia article and that has been enough to work with scott encoding in Plutarch so far. Your explanation could also be posted in plutonomicon perhaps and hyper-linked alongside that wikipedia article.

There are still quite a bit of good improvements in this PR that should stay though.

peter-mlabs:
Got it, I think I agree. The issue that I had was that the current "Concepts" page didn't make sense to me, and most of the resources online seemed to assume some knowledge of Church encoding, or were otherwise picking up in the middle of course notes and assuming some background material.

Perhaps once I get around to explaining the deriving mechanisms and why to use Scott encoding vs Data encoding, _that_ material should go in the "concepts" page, and this can be factored out into an appendix for people who need it?

L-as:
Maybe this should go into Plutonomicon or Wikipedia?

peter-mlabs:
I think Plutonomicon would be the place. I've not had the time to revise this, unfortunately.

PR #495, opened 2022-07-03
Vasil compatibility 

MangoIV:
### Checklist
- [x] implement new vasil types
- [x] sensible reexports
- [ ] some more important utility functions 
- [ ] tests for the types

### Needs 
- PR #486


Plutarch/Api/V2/Contexts.hs
```diff
@@ -0,0 +1,78 @@
...
+      ( Term
+          s
+          ( PDataRecord
+              '[ "inputs" ':= PBuiltinList (PAsData PTxInInfo) -- Transaction inputs
+               , "referenceInputs" ':= PBuiltinList (PAsData PTxInInfo)
+               , "outputs" ':= PBuiltinList (PAsData PTxOut) -- Transaction outputs
+               , "fee" ':= V1.PValue 'V1.Sorted 'V1.Positive -- The fee paid by this transaction.
+               , "mint" ':= V1.PValue 'V1.Sorted 'V1.NoGuarantees -- The value minted by the transaction.
+               , "dcert" ':= PBuiltinList (PAsData V1.PDCert) -- Digests of the certificates included in this transaction.
+               , "wdrl" ':= PBuiltinList (PAsData (V1.PTuple V1.PStakingCredential PInteger)) -- Staking withdrawals
```
L-as, line 64: This is wrong: https://plutonomicon.github.io/plutarch-plutus/plutus-ledger-api/html/PlutusLedgerApi-V2-Contexts.html#t:TxInfo

MangoIV: I don't think so. `Map` comes from `PlutusTx.AssocMap` and is defined as `newtype Map k v = Map { unMap :: [(k,v)]}`, it's `ToData` implementation transforms it into a BuiltinList of BuiltinPairs. There's no invariant enforced or anything. another reason I think that is the case is that none of the operations on `Map` require `Ord` on the keys. 
The only difference afaict is that V2 is implicit about the representation whereas V1 isn't.

MangoIV: aha, I found PlutusTx.Bulitins.mkMap, which indeed doesn't seem to use a list of pair but a Map primitive

MangoIV: Now the question is: can I assume sortedness? Looking at the api PlutusTx exposes I guess not. I will defined it as `PMap Unsorted ..`

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -105,12 +105,12 @@ plookup :: (PIsData k, PIsData v) => Term s (k :--> PMap _ k v :--> PMaybe v)
 plookup = phoistAcyclic $
   plam $ \key ->
     plookupDataWith
-      # (phoistAcyclic $ plam $ \pair -> pcon $ PJust $ pfromData $ psndBuiltin # pair)
+      # phoistAcyclic (plam $ \pair -> pcon $ PJust $ pfromData $ psndBuiltin # pair)
```
L-as, line 108: ?

MangoIV: I want apply-redact for the whole repo :D 
Hlint is so mad at us.

L-as: Please do this in another PR.

Plutarch/Api/V1.hs
```diff
@@ -99,6 +99,7 @@ import qualified Plutarch.Api.V1.Value as Value
 
 import Data.Coerce (coerce)
 
+-- note about V2: This should there are no changes in Scripts or V1 itself that affect this module
```
L-as, line 102: ?

L-as (changes requested):
AFAICT `compile`, FFI, etc. haven't been fixed.

PR #496, opened 2022-07-03
Make PlutusType deriving more ergonomic

MangoIV:
### Checklist
- [x] change the underlying classes and type families
- [ ] fix ambiguity errors

### Needs
- PR #486


Plutarch.hs
```diff
@@ -22,7 +22,7 @@ module Plutarch (
   PP.PMatch,
   PO.printTerm,
   PO.printScript,
-  (PL.#$),
+  -- (PL.#$), FIXME? wtf?
```
L-as, line 25: elaborate?

Plutarch/Bool.hs
```diff
@@ -126,19 +126,19 @@ x #|| y = pforce $ por # x # pdelay y
...
 
 -- | Hoisted, Plutarch level, lazily evaluated boolean or function.
 por :: Term s (PBool :--> PDelayed PBool :--> PDelayed PBool)
-por = phoistAcyclic $ plam $ \x -> pif' # x # (phoistAcyclic $ pdelay $ pcon PTrue)
+por = phoistAcyclic $ plam $ \x -> pif' # x # phoistAcyclic (pdelay $ pcon PTrue)
 
 -- | Hoisted, Plutarch level, strictly evaluated boolean or function.
 por' :: Term s (PBool :--> PBool :--> PBool)
-por' = phoistAcyclic $ plam $ \x -> pif' # x # (pcon PTrue)
+por' = phoistAcyclic $ plam $ \x -> pif' # x # pcon PTrue
```
L-as, line 141: ?

Plutarch/Internal/PlutusType.hs
```diff
@@ -1,7 +1,8 @@
-{-# LANGUAGE AllowAmbiguousTypes #-}
```
L-as, line 1: I assume you'll undo this

Plutarch.hs
MangoIV: It somehow showed the error to be on that line which was just very weird as it's an ambiguity error that affects the whole module :)

Plutarch/Internal/PlutusType.hs
MangoIV: If I don't need it after my changes that'd be neat I guess. I'll see

L-as:
This work needs to be redone after we move to 2.0. The correct fix here is to fix GHC.

PR #499, opened 2022-07-04
stop building ghc810

MangoIV:
as we remove ghc8107 support I removed it from the checks, the apps and the devShells

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -552,6 +552,7 @@
       devShells = perSystem (system:
         {
           "ghc9" = self.flake.${system}.devShell;
+          "ghc810" = self.ghc810Flake.${system}.devShell;
         });
 
       devShell = perSystem (system: self.devShells.${system}."ghc9");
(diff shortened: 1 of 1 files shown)


MangoIV:
I have not removed all the code to make it easier to revive it if we actually have to repair the support at some point, I hope that makes sense, @L-as ?

L-as:
It's blocking CI

---元のメッセージ---
　 2022年7月4日 14:11, Mango The Fourthに が送信

flake.nix
```diff
@@ -555,8 +551,8 @@
 
       devShells = perSystem (system:
         {
-          "ghc810" = self.ghc810Flake.${system}.devShell;
           "ghc9" = self.flake.${system}.devShell;
+          "ghc810" = self.ghc810Flake.${system}.devShell;
```
L-as, line 555: This should be removed

L-as (changes requested):
Otherwise it will be built by CI

flake.nix
MangoIV: I just thought it doesn't really matter as there's nothing in there really that has to be built extra. (Hls comes from nixpkgs and is reliably cached) but if you wish I'll remove it. ;)

L-as: A whole extra GHC potentially, and all the libraries we depend on. That's a lot.

PR #500, opened 2022-07-06
Add more docs on PMatch

przembot:
Adding additional explanation on how `PMatch` works, with listed prerequisites to have a better understanding of how it works internally.


L-as:
We only have PlutusType now

L-as:
Sorry for forgetting to respond to this, but the changes are all in master now I think.

PR #509, opened 2022-07-07
Add pre-commit-hooks to prepare for  hlint and clean up format checking

MangoIV:
- removed hacky formatchecking 
- prepared for running hlint check as soon as we have all the hints applied

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,44 +1,72 @@
 {
   description = "plutarch";
 
-  inputs.haskell-nix.url = "github:input-output-hk/haskell.nix";
-  inputs.nixpkgs.follows = "haskell-nix/nixpkgs-unstable";
-  inputs.nixpkgs-latest.url = "github:NixOS/nixpkgs?rev=a0a69be4b5ee63f1b5e75887a406e9194012b492";
+  inputs = {
+    haskell-nix.url = "github:input-output-hk/haskell.nix";
+    nixpkgs.follows = "haskell-nix/nixpkgs-unstable";
+    nixpkgs-latest.url = "github:NixOS/nixpkgs?rev=a0a69be4b5ee63f1b5e75887a406e9194012b492";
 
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -1,44 +1,72 @@
...
-  inputs.protolude.flake = false;
-
-  # 0.4.5 hasn't been published to Hackage...
-  inputs.flat.url = "github:Quid2/flat";
-  inputs.flat.flake = false;
-
-  inputs.emanote.url = "github:srid/emanote/master";
-
-  outputs = inputs@{ self, nixpkgs, nixpkgs-latest, iohk-nix, haskell-nix, hercules-ci-effects, haskell-nix-extra-hackage, ... }:
+  inputs = {
```
L-as, line 4: ?

plutarch-test/Main.hs
```diff
@@ -38,6 +38,7 @@ spec :: Spec
 spec = do
   BaseSpec.spec
   ExtraSpec.spec
+
```
L-as, line 41: Please don't include irrelevant changes

flake.nix
MangoIV: especially as it gets more. it's also less to type.

L-as: You have more than double the lines in an already long file.

L-as:
How do you format the code?

L-as:
How is this better than what we have before?

MangoIV:
- It's more readable
- easier to use
- will make adding more checks absolutely easy 
- most importantly: runs automatically so you never have an unformatted commit

L-as:
I've decided against this unfortunately. While the idea is nice, pre-commit is pretty much just bloatware.
The only useful feature is that it automatically adds the git commit hook when you enter a shell. That's trivial to reimplement. It is not at all worth the complexity of this *huge* framework. It would take ~20 lines of bash to reimplement.

PR #510, opened 2022-07-07
Add git hook to shell to check formatting before commit

L-as:
@MangoIV I think this does everything you want?

### DIFF
--- a/bin/check
+++ b/bin/check
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -xe
+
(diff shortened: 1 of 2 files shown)


MangoIV:
I just thought that one cleaned up proven to be working, nixified thing that works for both formatting and shell hooks both locally and in the ci was nice. it also had a really neat ui and integrated well with direnv. 
Other tools were easy to add. (e.g. hlint and shellcheck will need more bashscripting, why do this if there's a done thing?)
I don't get the sentiment of reducing "bloat" for the sake of it.

L-as:
It's extremely complex and offers no real benefit. I don't think ignoring files that haven't changed is a great feature, because just checking that can often be more expensive than just doing it. In our case, one of our checking tools is horribly slow, can't remember which, but just optimising that would be more fruitful.

What do you mean it doesn't work? This sets the pre-commit hook if it hasn't been set already.

L-as:
Done in central flake

PR #512, opened 2022-07-08
Derive Eq for sorted maps via data equality

### LINKED ISSUE #491: Make `PMap` equality use data equality

MangoIV:
- closes #491 
- forgot to make clean branch and hence need #511


Plutarch/Api/V1/AssocMap.hs
```diff
@@ -63,14 +64,24 @@ import Plutarch.Unsafe (punsafeCoerce, punsafeDowncast)
...
 data KeyGuarantees = Sorted | Unsorted
 
 type role PMap nominal nominal nominal nominal
 newtype PMap (keysort :: KeyGuarantees) (k :: PType) (v :: PType) (s :: S) = PMap (Term s (PBuiltinMap k v))
   deriving stock (Generic)
-  deriving anyclass (PlutusType, PIsData, PEq, PShow)
+  deriving anyclass (PlutusType, PIsData, PShow)
 instance DerivePlutusType (PMap keysort k v) where type DPTStrat _ = PlutusTypeNewtype
 
+instance PEq (PMap 'Unsorted k v)
```
L-as, line 77: I don't think this is correct. This instance should be removed, no?

MangoIV: is that not how it was derived before for unsorted map? this is the same as anyclass deriving

L-as: yep, and I'm pretty sure this is incorrect

MangoIV: What do you mean? Like in you don't want two maps that are equal modulo sortedness to be seen as unequal? I think that's fine tbh.

MangoIV: So you just want me to remove it? that means we don't have any notion of equality for Unsorted Maps?

L-as: Yes, it's confusing.

L-as: You can always just pto then compare.

MangoIV: okay then, lemme remove it

L-as (approved):
Great!

PR #515, opened 2022-07-09
Fix hrecField and pfield type variable order and deprecate hrecField 

### LINKED ISSUE #397: Change order of type variables in `hrecField` and `pfield`
### Problem

Using `hrecField` and `pfield` would be much nicer with the output polymorphism if we could do something like:
```haskell
pfield @"bla" @PSomething 
-- or 
pfield @"bla" @(PAsData PSomething) 
-- instead of 
pfield @"bla" @_ @_ @_ @_ @_ @PSomething
```
Same goes for hrecField. 

### proposed solution

In particular an ordering like `label`, `origin type`, `result type` would be good. 

### relevant issue 

#275

MangoIV:
- closes #397


L-as (approved):
LGTM

PR #438, opened 2022-04-15
Upstream a couple of PTryFrom instances 

MangoIV:
## Instances

- [x] PMaybeData
- [x] PCredential
- [x] PStakingCredential
- [x] PPubKeyHash
- [x] PValidatorHash
- [x] PTokenName 
- [x] PCurrencySymbol
- [x] PMap 
- [x] PValue
- [x] PPOSIXTime

## Tests

- [x] Succeeding property tests
- [x] Test `Extra.ByteString` utility functions

PR #520, opened 2022-07-13
Update `PTryFrom` instances for Plutarch 1.2

danielfarrelly:
Updates based on #438


Plutarch/Api/V1/Tx.hs
```diff
@@ -38,6 +43,24 @@ instance DerivePlutusType PTxId where type DPTStrat _ = PlutusTypeData
...
+    opq' <- tcont . plet $ pasConstr # opq
+    dataBs <- tcont $ \f ->
+      pif
+        (pfstBuiltin # opq' #== 0 #&& plength # (psndBuiltin # opq') #== 1)
+        (f $ phead #$ psndBuiltin # opq')
+        (ptraceError "ptryFrom(TxId): bad constructor")
+    unwrapped <- tcont . plet $ ptryFrom @(PAsData PByteString) dataBs snd
+    tcont $ \f ->
+      pif (plengthBS # unwrapped #== 28) (f ()) (ptraceError "ptryFrom(TxId): must be 28 bytes long")
+    pure (punsafeCoerce opq, pcon . PTxId $ pdcons # pdata unwrapped # pdnil)
```
TotallyNotChase, line 62: @L-as Is this the most ergonomic way to implement `PTryFrom` for data sum stuff currently? Surely users shouldn't be expected to perform structure validation when the target invariant is only for one of the fields within - I assume there is a better solution utilizing `PTryFromExcess`?

Plutarch/Api/V1/Tx.hs
```diff
@@ -38,6 +43,24 @@ instance DerivePlutusType PTxId where type DPTStrat _ = PlutusTypeData
 instance PUnsafeLiftDecl PTxId where type PLifted PTxId = Plutus.TxId
 deriving via (DerivePConstantViaData Plutus.TxId PTxId) instance PConstantDecl Plutus.TxId
 
+instance PTryFrom PData PTxId where
+  type PTryFromExcess PData PTxId = Flip Term PTxId
+  ptryFrom' opq cont = ptryFrom @(PAsData PTxId) opq (cont . first punsafeCoerce)
+
+instance PTryFrom PData (PAsData PTxId) where
+  type PTryFromExcess PData (PAsData PTxId) = Flip Term PTxId
```
L-as, line 51: There's no point in having this excess, since it's the same as the output.

Plutarch/Api/V1/Tx.hs
```diff
@@ -38,6 +43,24 @@ instance DerivePlutusType PTxId where type DPTStrat _ = PlutusTypeData
...
+  ptryFrom' opq = runTermCont $ do
+    opq' <- tcont . plet $ pasConstr # opq
+    dataBs <- tcont $ \f ->
+      pif
+        (pfstBuiltin # opq' #== 0 #&& plength # (psndBuiltin # opq') #== 1)
```
L-as, line 56: This is inefficient. You're traversing the list twice.

L-as: Optimally you should be able to first verify it's a `PDataSum`, but because it doesn't return any excess, it's not efficient.

Plutarch/Builtin.hs
```diff
@@ -3,6 +3,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Plutarch.Builtin (
+  Flip,
```
L-as, line 6: .…This should not be public API.

L-as (changes requested):
Only a bit remaining!

Plutarch/Api/V1/Tx.hs
TotallyNotChase: This seems like a big drawback, why doesn't `PDataSum` return an excess with fields?

Plutarch/Api/V1/Tx.hs
L-as: From my other comment:
>  Optimally you should be able to first verify it's a PDataSum, but because it doesn't return any excess, it's not efficient.

Plutarch/Api/V1/Tx.hs
L-as: (Human) laziness. I'm sure there's a way to do it, but I'd rather rethink the whole excess field bullshit.

Plutarch/Builtin.hs
MangoIV: maybe we should add it to `Extra.TryFrom`, otherwise everybody declares their own Flip? Maybe it's not an issue though.

TotallyNotChase: it really does feel weird for everyone to have to define it inline - even if it's supposed to be a "type lambda".

+1 to having it exposed in `plutarch-extra` at this point

L-as:
something something name not much shorter than definition.

Rather than doing
```haskell
import Something (Flip)
```

You just do
```haskell
newtypo Flip f b a = Flip (f a b) deriving stock Generic
```
It is barely any shorter to do the derive, and it's barely code reuse. Inlining is good to help people understand the code.

PR #521, opened 2022-07-14
Drop `PAsData` from TxInfo where it is redundant

emiflake:
Summary:
- For both V1 and V2 `PTxInfo`, I dropped unnecessary `PAsData`s, since `PSubtype PData a` for those types.
- Replaced all instances of `PIsData (PBuiltinList a)` with one that uses `PSubtype`.
- Added `plistData`, a wrapper for the `ListData` builtin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 # Unreleased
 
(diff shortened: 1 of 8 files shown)


L-as (approved):
LGTM, thanks!

PR #522, opened 2022-07-17
fix broken GUIDE.md links

gelisam:
`GUIDE.md` has been renamed to `README.md`, but several links were still pointing to the old name.


TotallyNotChase (approved):
Thank you!

Issue #490, opened 2022-06-28
Do we have the same issue as https://github.com/input-output-hk/plutus/pull/4710 ?


L-as:
https://github.com/input-output-hk/plutus/pull/4710

TotallyNotChase:
cc @blamario

L-as:
We don't

PR #524, opened 2022-07-19
`PValue` ordering

TotallyNotChase:
`POrd` instance for `PValue` - with a somewhat optimized implementation. 

Also some type signature fixes

Can't add tests right now - might try to later. But it's not like other value stuff have tests either...

### DIFF
--- a/Plutarch/Api/V1/AssocMap.hs
+++ b/Plutarch/Api/V1/AssocMap.hs
@@ -39,6 +39,9 @@ module Plutarch.Api.V1.AssocMap (
   pdifference,
   punionWith,
   punionWithData,
+
+  -- * Partial order operations
+  pcheckBinRel,
 ) where
 
 import qualified PlutusLedgerApi.V1 as Plutus
@@ -60,6 +63,7 @@ import Plutarch.Lift (
  )
 import qualified Plutarch.List as List
 import Plutarch.Prelude hiding (pall, pany, pfilter, pmap, pnull, psingleton)
+import qualified Plutarch.Prelude as PPrelude
 import Plutarch.Show (PShow)
 import Plutarch.Unsafe (punsafeCoerce, punsafeDowncast)
 import qualified PlutusCore as PLC
@@ -454,3 +458,55 @@ pmapData = phoistAcyclic $
         )
         (const pnil)
         # pto map
+
+{- | Given a comparison function and a "zero" value, check whether a binary relation holds over
+2 sorted 'PMap's.
+
+This is primarily intended to be used with 'PValue'.
+-}
+pcheckBinRel ::
+  forall k v s.
+  (POrd k, PIsData k, PIsData v) =>
+  Term
+    s
+    ( (v :--> v :--> PBool)
+        :--> v
+        :--> PMap 'Sorted k v
+        :--> PMap 'Sorted k v
+        :--> PBool
+    )
+pcheckBinRel = phoistAcyclic $
+  plam $ \f z m1 m2 ->
+    let inner = pfix #$ plam $ \self l1 l2 ->
+          pelimList
+            ( \x xs ->
+                plet (pfromData $ psndBuiltin # x) $ \v1 ->
+                  pelimList
+                    ( \y ys -> unTermCont $ do
+                        v2 <- tcont . plet . pfromData $ psndBuiltin # y
+                        k1 <- tcont . plet $ pfromData $ pfstBuiltin # x
+                        k2 <- tcont . plet $ pfromData $ pfstBuiltin # y
+                        pure $
+                          pif
+                            (k1 #== k2)
+                            ( f # v1 # v2 #&& self
+                                # xs
+                                # ys
+                            )
+                            $ pif
+                              (k1 #< k2)
+                              (f # v1 # z #&& self # xs # l2)
+                              $ f # z # v2 #&& self
+                                # l1
+                                # ys
+                    )
+                    ( f # v1 # z
+                        #&& PPrelude.pall
+                          # plam (\p -> f # pfromData (psndBuiltin # p) # z)
+                          # xs
+                    )
+                    l2
+            )
+            (PPrelude.pall # plam (\p -> f # z #$ pfromData $ psndBuiltin # p) # l2)
+            l1
+     in inner # pto m1 # pto m2
--- a/Plutarch/Api/V1/AssocMap.hs
+++ b/Plutarch/Api/V1/AssocMap.hs
@@ -67,6 +67,7 @@ import qualified PlutusCore as PLC
 import Prelude hiding (all, any, filter, lookup, null)
 
 import Data.Proxy (Proxy (Proxy))
+import Data.Traversable (for)
 
 data KeyGuarantees = Sorted | Unsorted
 
@@ -108,17 +109,17 @@ instance
   type PConstanted (PlutusMap.Map k v) = PMap 'Unsorted (PConstanted k) (PConstanted v)
   pconstantToRepr m = (\(x, y) -> (Plutus.toData x, Plutus.toData y)) <$> PlutusMap.toList m
   pconstantFromRepr m = fmap PlutusMap.fromList $
-    flip traverse m $ \(x, y) -> do
+    for m $ \(x, y) -> do
       x' <- Plutus.fromData x
       y' <- Plutus.fromData y
       Just (x', y')
 
 -- | Tests whether the map is empty.
-pnull :: Term s (PMap _ k v :--> PBool)
+pnull :: Term s (PMap any k v :--> PBool)
 pnull = plam (\map -> List.pnull # pto map)
 
 -- | Look up the given key in a 'PMap'.
-plookup :: (PIsData k, PIsData v) => Term s (k :--> PMap _ k v :--> PMaybe v)
+plookup :: (PIsData k, PIsData v) => Term s (k :--> PMap any k v :--> PMaybe v)
 plookup = phoistAcyclic $
   plam $ \key ->
     plookupDataWith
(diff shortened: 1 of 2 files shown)


TotallyNotChase:
Fourmolu breaks with 
```
  AST of input and AST of formatted code differ.
    at Plutarch/Api/V1/Value.hs:127:12-47
    at Plutarch/Api/V1/Value.hs:130:12-47
  Please, consider reporting the bug.
```

Plutarch/Api/V1/Value.hs
```diff
@@ -116,6 +119,26 @@ instance PEq (PValue 'Sorted 'Positive) where
...
+
+Use 'pcheckBinRel' if 'AmountGuarantees' is 'NoGuarantees'.
+-}
+instance POrd (PValue 'Sorted 'NonZero) where
+  a #< b = f # a # b
+    where
+      f = phoistAcyclic $ pcheckBinRel #$ phoistAcyclic $ plam (\x y -> pfromData x #< pfromData y)
+  a #<= b = f # a # b
+    where
+      f = phoistAcyclic $ pcheckBinRel #$ phoistAcyclic $ plam (\x y -> pfromData x #<= pfromData y)
```
TotallyNotChase, line 140: I added `POrd` instances but I don't know if these should be standalone `ple` and `pleq` functions instead.

Plutarch/Api/V1/Value.hs
```diff
@@ -161,6 +184,105 @@ instance
...
+                  ( pinnerCheckBinRel # f # pfromData (psndBuiltin # x) # AssocMap.pempty
+                      #&& pall
+                        # plam
+                          ( \p ->
+                              let v1dat = psndBuiltin # p
+                               in pinnerCheckBinRel # f # pfromData v1dat # AssocMap.pempty
+                          )
+                        # xs
+                  )
+                  l2
```
TotallyNotChase, line 230: this is optimized for cpu/mem but not script size - unionWith also uses this whole key ordering logic deep within. I wonder if there's a way to manually do common subexpr elim here for optimality.

Plutarch/Api/V1/Value.hs
```diff
@@ -286,14 +408,14 @@ plovelaceValueOf = phoistAcyclic $
 punionWith ::
   Term
     s
-    ( (PInteger :--> PInteger :--> PInteger) :--> PValue 'Sorted _ :--> PValue 'Sorted _
+    ( (PInteger :--> PInteger :--> PInteger) :--> PValue 'Sorted any0 :--> PValue 'Sorted any1
```
TotallyNotChase, line 411: I allow different `AmountGuarantees` in the binary `PValue` functions in a lot of places since technically it should be valid.

Plutarch/Api/V1/Value.hs
```diff
@@ -161,6 +184,105 @@ instance
...
+    dat0 :: ClosedTerm (PAsData PInteger)
+    dat0 = pconstantData 0
+    pinnerCheckBinRel ::
+      ClosedTerm
+        ( (PAsData PInteger :--> PAsData PInteger :--> PBool)
+            :--> ( PMap 'Sorted PTokenName PInteger
+                    :--> (PMap 'Sorted PTokenName PInteger :--> PBool)
+                 )
+        )
+    pinnerCheckBinRel = phoistAcyclic $
```
L-as, line 252: There is a lot of common logic shared between this and the main one. I think you can formulate this as something that works on `PMap a b` where `POrd a`, given some "default value".

Plutarch/Api/V1/Value.hs
```diff
@@ -334,28 +456,30 @@ pnormalize = phoistAcyclic $
       pif (intData #== zeroData) (pcon PNothing) (pcon $ PJust intData)
 
 -- | Assert the value is properly sorted and normalized.
-passertSorted :: Term s (PValue _ _ :--> PValue 'Sorted 'NonZero)
+passertSorted :: forall anyKey anyAmount s. Term s (PValue anyKey anyAmount :--> PValue 'Sorted 'NonZero)
```
L-as, line 459: Why `forall`?

Plutarch/Api/V1/Value.hs
L-as: It's better this way.

L-as (changes requested):
Thanks!

It seems like we're in fact not vulnerable to #490, because their issue is that they don't represent innullity in the types.

Plutarch/Api/V1/Value.hs
```diff
@@ -161,6 +184,105 @@ instance
   where
   inv a = punsafeCoerce $ PlutusTx.inv (punsafeCoerce a :: Term s (PValue 'Sorted 'NoGuarantees))
 
+-- | Check whether a binary relation holds over 2 sorted 'PValue's.
+pcheckBinRel ::
```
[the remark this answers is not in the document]
L-as: See comment below about refactoring to operate on `PMap`.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -108,35 +109,34 @@ instance
...
-plookup :: (PIsData k, PIsData v) => Term s (k :--> PMap _ k v :--> PMaybe v)
+plookup :: (PIsData k, PIsData v) => Term s (k :--> PMap any k v :--> PMaybe v)
 plookup = phoistAcyclic $
   plam $ \key ->
     plookupDataWith
       # phoistAcyclic (plam $ \pair -> pcon $ PJust $ pfromData $ psndBuiltin # pair)
       # pdata key
 
 -- | Look up the given key data in a 'PMap'.
-plookupData :: (PIsData k, PIsData v) => Term s (PAsData k :--> PMap _ k v :--> PMaybe (PAsData v))
```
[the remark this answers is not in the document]
TotallyNotChase: partial type signatures seem to break redundant constraint detection - potential GHC bug

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -378,32 +376,30 @@ punionWithData = phoistAcyclic $
...
-  Term s (PMap g k a :--> PMap _ k b :--> PMap g k a)
+pdifference :: PIsData k => Term s (PMap g k a :--> PMap any k b :--> PMap g k a)
 pdifference = phoistAcyclic $
   plam $ \left right ->
     pcon . PMap $
       precList
         ( \self x xs ->
             plet (self # xs) $ \xs' ->
               pfoldAt
-                # (pfromData $ pfstBuiltin # x)
```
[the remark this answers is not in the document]
TotallyNotChase: all the minor syntax changes was because I got fed up of hlint screaming at me while working on these files

could have a `hlint.yaml` with your preferences in the repo

Plutarch/Api/V1/Value.hs
```diff
@@ -161,6 +184,105 @@ instance
...
+            )
+            ( pall
+                # plam
+                  ( \p ->
+                      let v1dat = psndBuiltin # p
+                       in pinnerCheckBinRel # f # AssocMap.pempty # pfromData v1dat
+                  )
+                # l2
+            )
+            l1
```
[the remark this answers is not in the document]
TotallyNotChase: `pelimList` is usually less noisy and more natural for lists IMO. But `pmatch` might look better in this case if it's utilized alongside `TermCont`

Plutarch/Api/V1/Value.hs
TotallyNotChase: I forgot to remove this lol - was playing around with the type variable ordering

Plutarch/Api/V1/Value.hs
```diff
@@ -161,6 +184,105 @@ instance
...
+     in inner # pto (pto $ pto m1) # pto (pto $ pto m2)
```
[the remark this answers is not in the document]
TotallyNotChase: sure, I updated it to `(pto . pto $ pto m1)` though.

Plutarch/Api/V1/Value.hs
TotallyNotChase: it's intentionally written like this to optimize for CPU/Mem. But now that I think about it, there might be a minimal trade off in CPU/Mem for a decent script size saving here - I implemented that.

Plutarch/Api/V1/Value.hs
```diff
@@ -502,3 +405,11 @@ zeroData = pdata 0
...
 pmapAmounts = phoistAcyclic $
   plam $ \f v -> pcon $ PValue $ AssocMap.pmap # plam (AssocMap.pmap # f #) # pto v
+
+{- | Given an amount comparison function, check whether a binary relation holds over
+2 sorted 'PValue's.
+-}
+pcheckBinRel :: Term s ((PInteger :--> PInteger :--> PBool) :--> PValue 'Sorted any0 :--> PValue 'Sorted any1 :--> PBool)
+pcheckBinRel = phoistAcyclic $
+  plam $ \f v1 v2 ->
+    AssocMap.pcheckBinRel # (AssocMap.pcheckBinRel # f # 0) # AssocMap.pempty # pto v1 # pto v2
```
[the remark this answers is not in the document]
L-as: It probably won't, because the semantics might be different. We don't know whether the inner function would err after 3, 2, or 1 applications. η-reduction only happens when the inner function is a lambda abstraction. That is probably not the case here, so you should η-reduce it manually.

Plutarch/Api/V1/Value.hs
L-as: This reasoning doesn't make sense to me? Why would you save CPU by duplicating the logic?
If you're thinking that you'll somehow have more applications because you use a more abstraction version of the function,
that is possible (albeit I'm not sure it's true), but in that case, you still don't want to duplicate the logic.
You want to instantiate and monomorphise the logic at several types.

L-as:
Maybe you want to η-reduce then we can merge?

Plutarch/Api/V1/Value.hs
```diff
@@ -411,5 +411,9 @@ pmapAmounts = phoistAcyclic $
...
 pcheckBinRel :: Term s ((PInteger :--> PInteger :--> PBool) :--> PValue 'Sorted any0 :--> PValue 'Sorted any1 :--> PBool)
 pcheckBinRel = phoistAcyclic $
-  plam $ \f v1 v2 ->
-    AssocMap.pcheckBinRel # (AssocMap.pcheckBinRel # f # 0) # AssocMap.pempty # pto v1 # pto v2
+  plam $ \f ->
+    subReduction2 $
+      AssocMap.pcheckBinRel # (AssocMap.pcheckBinRel # f # 0) # AssocMap.pempty
+  where
+    subReduction2 :: Term s (PInner a :--> PInner b :--> c) -> Term s (a :--> b :--> c)
+    subReduction2 = punsafeCoerce
```
L-as, line 419: You can actually implement this safely through `pdowncastF`, albeit it will be a bit complicated.
Thanks!

L-as (approved):
LGTM

Plutarch/Api/V1/Value.hs
L-as: That doesn't address my comment.

PR #528, opened 2022-07-22
Remove POrd instance of PValue - add `PPartialOrd`

### LINKED ISSUE #526: Remove POrd for PValue
It doesn't form a total order and can break e.g. `Data.Map` if you have a map of terms of values.
It can possibly also break `PMap`.
cc @TotallyNotChase

TotallyNotChase:
Closes #526

### DIFF
--- a/Plutarch/Bool.hs
+++ b/Plutarch/Bool.hs
@@ -4,8 +4,8 @@
 module Plutarch.Bool (
   PBool (..),
   PEq (..),
-  POrd (..),
   PPartialOrd (..),
+  POrd,
   pif,
   pif',
   pnot,
@@ -79,7 +79,8 @@ class PEq t where
 
 infix 4 #==
 
-class PEq t => POrd t where
+-- | Partial ordering relation.
+class PEq t => PPartialOrd t where
   (#<=) :: Term s t -> Term s t -> Term s PBool
   default (#<=) :: (POrd (PInner t)) => Term s t -> Term s t -> Term s PBool
   x #<= y = pto x #<= pto y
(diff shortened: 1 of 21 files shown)


Plutarch/Bool.hs
```diff
@@ -89,6 +90,18 @@ class PEq t => POrd t where
 infix 4 #<=
 infix 4 #<
 
+-- | Partial ordering relation.
+class PPartialOrd t where
+  pleq :: Term s t -> Term s t -> Term s PBool
```
TotallyNotChase, line 95: should these methods be Plutarch level functions instead?

Plutarch/Bool.hs
```diff
@@ -89,6 +90,18 @@ class PEq t => POrd t where
 infix 4 #<=
 infix 4 #<
 
+-- | Partial ordering relation.
+class PPartialOrd t where
```
TotallyNotChase, line 94: Should this become a superclass of `POrd`? It'd require adding all the instances everywhere else

L-as:
One possibility is to choose the design Rust chose, but when would you depend on a PPartialOrd constraint rather than POrd? We could move #< etc. to PPartialOrd, make a superclass relation, and then derive PPartialOrd for everything that has POrd, but would you ever use the PPartialOrd instance? Is it confusing that you can use < for some things, but only #< for others?
I feel like Rust's approach might be the best solution, but really, we want PartialOrd too.

blamario:
Wouldn't Rust's approach be suboptimal in UPLC, if we adopt `pcompare :: Term s (a #-> b #-> PMaybe POrdering)`? Or do you mean something else about Rust?

I imagine there won't be much user-written polymorphic code; when writing a contract people tend to know what type they're dealing with. It's only the library writers who will have to decide between `PPartialOrd` and `POrd`. Sorted lists and maps will have to insist on `POrd`. I think it's fine to move `#<=` to `PPartialOrd` and to leave `POrd` with nothing but additional laws, and perhaps with the `pcompare` method added.

TotallyNotChase:
I also think moving things into `PPartialOrd` and having just `POrd` with a superclass relation and additional laws is good. I'm against adding `pcompare` though, due to performance concerns.

TotallyNotChase:
Moved the `POrd` methods into `PPartialOrd` and made `POrd` an empty class with a superclass constraint on `PPartialOrd`. 
Suggest the typeclass laws that should be put into the `POrd` haddock

L-as (approved):
LGTM, great!

TotallyNotChase:
@L-as wait we still needed to add laws?

L-as:
@TotallyNotChase Well we can do that in another PR and it's not essential for 1.2 IMO.

PR #533, opened 2022-08-03
Add `PRationalData` to `plutarch-extra`

TotallyNotChase:
`PlutusTx.Ratio.Rational` synonym - since it's often used in datums.

Also some friendly notes on `PRational` to ensure people don't accidentally assume it has the same data encoding as PlutusTx's rational.

P.S: I seem to have accidentally committed an unfinished `PRationalData` file into plutarch-extra and it's just been there sitting as dead code lol (pretend the file is new when reviewing).


L-as:
ote that this will be redundant in 2.0

L-as (approved):

PR #536, opened 2022-08-10
CompiledTerm and verbose plift

SeungheonOh:


SeungheonOh:
Should `plift'` be re written to use `pliftTrace`?

L-as:
Please make `plift'` use this

Plutarch/Lift.hs
```diff
@@ -105,20 +106,27 @@ data LiftError
...
-    (Left e, _, _) -> Left $ LiftError_EvalError e
+          Just h -> (Right h, trace)
+          Nothing -> (Left LiftError_FromRepr, trace)
+        Left e -> (Left $ LiftError_KnownTypeError e, trace)
+    (Left e, _, _) -> (Left $ LiftError_EvalError e, mempty)
+
+{- | Convert a Plutarch term to the associated Haskell value. Fail otherwise.
+This will fully evaluate the arbitrary closed expression, and convert the resulting value.
+-}
+plift' :: forall p. PUnsafeLiftDecl p => Config -> ClosedTerm p -> Either LiftError (PLifted p)
```
L-as, line 128: This should really be deprecated.

Plutarch/Lift.hs
```diff
@@ -105,20 +106,27 @@ data LiftError
...
-{- | Convert a Plutarch term to the associated Haskell value. Fail otherwise.
-This will fully evaluate the arbitrary closed expression, and convert the resulting value.
+{- | Like `plift'` but provides trace from evaluation.
+
+ @since 1.2.1
 -}
-plift' :: forall p. PUnsafeLiftDecl p => Config -> ClosedTerm p -> Either LiftError (PLifted p)
-plift' config prog = case compile config prog of
-  Left msg -> Left $ LiftError_CompilationError msg
+pliftTrace :: forall p. PUnsafeLiftDecl p => Config -> ClosedTerm p -> (Either LiftError (PLifted p), [Text])
```
L-as, line 113: Is there any other information we are not returning? We should maybe also call this something else. `pliftMaybe`? The current name is not that bad though.

L-as (changes requested):
Sorry for another request!

L-as:
Thanks BTW

Plutarch/Lift.hs
SeungheonOh: We can return cek budgets, but I don't think that'd be useful. If traces are the only thing we are returning alongside, `pliftTrace` would be a better name.

SeungheonOh:
I didn't change the name because traces are pretty much only thing we are providing alongside the lifted value.

Plutarch/Lift.hs
L-as: We should return the budget, because someone will come along in the future, and make another function that returns the budget.

Plutarch/Lift.hs
```diff
@@ -105,24 +105,26 @@ data LiftError
...
+pliftTrace config prog = case compile config prog of
+  Left msg -> (Left $ LiftError_CompilationError msg, mempty)
   Right script -> case evalScriptHuge script of
-    (Right (Scripts.unScript -> UPLC.Program _ _ term), _, _) ->
+    (Right (Scripts.unScript -> UPLC.Program _ _ term), _, trace) ->
       case readKnownConstant term of
```
L-as, line 118: I think what we really want is something like this
```haskell
compile'' :: Config -> ClosedTerm a -> CompiledTerm a
evalCompiledTerm :: CompiledTerm a -> (Either EvalError (CompiledTerm a), ExBudget, [Text])
pliftCompiled :: CompiledTerm a -> Either KnownTypeError (Maybe (PLifted a))
```
The question is, what should the name of `compile''` be?

L-as:
FWIW I just realised you can already do this by using `Plutarch.FFI` in combination with `evalScriptHuge`.

Plutarch/Lift.hs
SeungheonOh: I think the idea of `CompiledTerm` is very helpful. It opens the door for some big optimization. For example, it will allow shrinkers to make reasonable alternates and etc.

SeungheonOh:
Is `CompiledTerm a` always isomorphic to `Term s a` using `RCompiled`given that we made `CompiledTerm a` from Plutarch?

L-as:
@SeungheonOh `CompiledTerm a` and `ClosedTerm a` are isomorphic from one perspective yeah (depends on your equality relation). This makes me think, maybe we don't need `CompiledTerm`? Seems like we can just use `ClosedTerm a` everywhere.

SeungheonOh:
I think the notion of compiled term is very helpful. If not we can just make `evalTerm` which runs closed term and put it into `RCompiled`.

SeungheonOh:
We do not need this, instead we can use `evalTerm` from #538

L-as:
how does this differ from plift'?

Issue #423, opened 2022-04-06
Add `PShow` instance for Api types


srid:
It would be good to be able to `pshow` a `PTxOut`, etc. for debugging purposes.

L-as:
Implementing it for PDataSum would be enough

PR #537, opened 2022-08-10
Add pshow instances

tbidne:
Added `PShow` instances to more types. The intention is to add these to `V2`, though most had to also be added to `V1` to prevent orphans.

Example output:
```haskell
-- without parens (pshow)
PTxOut [address = (PAddress [credential = (PPubKeyCredential [_0 = (PPubKeyHash 0x737472)]), stakingCredential = (PDNothing [])]), value = (PValue (PMap [])), datum = (POutputDatumHash [datumHash = (PDatumHash 0x68617368)]), referenceScript = (PDJust [_0 = (PScriptHash 0x7363726970742068617368)])]

-- with parens (pshow' True)
(PTxOut [address = (PAddress [credential = (PPubKeyCredential [_0 = (PPubKeyHash 0x737472)]), stakingCredential = (PDNothing [])]), value = (PValue (PMap [])), datum = (POutputDatumHash [datumHash = (PDatumHash 0x68617368)]), referenceScript = (PDJust [_0 = (PScriptHash 0x7363726970742068617368)])])
```

### DIFF
--- a/Plutarch/DataRepr/Internal.hs
+++ b/Plutarch/DataRepr/Internal.hs
@@ -30,12 +30,15 @@ import Data.Kind (Constraint, Type)
 import Data.List (groupBy, maximumBy, sortOn)
 import Data.Proxy (Proxy (Proxy))
 import Data.SOP.NP (cana_NP)
+import Data.String (fromString)
 import GHC.Generics (Generic)
 import GHC.TypeLits (
   KnownNat,
+  KnownSymbol,
   Nat,
   Symbol,
   natVal,
+  symbolVal,
   type (+),
  )
 import Generics.SOP (
@@ -46,6 +49,7 @@ import Generics.SOP (
   NS (S, Z),
   SListI,
   SOP (SOP),
+  Top,
   case_SList,
   hcollapse,
   hindex,
@@ -58,6 +62,7 @@ import Plutarch (
   POpaque,
   PType,
   PlutusType,
+  PlutusTypeNewtype,
   S,
   Term,
   pcon,
@@ -101,6 +106,7 @@ import Plutarch.DataRepr.Internal.HList (
 import Plutarch.Integer (PInteger)
 import Plutarch.Internal.Generic (PCode, PGeneric, gpfrom, gpto)
 import Plutarch.Internal.PlutusType (
+  DerivePlutusType (DPTStrat),
   DerivedPInner,
   PlutusTypeStrat,
   PlutusTypeStratConstraint,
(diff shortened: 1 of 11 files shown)


SeungheonOh:
This is great. Thank you

L-as:
This is nice, but optimally, we want to fix this in the general case. I.e., we want to derive it.

L-as:
@tbidne would you be willing to take a look at the generic stuff in `Plutarch.Show` and see if you can make it work for Data-encoded ADTs? #423

Plutarch/DataRepr/Internal.hs
```diff
@@ -174,6 +180,52 @@ instance PPartialOrd (PDataRecord '[]) where
...
+  pshow' b _ = withParens b "[]"
+
+instance
+  (All Top xs, KnownSymbol label, PIsData x, PShow x, PShow (PDataRecordShowHelper xs)) =>
+  PShow (PDataRecord ((label ':= x) ': xs)) where
+  pshow' b xs = withParens b $ "[" <> pmatch xs go
+    where
+      go :: PDataRecord ((label ':= x) : xs) s -> Term s PString
+      go (PDCons y ys) =
+        mconcat
```
L-as, line 193: Doesn't `<>` make a lot more sense here?

Plutarch/DataRepr/Internal.hs
```diff
@@ -174,6 +180,52 @@ instance PPartialOrd (PDataRecord '[]) where
...
+--
+-- @
+-- showList' :: Show a => [a] -> String
+-- showList' [] = "[]"
+-- showList' (x:xs) = '[' : show x ++ go xs
+-- where
+--   go [] = "]"
+--   go (y:ys) = ',' : show y ++ go ys
+-- @
+newtype PDataRecordShowHelper as s = PDataRecordShowHelper (Term s (PDataRecord as))
```
L-as, line 210: I'm honestly not a big fan of this helper datatype, there should be cleaner ways to do it, but it's not an issue honestly.
You don't have to change this.

L-as (changes requested):
Thanks! This is great.

L-as:
`./bin/format` BTW

L-as:
Huge thanks!

Wrt. PData and such, you can't derive PShow for PData. If you do, and you run pshow, it will just loop AFAICT, because it contains itself.

So I think that you should derive for POutputDatum, but put a manual instance on PData that says "<data>" or similar. Ideally, we'd have a proper PShow PData, but that can be done another day.

Thanks for this!

Plutarch/DataRepr/Internal.hs
L-as: So normally, I'd do something with sop-core's SList/All to handle this. But it's not worth the effort IMO. It can't be made a lot cleaner.

L-as: E.g. [`cpara_SList`](https://hackage.haskell.org/package/generics-sop-0.5.1.2/docs/Generics-SOP.html#v:cpara_SList)

L-as (approved):
Beautiful!
Waiting for CI.

PR #538, opened 2022-08-11
add `evalTerm`

SeungheonOh:
`evalTerm` evaluates given term and puts back into Plutarch term using `RCompiled`. 

@L-as


SeungheonOh:
It was broken. It's fixed.

Issue #540, opened 2022-08-13
Make `Term` more clear with type tags


SeungheonOh:
I was experimenting with new `evalTerm`. While it's helpful, `Term` itself doesn't carry any information about "what" it's holding. And I think those information should be carried in `Term` itself to be extra clear on if term is evaluated, compiled, or neither. Following is my first attempt in making typesafe wrapper, it's very unpleasant to work with it because it fails to share operators with `Term`. https://github.com/Liqwid-Labs/liqwid-plutarch-extra/pull/58

If we embed the information inside of `Term`, it would allow us to differentiate what term is evaluated or compiled without making messy wrapper around `Term` every time we need testing. Agora had face some serious performance issue in it's benchmarks and tests due to Plutarch `Term`'s lack of clarity regarding evaluated/compiled, and ended up have to externally deal with `Script`. 

Adding this tags will be simple:
```hs
data TermState = Evaluated | NotEvaluated
newtype Term' (e :: TermState) (s :: S) (a :: S -> Type) = Term {asRawTerm :: Word64 -> TermMonad TermResult}
type Term = Term' NotEvaluated
```
and some functions need be updated to currently handle `TermState`. For example, `papp` will have to always return `NotEvaluated` and `plift` can be optimized and not compile when `Evaluated` is given. 

I'm not sure if `TermState` should also have `Compiled`.

PR #541, opened 2022-08-13
Tagging evaluated terms

SeungheonOh:
#540

### DIFF
--- a/Plutarch/Evaluate.hs
+++ b/Plutarch/Evaluate.hs
@@ -11,7 +11,7 @@ module Plutarch.Evaluate (
 import qualified Plutarch.Internal.Evaluate as E
 
 import Data.Text (Text)
-import Plutarch.Internal (ClosedTerm, Config, RawTerm (RCompiled), Term (..), TermResult (TermResult), compile)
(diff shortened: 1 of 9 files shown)


SeungheonOh:
We can make `pconstant` and `pcon` to return evaluated term in right cases, and I tried it. But that would be a more significant rewrites and I'm not sure if that's what we want to do.

L-as:
This is a very big change that is also incompatible with Plutarch 2.0. It's not clear to me why you need this though, what's wrong with having a `EvaluatedTerm` newtype?

SeungheonOh:
new type wrapper requires having a whole use sets of functions rewritten for evaluated terms. Instead, having this tag allows us to use what Plutarch itself offers like papp.

Best,
Seungheon Oh
***@***.***
________________________________

SeungheonOh:
new type wrapper requires having a whole use sets of functions rewritten for evaluated terms. Instead, having this tag allows us to use what Plutarch itself offers like papp.

________________________________

SeungheonOh:
Actually, I getting more skeptical about this because it's largely confusing to use and write codes on it. While it would still be very helpful to use operators from Plutarch itself when using EvaluatedTerm, this is a too much effort just for that.

PR #547, opened 2022-08-18
Add PShow instances to print PValue and PData values

facundominguez:


facundominguez:
Hello! This PShow instances are being useful to debug plutarch programs, I thought of sharing it here in case you want it in.

I'm testing it on a fork of plutarch, so minor adaptations might be necessary to make it work here.

L-as:
Thanks!

Plutarch/Api/V1/Value.hs
```diff
@@ -123,6 +124,11 @@ newtype PValue (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (s :: S)
   deriving anyclass (PlutusType, PIsData, PShow)
 instance DerivePlutusType (PValue keys amounts) where type DPTStrat _ = PlutusTypeNewtype
 
+instance PShow (PValue keys amounts) where
```
L-as, line 127: Why not just derive this?

L-as: Maybe you're unaware of #537 ?

facundominguez: > Maybe you're unaware of #537 ?

I was unaware indeed. I removed the related commits from this PR.

L-as:
Thanks!

PR #549, opened 2022-08-19
expose Plutarch.Test.Property.Gen (Fix #548)

### LINKED ISSUE #548: Expose `Arbitrary` instances from `Plutarch.Test.Property.Gen`
Currently one cannot access the `Arbitrary` instances defined there for plutus-ledger-api types from outside the library.

It would be useful to either expose that module, or one that exports the instances.

Unless this was intentional? I could see a few reasons for that.

Saizan:
As per title.

### DIFF
--- a/plutarch-test/plutarch-test.cabal
+++ b/plutarch-test/plutarch-test.cabal
@@ -108,12 +108,10 @@ library
   hs-source-dirs:  common
   exposed-modules:
     Plutarch.Test
-    Plutarch.Test.Property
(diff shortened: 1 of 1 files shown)


L-as:
Why not just expose everything? Don't you use the other stuff?

Saizan:
I haven't yet, but I guess I might soon. I'll expose everything.

L-as (approved):
LGTM!

PR #555, opened 2022-08-22
change txid length from 28 to 32

Geometer1729:
It seems this 28 should be a 32.

Changing it to 32 locally my code to start working, this included an equality check with a txid from the script context so it seems like the script context must have txids of length 32.


L-as (approved):
Thank you!

PR #560, opened 2022-08-26
Clearer error from `PSubType`

SeungheonOh:
Before:
```
<interactive>:1:1: error:
    • Couldn't match type ‘'False’ with ‘'True’
        arising from a use of ‘pdata’
    • In the expression: pdata a
```

After:
```
<interactive>:1:1: error:
    • Couldn't match type ‘'Unrelated PData PInteger’ with ‘'SuperType’
        arising from a use of ‘pdata’
    • In the expression: pdata a
```

### DIFF
--- a/Plutarch/TryFrom.hs
+++ b/Plutarch/TryFrom.hs
@@ -22,6 +22,8 @@ import Plutarch.Internal.Witness (witness)
 
 import Plutarch.Reducible (Reduce)
 
+import GHC.TypeLits (ErrorMessage (ShowType, Text, (:<>:)), TypeError)
+
 data PSubtypeRelation
   = SuperType
   | Unrelated PType PType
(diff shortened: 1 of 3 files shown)


L-as:
TypeError

SeungheonOh:
It uses TypeError now: 
```
<interactive>:1:1: error:
    • "PData" is not subtype of "PInteger"
    • In the expression: pdata a
```

Plutarch/TryFrom.hs
```diff
@@ -21,12 +22,18 @@ import Plutarch.Internal.Witness (witness)
 
 import Plutarch.Reducible (Reduce)
 
-type family Helper (a :: PType) (b :: PType) (bi :: PType) :: Bool where
-  Helper _ b b = 'False
+import GHC.TypeLits (ErrorMessage (ShowType, Text, (:<>:)), TypeError)
+
+data PSubtypeRelation
+  = SuperType
```
L-as, line 28: We should find better names

SeungheonOh: `IsSubType`?

L-as: PSubtypeRelation PNoSubtypeRelation

SeungheonOh: Done.

L-as (approved):

PR #561, opened 2022-08-29
Utility functions for `PMaybe` and `PMaybeData`

SeungheonOh:
A one of new PRs that will be made for Liqwid-Plutarch-Extra upstreaming.

PR #562, opened 2022-08-29
Utility functions for `PMap`

SeungheonOh:
Depending on #561


plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,335 @@
+{-# LANGUAGE QuantifiedConstraints #-}
+
+{- |
+ Module: Plutarch.Extra.Map
+ Copyright: (C) Liqwid Labs 2022
+ License: Apache 2.0
```
L-as, line 6: ? I can't accept it with this in. The license must be MIT. In addition, we don't have any header like this anywhere, albeit I'd be OK with it being shortened to a 2 lines with just the original author and sponsor.

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,335 @@
...
+import qualified Plutarch.List as PList
+
+{- | If a value exists at the specified key, apply the function argument to it;
+ otherwise, do nothing.
+
+ This is necessarily linear in the size of the map performance-wise, as we
+ have to scan the entire map to find the key in the worst case.
+-}
+padjust ::
+    forall (k :: S -> Type) (v :: S -> Type) (s :: S).
```
L-as, line 62: PType

L-as (changes requested):

L-as:
I can't update this PR.

SeungheonOh:
I'll make a new PR from my fork. Liqwid-Labs fork seems to have problem with permission

PR #563, opened 2022-08-30
PFromInteger: Better PPositive Show + HasCallStack

tbidne:
* Adds the integer to `PPositive`'s `pfromInteger` exception description.
* Adds `HasCallStack` to `pfromInteger`.

### DIFF
--- a/Plutarch/Positive.hs
+++ b/Plutarch/Positive.hs
@@ -4,6 +4,7 @@
 module Plutarch.Positive (PPositive, ppositive, ptryPositive) where
 
 import Data.Functor.Const (Const)
+import Data.Text (pack)
 import GHC.Generics (Generic)
 
 import Plutarch.Bool (PEq, POrd, PPartialOrd, pif, (#<=))
@@ -11,6 +12,7 @@ import Plutarch.Builtin (PAsData, PData, PIsData, pdata)
 import Plutarch.Integer (PInteger, PIntegral)
 
 import Plutarch.Maybe (PMaybe (PJust, PNothing))
+import Plutarch.Show (pshow)
 
 import Plutarch (
   DerivePlutusType (DPTStrat),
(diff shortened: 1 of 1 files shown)


Plutarch/Num.hs
```diff
@@ -38,8 +39,8 @@ class PNum (a :: PType) where
   default psignum :: PNum (PInner a) => Term s (a :--> a)
   psignum = punsafeCoerce (psignum :: Term s (PInner a :--> PInner a))
 
-  pfromInteger :: Integer -> Term s a
-  default pfromInteger :: PNum (PInner a) => Integer -> Term s a
+  pfromInteger :: HasCallStack => Integer -> Term s a
```
L-as, line 42: I don't think this does anything in practice, because you're using `fromInteger` implicitly. If you want this to work, you need to do this through `RebindableSyntax` and add it to `fromInteger` directly.

L-as: You never use pfromInteger in practice. It's almost purely internal. It's only use is to make overloaded integers work.

L-as: Yes, but this isn't actually useful. It doesn't contain anything we don't already know. If you're willing to take on the task (not so big), you can make a `Plutarch.Rebindable` module that would function as an alternative prelude.
See https://downloads.haskell.org/ghc/9.4.1/docs/users_guide/exts/rebindable_syntax.html

Do-syntax should do the same thing as Plutarch.Monadic, but if possible, also work for monads.
If-expressions should obviously use pif, but again, work with Haskell booleans too if possible.
The same general idea should apply to all of the rebindable names.

In addition, a HasCallStack constraint should be added to _all_ functions here.

L-as: Sure

L-as (approved):
LGTM

PR #572, opened 2022-09-08
PPair instances: PPartialOrd and POrd

cjay:


Plutarch/Pair.hs
```diff
@@ -18,3 +18,24 @@ data PPair (a :: PType) (b :: PType) (s :: S)
   deriving anyclass (PlutusType, PEq, PShow)
 
 instance DerivePlutusType (PPair a b) where type DPTStrat _ = PlutusTypeScott
+
+instance (PPartialOrd a, PPartialOrd b) => PPartialOrd (PPair a b) where
```
L-as, line 22: Why not derive this

Plutarch/Pair.hs
```diff
@@ -18,3 +18,24 @@ data PPair (a :: PType) (b :: PType) (s :: S)
...
+  a #< b = unTermCont $ do
+    PPair a1 a2 <- tcont $ pmatch a
+    PPair b1 b2 <- tcont $ pmatch b
+    pure $
+      pif
+        (a1 #== b1)
+        (a2 #< b2)
+        (a1 #< b1)
+
+instance (POrd a, POrd b) => POrd (PPair a b)
```
L-as, line 41: you can just add this to `deriving anyclass`

Plutarch/Pair.hs
cjay: I couldn't find infrastructure for deriving that. Is there any? 
Sorry, I haven't really had time to learn all possibilities of deriving-via yet.

L-as: You can derive all classes that have a default for every method.

cjay: The default needs an `POrd` instance for `PInner` of the type, which in the case of `PPair` is `PScottEncoded ...`. That one doesn't have a `POrd` instance. Are you sure this should be possible somehow?

L-as: This is an issue. Look at the default methods. You can change the constraint to PPartialOrd, but what about deriving POrd? You can add the constraint as a superclass, but perhaps adding a dummy method with a POrd constraint on PInner works best.

L-as:
bump

Plutarch/Pair.hs
cjay: What mechanism should cause the lexical ordering behavior that you see in my manual implementation? I don't see how changing constraints would help with that. I suppose a better default implementation would have to use generics.

PR #575, opened 2022-09-14
Using plutarch-core

Icelandjack:
These are just preliminary changes


Icelandjack:
It includes a lot of the `PPlutus'` changes, I will have to remove that later :)

L-as:
- generics stuff is part of backend
- pinner can be type family
- reduce reduces newtypes for use in HKDs

PR #580, opened 2022-09-20
`Plutarch.Extra.List` and `Plutarch.Extra.Ord`

SeungheonOh:
This PR should allow edits by maintainers now. It seems like I have to use personal fork to allow this due to some permission issue.

### DIFF
--- a/plutarch-extra/Plutarch/Extra/List.hs
+++ b/plutarch-extra/Plutarch/Extra/List.hs
@@ -1,6 +1,116 @@
-module Plutarch.Extra.List (preverse, pcheckSorted) where
+module Plutarch.Extra.List (
+  -- * Construction
+  preplicate,
+
+  -- * Transformation
+  pmapMaybe,
+  preverse,
+
+  -- * Search
+  pfindJust,
+  plookupAssoc,
+
+  -- * Elimination
+  precListLookahead,
+
+  -- * Check
+  pcheckSorted,
+) where
 
 import Plutarch.Prelude
+
+{- | Similar to 'pmap', but allows elements to be thrown out. More precisely,
+ for elements where the function argument returns 'PNothing', the
+ corresponding element is removed, while for elements where the function
+ argument returns `PJust`, the value is used in the result.
+-}
+pmapMaybe ::
+  forall (ell :: PType -> PType) (b :: PType) (a :: PType) (s :: S).
+  (PElemConstraint ell a, PElemConstraint ell b, PListLike ell) =>
+  Term s ((a :--> PMaybe b) :--> ell a :--> ell b)
+pmapMaybe = phoistAcyclic $ plam $ \f -> precList (go f) (const pnil)
+  where
+    go ::
+      forall (s' :: S).
+      Term s' (a :--> PMaybe b) ->
+      Term s' (ell a :--> ell b) ->
+      Term s' a ->
+      Term s' (ell a) ->
+      Term s' (ell b)
+    go f self x xs = pmatch (f # x) $ \case
+      PNothing -> self # xs
+      PJust y -> pcons # y #$ self # xs
+
+{- | A combination of 'pmap' and 'pfind', but without needing an intermediate
+ structure. More precisely, searched for the first element in a list-like
+ structure that produces a 'PJust' argument, returning it if found; otherwise,
+ produces 'PNothing'.
+-}
+pfindJust ::
+  forall (b :: PType) (ell :: PType -> PType) (a :: PType) (s :: S).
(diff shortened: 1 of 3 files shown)


SeungheonOh:
Test will be added shortly

plutarch-extra/Plutarch/Extra/List.hs
```diff
@@ -1,7 +1,117 @@
...
 
 import Plutarch.Prelude
 
+{- | Similar to 'pmap', but allows elements to be thrown out. More precisely,
+ for elements where the function argument returns 'PNothing', the
+ corresponding element is removed, while for elements where the function
+ argument returns `PJust`, the value is used in the result.
+-}
+pmapMaybe ::
+  forall (ell :: PType -> PType) (b :: PType) (a :: PType) (s :: S).
```
L-as, line 28: ell?

plutarch-extra/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,681 @@
...
+    EQ -> 1
+    GT -> 2
+  pconstantFromRepr = \case
+    0 -> pure LT
+    1 -> pure EQ
+    2 -> pure GT
+    _ -> Nothing
+
+instance PlutusType POrdering where
+  type PInner POrdering = PInteger
```
L-as, line 93: why

plutarch-extra/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,681 @@
...
+-- * Arg 0 vs arg 2 and arg 1 versus arg 3
+
+-- * Arg 0 vs arg 1 and arg 2 versus arg 3
+
+-- * Arg 1 versus arg 2
+
+--
+-- We perform the layer parallelism sequentially; it doesn't affect the
+-- semantics, it just makes us sad because it's slow.
+psort4 ::
```
L-as, line 591: What's up with all this duplication? Just make a psortN

L-as: The N would be a Haskell Integer

SeungheonOh: It's heavily optimized by @cjay using some sorting network. You wouldn't get that performance without these.

plutarch-extra/Plutarch/Extra/Ord.hs
SeungheonOh: Should this use SocttPlut?

kozross: Possibly. I did the first thing that could possibly work, which seemed to perform OK, and actually allowed some optimizations (which I have measurements for).

plutarch-extra/Plutarch/Extra/Ord.hs
L-as: There is a huge amount of code duplication still. It seems like you can encode the sorting network as a table of numbers since the code is mostly the same. You can err for N more than 4.

plutarch-extra/Plutarch/Extra/Ord.hs
L-as: This is less efficient to match on and not really cheaper to construct. Scott-encoding is probably better.

plutarch-extra/Plutarch/Extra/List.hs
kozross: Like the capital letter of the alphabet. Calling to mind 'List'.

plutarch-extra/Plutarch/Extra/Ord.hs
cjay: If you mean a table of numbers on the Haskell level, yeah it could probably be done while generating the same code on the script level, but that seems like a pointless exercise to me, unless you want to work with much larger sorting networks.
If you mean a table on the Plutarch-level, that seems useless if you want to sort arbitrary element types. If you can serialize into bytestrings then yeah, you could work with actual indices. But then you'd have to build the intermediate bytestrings. Maybe it would save significant script size on larger sorting networks, but probably with much higher execution cost.

L-as: The former, it's much more succinct than adding a new function every time.

L-as: Also, you can have a generic fallback for n > 4.

L-as:
Do you still want this?

SeungheonOh:
Yes It would be nice to have, but not urgent.

PR #581, opened 2022-09-21
Ensure that zero values get rejected from Value singleton functions

kozross:
Currently, the functions `pconstantSingleton`, `pconstantPositiveSingleton`, `psingleton` and `psingletonData` do not properly reject zero arguments. While technically what they do is not wrong (they don't end up containing zero amounts), it's still not correct behaviour: an empty `PValue` is not a 'singleton' by any definition.

This PR fixes this problem to instead error if given a zero argument in those cases.

### DIFF
--- a/Plutarch/Api/V1/Value.hs
+++ b/Plutarch/Api/V1/Value.hs
@@ -247,7 +247,7 @@ pconstantSingleton ::
   ClosedTerm PInteger ->
   ClosedTerm (PValue 'Sorted 'NonZero)
 pconstantSingleton symbol token amount
-  | plift amount == 0 = mempty
+  | plift amount == 0 = ptraceError "pconstantSingleton: Given a zero argument, expecting nonzero."
   | otherwise = punsafeDowncast (AssocMap.psingleton # symbol #$ AssocMap.psingleton # token # amount)
 
 -- | Construct a constant singleton 'PValue' containing only the given positive quantity of the given currency.
@@ -257,7 +257,7 @@ pconstantPositiveSingleton ::
   ClosedTerm PInteger ->
   ClosedTerm (PValue 'Sorted 'Positive)
 pconstantPositiveSingleton symbol token amount
-  | plift amount == 0 = mempty
+  | plift amount == 0 = ptraceError "pconstantPositiveSingleton: Given a zero argument, expecting nonzero."
   | plift amount < 0 = error "Negative amount"
   | otherwise = punsafeDowncast (AssocMap.psingleton # symbol #$ AssocMap.psingleton # token # amount)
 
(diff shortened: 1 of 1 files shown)


Plutarch/Api/V1/Value.hs
```diff
@@ -247,7 +247,7 @@ pconstantSingleton ::
   ClosedTerm PInteger ->
   ClosedTerm (PValue 'Sorted 'NonZero)
 pconstantSingleton symbol token amount
-  | plift amount == 0 = mempty
+  | plift amount == 0 = ptraceError "pconstantSingleton: Given a zero argument, expecting nonzero."
```
L-as, line 250: pthrow

Plutarch/Api/V1/Value.hs
```diff
@@ -257,7 +257,7 @@ pconstantPositiveSingleton ::
   ClosedTerm PInteger ->
   ClosedTerm (PValue 'Sorted 'Positive)
 pconstantPositiveSingleton symbol token amount
-  | plift amount == 0 = mempty
+  | plift amount == 0 = ptraceError "pconstantPositiveSingleton: Given a zero argument, expecting nonzero."
```
L-as, line 260: pthrow

Plutarch/Api/V1/Value.hs
```diff
@@ -270,7 +270,7 @@ psingleton = phoistAcyclic $
   plam $ \symbol token amount ->
     pif
       (amount #== 0)
-      mempty
+      (ptraceError "psingleton: Given a zero argument, expecting nonzero.")
```
L-as, line 273: This is a huge breaking change. I suggest either changing the type to use GDP or make a new function.

Plutarch/Api/V1/Value.hs
```diff
@@ -286,7 +286,7 @@ psingletonData = phoistAcyclic $
   plam $ \symbol token amount ->
     pif
       (amount #== zeroData)
-      mempty
+      (ptraceError "psingletonData: Given a zero argument, expecting nonzero.")
```
L-as, line 289: same issue

Plutarch/Api/V1/Value.hs
kozross: I would argue the original implementation is a bug. The name 'singleton' has pretty strong implications of behaviour: I certainly don't expect any _other_ function with such a name to potentially return an empty structure sometimes.

kozross: At _minimum_, the documentation should be yelling this from the rooftops.

L-as: When is the current behaviour problematic in practice? I think it's precisely what you want.

kozross: If I give 0 as an argument, what I am getting is not a 'singleton', since it's empty. I would argue that this is misleading _at best_. Thus, either the name is wrong, or the semantics are wrong.

L-as: But this is a map from an index to a monoid. When do you care whether the entry is there or not when you want the value to be zero?

kozross: Yeah, upon reading the comment regarding currencies in `plutus-ledger-api`, that makes sense. Gonna withdraw this PR.

PR #582, opened 2022-09-22
Seungheonoh/extramap

SeungheonOh:
replaces #562 due to permission issue.

### DIFF
--- a/plutarch-extra/Plutarch/Extra/Map.hs
+++ b/plutarch-extra/Plutarch/Extra/Map.hs
@@ -1,0 +1,364 @@
+{-# LANGUAGE QuantifiedConstraints #-}
+
+{- |
+ Module: Plutarch.Extra.Map
+ Copyright: (C) Liqwid Labs 2022
+ License: Apache 2.0
+ Maintainer: Koz Ross <koz@mlabs.city>
+ Portability: GHC only
+ Stability: Experimental
+
+ Various helpers for 'PMap' use. This module is intended to be imported
+ qualified.
+-}
+module Plutarch.Extra.Map (
+    -- * Lookup
+    ptryLookup,
+
+    -- * Comparisons
+    pkeysEqual,
+    pkeysEqualUnsorted,
+
+    -- * Modification
+    pupdate,
+    padjust,
+
+    -- * Folds
+    pfoldMapWithKey,
+    pfoldlWithKey,
+
+    -- * Conversion
+    punsortedMapFromFoldable,
+    psortedMapFromFoldable,
+    pkeys,
+
+    -- * Key-value pair manipulation
+    pkvPairKey,
+    pkvPairValue,
+    pkvPairLt,
+) where
+
+import Plutarch.Prelude
+import Data.Foldable (foldl')
+import Plutarch.Api.V1.AssocMap (
+    KeyGuarantees (Sorted, Unsorted),
+    PMap (PMap),
+    pdelete,
+    pempty,
+    pinsert,
+    plookup,
+ )
+import Plutarch.Builtin (ppairDataBuiltin)
+import Plutarch.Extra.Maybe (passertPJust)
+import qualified Plutarch.List as PList
+
+{- | If a value exists at the specified key, apply the function argument to it;
+ otherwise, do nothing.
+
+ This is necessarily linear in the size of the map performance-wise, as we
+ have to scan the entire map to find the key in the worst case.
+
+ @since 3.3.0
+-}
+padjust ::
+    forall (k :: S -> Type) (v :: S -> Type) (s :: S).
+    (PIsData k, PEq k, PIsData v) =>
+    Term s ((v :--> v) :--> k :--> PMap 'Unsorted k v :--> PMap 'Unsorted k v)
+padjust = phoistAcyclic $
+    plam $ \f key kvs ->
+        pmatch kvs $ \(PMap kvs') ->
+            pcon . PMap $ PList.pmap # (go # f # key) # kvs'
+  where
+    go ::
+        forall (s' :: S).
+        Term
+            s'
+            ( (v :--> v)
+                :--> k
+                :--> PBuiltinPair (PAsData k) (PAsData v)
+                :--> PBuiltinPair (PAsData k) (PAsData v)
+            )
+    go = phoistAcyclic $
+        plam $ \f target kv ->
+            pif
+                ((pkvPairKey # kv) #== target)
+                (ppairDataBuiltin # (pfstBuiltin # kv) # pdata (f #$ pkvPairValue # kv))
+                kv
+
+{- | As 'pkeysEqual', but requires only 'PEq' constraints for the keys, and
+ works for 'Unsorted' 'PMap's. This requires a number of equality comparisons
+ between keys proportional to the product of the lengths of both arguments:
+ that is, this function is quadratic.
+
+ @since 3.3.0
+-}
+pkeysEqualUnsorted ::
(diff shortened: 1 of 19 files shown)


plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+      forall (s' :: S).
+      Term
+        s'
+        ( (v :--> v)
+            :--> k
+            :--> PBuiltinPair (PAsData k) (PAsData v)
+            :--> PBuiltinPair (PAsData k) (PAsData v)
+        )
+    go = phoistAcyclic $
+      plam $ \f target kv ->
```
L-as, line 69: rename target to key or key to target

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+ have to scan the entire map to find the key in the worst case.
+-}
+padjust ::
+  forall (k :: PType) (v :: PType) (s :: S).
+  (PIsData k, PEq k, PIsData v) =>
+  Term s ((v :--> v) :--> k :--> PMap 'Unsorted k v :--> PMap 'Unsorted k v)
+padjust = phoistAcyclic $
+  plam $ \f key kvs ->
+    pmatch kvs $ \(PMap kvs') ->
+      pcon . PMap $ PList.pmap # (go # f # key) # kvs'
```
L-as, line 57: Do we not have a pmap on maps?

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+
+{- | As 'pkeysEqual', but requires only 'PEq' constraints for the keys, and
+ works for 'Unsorted' 'PMap's. This requires a number of equality comparisons
+ between keys proportional to the product of the lengths of both arguments:
+ that is, this function is quadratic.
+-}
+pkeysEqualUnsorted ::
+  forall (k :: PType) (a :: PType) (b :: PType) (s :: S).
+  (PIsData k, PIsData a, PIsData b) =>
+  Term s (PMap 'Unsorted k a :--> PMap 'Unsorted k b :--> PBool)
```
L-as, line 83: When is this more efficient than just sorting before checking?

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+
+{- | Given a 'Foldable' of key-value pairs, construct an unsorted 'PMap'.
+ Performs linearly with respect to its argument.
+
+ = Note
+
+ If there are duplicate keys in the input, the /last/ key will \'win\' in a
+ lookup.
+-}
+punsortedMapFromFoldable ::
```
L-as, line 216: Isn't this basically an IsList instance? Why not just implement that?

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+
+ If there are duplicate keys, only the /last/ key-value pair will remain in
+ the result.
+-}
+psortedMapFromFoldable ::
+  forall (k :: PType) (v :: PType) (f :: Type -> Type) (s :: S).
+  (Foldable f, POrd k, PIsData k, PIsData v) =>
+  f (Term s k, Term s v) ->
+  Term s (PMap 'Sorted k v)
+psortedMapFromFoldable = foldl' go pempty
```
L-as, line 245: You could also make an instance for this.

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -0,0 +1,324 @@
...
+ 'PNothing', the value is deleted; otherwise, it is modified to the result.
+
+ Performance will be equivalent to a lookup followed by an insert (or delete),
+ as well as the cost of calling the \'updater\'.
+-}
+pupdate ::
+  forall (k :: PType) (v :: PType) (s :: S).
+  (PIsData k, PIsData v, POrd k) =>
+  Term s ((v :--> PMaybe v) :--> k :--> PMap 'Sorted k v :--> PMap 'Sorted k v)
+pupdate = phoistAcyclic $
```
L-as, line 296: This is quite inefficiently implemented.

L-as (changes requested):

L-as:
Thanks!

plutarch-extra/Plutarch/Extra/Map.hs
L-as: Make pmapKeyed?

SeungheonOh: okay

plutarch-extra/Plutarch/Extra/Map.hs
SeungheonOh: It's more efficient in terms of memories

plutarch-extra/Plutarch/Extra/Map.hs
SeungheonOh: I don't think you can implement `toList` if we were to implement `IsList`

SeungheonOh: at list without `plift`s which we don't want.

plutarch-extra/Plutarch/Extra/Map.hs
L-as: You should note that

plutarch-extra/Plutarch/Extra/Map.hs
L-as: I'm not sure what you mean.

SeungheonOh: https://github.com/Liqwid-Labs/plutarch-quickcheck/blob/34f1157dd5d68d289beda59fc5c6f68da42d42c9/src/Plutarch/Test/QuickCheck/Instances.hs#L829

I can't think of a better way here. And I'm sure this isn't what we want.

kozross: Yeah, the `toList` direction would be _horrid_.

L-as:
```haskell
instance IsList (Term s (PMap 'Unsorted k v)) where
  type Item (Term s (PMap 'Unsorted k v)) = [(Term s k, Term s v)]
  fromList = punsortedMapFromFoldable

instance IsList (Term s (PMap 'Sorted k v)) where
  type Item (Term s (PMap 'Unsorted k v)) = [(Term s k, Term s v)]
  fromList = psortedMapFromFoldable
```

SeungheonOh: wait we can make `IsList` with no `toList`?

plutarch-extra/Plutarch/Extra/Map.hs
```diff
@@ -250,6 +254,16 @@ psortedMapFromFoldable = foldl' go pempty
       Term s' (PMap 'Sorted k v)
     go acc (key, val) = pinsert # key # val # acc
 
+instance (PIsData k, PIsData v, POrd k) => IsList (Term s (PMap 'Unsorted k v)) where
```
L-as, line 257: Orphan instances are very bad. Just move the relevant functions into the defining module.

L-as: to the module that defines PMap

SeungheonOh: Ah, okay. Thank you.

L-as (approved):
LGTM

PR #586, opened 2022-10-04
Unsure if correct (!): point out that Type is the same as *

lemmon-714:
Figured it's better to propose a change than create an issue. Is this edit true? Is this behaviour of mine correct?

Follow-up question: Why does it sometimes print Type and sometimes *? Should be added as well, such little unclarities tend to block learning, at least for me.


lemmon-714:
And I see it's pointed out above.

MangoIV:
I don't think that's relevant, that's Haskell knowledge, not Plutarch knowledge.

PR #588, opened 2022-10-04
Added another example, please check if correct

lemmon-714:
Am uncertain whether 
a) this is correct at all
b) the usage of h's and P's is correct

Also, please let me know in case I'm using the system incorrectly and should instead ask on discord first, thank you.

I added it below instead of editing the existing one since the latter way might mess with the didactic flow.

### DIFF
--- a/docs/Introduction/Plutarch Terms/Plutarch Constants.md
+++ b/docs/Introduction/Plutarch Terms/Plutarch Constants.md
@@ -97,6 +97,13 @@ These types deserve some explaination.
 
 That is, if we ask `justTerm` what it will return when evaluated, it responds, "You should interpret the value I give you as either `Nothing` or `Just Integer`." Of course, we know that the result will always be `Just 3`; but this is the general mechanism to declare a function requiring a `Maybe`.
 
+If you don't want to pretend to not know `x` during compile time, another example may be:
+
+```hs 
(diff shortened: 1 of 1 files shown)


docs/Introduction/Plutarch Terms/Plutarch Constants.md
```diff
@@ -97,6 +97,13 @@ These types deserve some explaination.
 
 That is, if we ask `justTerm` what it will return when evaluated, it responds, "You should interpret the value I give you as either `Nothing` or `Just Integer`." Of course, we know that the result will always be `Just 3`; but this is the general mechanism to declare a function requiring a `Maybe`.
 
+If you don't want to pretend to not know `x` during compile time, another example may be:
+
+```hs 
+hPJustPInteger :: Term s PInteger -> Term s (PMaybe PInteger)
+hPJustPInteger x = pcon (PJust x)
```
L-as, line 104: what's up with the naming? looks good otherwise

lemmon-714: Following the scheme outlined in the beginning of the docs of naming Plutarch stuff p*, and Haskell stuff h*. It's not merely about strict adherence to those conventions but about detecting misunderstanding. If that question does not make much sense please disregard, it will fall into place later.

PR #589, opened 2022-10-04
Correct pif to hif; add apostrophes to distinguish

lemmon-714:
- imo it should say "lazy `hif`", as we are in Haskell-land again. I don't know if regular `pif` is Haskell-land as well; if so, imo should be renamed as well.
- the wrapper is around `pif'`, not `pif`
- `hif` is already used in the beginning, renamed to `hif'`

### DIFF
--- a/docs/Introduction/Delay and Force.md
+++ b/docs/Introduction/Delay and Force.md
@@ -25,12 +25,12 @@ A delayed term evaluates when it is _forced_ using the `pforce` function. Forcin
 pforce :: Term s (PDelayed a) -> Term s a
 ```
 
-Thus, if we wanted a lazy `pif`, we could do the following:
+Thus, if we wanted a lazy `hif`, we could do the following:
 
 ```hs
--- | Utilizing Haskell level functions with `pdelay` and `pforce` to have lazy wrapper around `pif`.
-hif :: Term s PBool -> Term s a -> Term s a -> Term s a
-hif cond whenTrue whenFalse = pforce $ pif' # cond # pdelay whenTrue # pdelay whenFalse
+-- | Utilizing Haskell level functions with `pdelay` and `pforce` to have lazy wrapper around `pif'`.
+hif' :: Term s PBool -> Term s a -> Term s a -> Term s a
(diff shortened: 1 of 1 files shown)


L-as:
Where does hif come from?

lemmon-714:
It is used without explicit definition in the beginning of the file as haskell-if (can't use regular if-then-else because of the then-else, which the pif lacks). Would assume it has the definition
```hs
hif :: Bool -> a -> a
hif cond whenTrue whenFalse = if cond then whenTrue else whenFalse
```
Shall I add it?

L-as:
Yes please

docs/Introduction/Delay and Force.md
```diff
@@ -25,12 +25,12 @@ A delayed term evaluates when it is _forced_ using the `pforce` function. Forcin
 pforce :: Term s (PDelayed a) -> Term s a
 ```
 
-Thus, if we wanted a lazy `pif`, we could do the following:
+Thus, if we wanted a lazy `hif`, we could do the following:
```
L-as, line 32: I think this _should_ be pif.
It's the part below that's incorrect.

lemmon-714: wdym part below, my edit or the function following?
```hs
-- | Utilizing Haskell level functions with `pdelay` and `pforce` to have lazy wrapper around `pif`.
hif :: Term s PBool -> Term s a -> Term s a -> Term s a
hif cond whenTrue whenFalse = pforce $ pif' # cond # pdelay whenTrue # pdelay whenFalse
```
If the former, I think it's hif because it's haskell-level, not plutarch level. But it would also be better to specify in the definition of hif at the top that it applies only to `Term`s, as that hints towards it being non-lazy.

L-as: This should be called `pif'`, sorry for the very late response. It's not a Haskell-if after all.

cjay:
Superceded by https://github.com/Plutonomicon/plutarch-plutus/pull/608

L-as:
Thanks for your work!

PR #594, opened 2022-10-16
Make ptruncate use pquot instead of emulating it

cjay:


L-as:
Are we sure QuotientInteger has the same semantics?

cjay:
I haven't found a specification of what the builtin is supposed to do exactly, but the existing tests would catch if it rounded differently, so pretty sure it behaves like `quot` in Haskell.

```
tests/Plutarch/RationalSpec.hs
4:import Plutarch.Rational (pproperFraction, ptruncate)
32:      "5/4" @| ptruncate # rat (5 / 4) @== pconstant @PInteger 1
33:      "7/4" @| ptruncate # rat (7 / 4) @== pconstant @PInteger 1
34:      "1/4" @| ptruncate # rat (1 / 4) @== pconstant @PInteger 0
35:      "-7/4" @| ptruncate # rat (-7 / 4) @== pconstant @PInteger (-1)
```

L-as:
you can check the implementation

cjay:
If `PlutusTx.Builtins.Internal` is the right place: `quotientInteger` makes use of the normal `quot`.

L-as (approved):
Looks good to me

PR #601, opened 2022-11-18
Fix V1 script types being re-exported in V2 api

TotallyNotChase:
I'll leave the versioning (in case of newer release) up to @L-as


L-as (approved):
LGTM

PR #602, opened 2022-11-28
add constructor export for `Script` type, add `deserialiseScript`

emiflake:
`Script` previously did export its constructor when it lived in `plutus-ledger-api`, so I think it should do if it lives in Plutarch, too. Additionally, this adds `deserialiseScript`. We had `serialiseScript` already, but not deserialise.


L-as (approved):
LGTM

PR #604, opened 2022-12-04
Add Intersection to Maps

lemmon-714:
I copied and renamed the union-function, except for the four marked spots. Manually tested only.

Refactoring is possible.

### DIFF
--- a/Plutarch/Api/V1/AssocMap.hs
+++ b/Plutarch/Api/V1/AssocMap.hs
@@ -39,6 +39,8 @@ module Plutarch.Api.V1.AssocMap (
   pdifference,
   punionWith,
   punionWithData,
+  pinsecWith,
+  pinsecWithData,
 
   -- * Partial order operations
   pcheckBinRel,
@@ -407,7 +409,7 @@ instance
   where
   inv a = pmap # plam PlutusTx.inv # a
 
-{- | Combine two 'PMap's applying the given function to any two values that
+{- | Union two 'PMap's applying the given function to any two values that
  share the same key.
 -}
 punionWith ::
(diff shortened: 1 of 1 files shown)


lemmon-714:
My refactoring WIP is btw extracting the four marked spots into a config type; this would not only circumvent the copypasting but also allow adding more merge types; in particular 
- some kind of vector-arithmetic-like, which errors if there is any mismatch between keys;
- something to be used with PValue that drops a specified null (0 resp. the empty Map in that case), avoiding the need to pnormalize after each union. 

But that is postponed for premature optimisation reasons.

L-as:
Sorry for not getting back to this, I will take a look at this ASAP

lemmon-714:
Lol no worries mate, I'm not in a rush. Thank you

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -407,7 +409,7 @@ instance
   where
   inv a = pmap # plam PlutusTx.inv # a
 
-{- | Combine two 'PMap's applying the given function to any two values that
+{- | Union two 'PMap's applying the given function to any two values that
```
cjay, line 412: Union doesn't seem to be a verb, so I'd word those as "build the union/intersection of".

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -463,10 +465,10 @@ mapUnionCarrier = phoistAcyclic $ plam \combine self ->
                           )
           }
 
-mapUnion :: forall k v s. (POrd k, PIsData k) => Term s ((PAsData v :--> PAsData v :--> PAsData v) :--> MapUnionCarrier k v)
-mapUnion = phoistAcyclic $ plam \combine -> punsafeCoerce pfix # (mapUnionCarrier # combine :: Term _ (MapUnionCarrier k v :--> MapUnionCarrier k v))
+mapUnion :: forall k v s. (POrd k, PIsData k) => Term s ((PAsData v :--> PAsData v :--> PAsData v) :--> MapMergeCarrier k v)
+mapUnion = phoistAcyclic $ plam \combine -> punsafeCoerce pfix # (mapUnionCarrier # combine :: Term _ (MapMergeCarrier k v :--> MapMergeCarrier k v))
 
-{- | Combine two 'PMap's applying the given function to any two data-encoded
+{- | Union two 'PMap's applying the given function to any two data-encoded
```
cjay, line 471: Same here.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -480,7 +482,68 @@ punionWithData ::
     )
 punionWithData = phoistAcyclic $
   plam $ \combine x y ->
-    pcon $ PMap $ (pmatch (mapUnion # combine) \(MapUnionCarrier {merge}) -> merge) # pto x # pto y
+    pcon $ PMap $ (pmatch (mapUnion # combine) \(MapMergeCarrier {merge}) -> merge) # pto x # pto y
+
+{- | Intersect two 'PMap's applying the given function to any two values that
+ share the same key.
+-}
+pinsecWith ::
```
cjay, line 490: `insec` is an unusual abbreviation, I'd just use `intersect`. Multiple places.

Plutarch/Api/V1/AssocMap.hs
```diff
@@ -480,7 +482,68 @@ punionWithData ::
...
+                              (pfromData xk #< pfromData yk)
+                              (mergeInsert # y # ys' # xs) -- diff to mapUnionCarrier here
+                              (mergeInsert # x # xs # ys') -- diff to mapUnionCarrier here
+                          )
+          }
+
+mapInsec :: forall k v s. (POrd k, PIsData k) => Term s ((PAsData v :--> PAsData v :--> PAsData v) :--> MapMergeCarrier k v)
+mapInsec = phoistAcyclic $ plam \combine -> punsafeCoerce pfix # (mapInsecCarrier # combine :: Term _ (MapMergeCarrier k v :--> MapMergeCarrier k v))
+
+{- | Intersect two 'PMap's applying the given function to any two data-encoded
```
cjay, line 532: Instead of "applying", it could be more explicit: "merging values that share the same key, using the given function".

cjay (changes requested):
Looks good otherwise.

cjay:
Superseded by https://github.com/Plutonomicon/plutarch-plutus/pull/610.

L-as:
Thanks for the work! This has been merged now.

Issue #556, opened 2022-08-22
punionWith flips arguments


facundominguez:
I found the following inputs to `punionWith` which seem to flip the order of the arguments.

```
firstPValue:
  [ (0x,[(0x,3000000)])
  , (0x5ad56f5..., [(0xfb8c7da5...,1)])
  , (0xa912171...,
      [ (0x676f6c6...,24)
      , (0x73696c7.,91)
      ]
    )
  ]

secondPValue:
  [(0xa912171..., [(0x676f6c6...,1)])]

punionWith (-) firstPValue secondPValue:
  [ (0x,[(0x,3000000)])
  , (0x5ad56f5..., [(0xfb8c7da5...,1)])
  , (0xa912171...,
      [ (0x676f6c6...,-23)
      , (0x73696c7...,91)
      ]
    )
  ]
```

It turns out that `punionWith (-)`, ends up calling this code which does appear to flip arguments:
https://github.com/Plutonomicon/plutarch-plutus/blob/77bbe609d35d430d747472f094ee16bb4cb8fa05/Plutarch/Api/V1/AssocMap.hs#L395

The fix could be to declare that `punionWith` is to be used with commutative operations only, or it could be that `punionWith` needs to stop flipping arguments. In any case, I'd be thankful for any hints on how `punionWith` is expected to work.

L-as:
This is horrible @blamario

L-as:
I'm not sure the code will still work if you change that snippet, but we could keep track of flipping to flip application to the combine function if necessary.

L-as:
Brian Kush just noted that (obviously) `punionWith` would never work with `-`, because if you `punionWith (-)` with two disjoint values, you get the sum necessarily.
`punionWith` needs to be completely redesigned. There should be two extra functions for handling the two extra cases where the LHS or RHS doesn't exist in the most efficient way possible.

@blamario

blamario:
Are these two extra functions necessary? I'm not sure `punionWithDefault (-) 0 a b` would be any more efficient or convenient than `punionWith (+) a (pinv b)`. Are there any other use cases?

PR #610, opened 2023-02-07
AssocMap argument order fix, zip, and intersection

cjay:
Supersedes https://github.com/Plutonomicon/plutarch-plutus/pull/604

Goldens ~have~had ex units ~mismatch~improvement unrelated to the changes here.


plutarch-test/goldens/api.map.bench.golden
```diff
@@ -21,8 +21,18 @@ difference.emptyRight {"exBudgetCPU":4494995,"exBudgetMemory":15848,"scriptSizeB
 difference.emptyResult {"exBudgetCPU":7452717,"exBudgetMemory":22006,"scriptSizeBytes":179}
 unionWith.const {"exBudgetCPU":9281337,"exBudgetMemory":29270,"scriptSizeBytes":263}
 unionWith.double {"exBudgetCPU":9372814,"exBudgetMemory":28772,"scriptSizeBytes":259}
-unionWith.(+) {"exBudgetCPU":9144040,"exBudgetMemory":28276,"scriptSizeBytes":265}
-unionWith.flip (+) {"exBudgetCPU":9144040,"exBudgetMemory":28276,"scriptSizeBytes":265}
+unionWith.(+) {"exBudgetCPU":9372814,"exBudgetMemory":28772,"scriptSizeBytes":259}
```
L-as, line 24: This is slightly more expensive now.

L-as (changes requested):
Thanks for fixing this up, but I realised we have a critical issue, see #556 .

This implementation has the same issue. In the case of intersections, there is only one issue, the flipping should just be undone.

In the case of unions, it's more complex, because we need to handle the case where neither side is present. Using default left and right values seems to be the cheapest solution (PMaybe is more expensive).

This is a breaking change. We should up the version to 1.5.

plutarch-test/goldens/api.map.bench.golden
cjay: I changed those to to have matching keys in the argument Maps. Before, the combine function wouldn't even be called.

cjay:
Sorry for the delay. There is one weird bug in `pzipMapsWith` that I haven't figured out yet:
```
  ./Plutarch/Test.hs:87:17: 
  4) Plutarch.Api.api.map <golden>.zipMapsWith.(-).assert
       expected: "(program 1.0.0 [(#4161, #182a), (#4162, #0d), (#4163, #26)])"
        but got: "(program 1.0.0 [(#4161, #182a), (#4162, #0d), (#4163, #08)])"
```

Also, with the mutual recursion trick, adding the extra funcion made things more expensive due to the record lookup. I thought I could move the record lookup outside of the recursion, but that resulted in non-termination. Maybe it's better to pass a bool arg around for the zip to remember arg order, will check the cost.

I also think the mutual recursion isn't even needed outside of the current impl of `zipMapsWith`, since the one `pmatch` that the recursive call of `merge` does could also just be inlined, and then the usual `pfix` could be used.

L-as:
@cjay Do note that records are Scott-encoded, unfold the Scott-encoding in your mind. You will note that it is equivalent to CPS.

L-as:
Also, yeah, you should just use a boolean.

L-as:
@cjay it seems to me like you can merge the union and intersection operations into zipWith, by adding Haskell-level parameters to `zipWith`. `punionWith` etc. should be removed as it's misleading to users and likely hiding bugs. The intersection operation can be done by making it `zipWithFilter` operation by making the combination function return `PMaybe`. The left and right defaults will then default to `PNothing` in the simple intersection case, hence they are removed.

cjay:
I added Haskell-level default args for `zipWith` as you suggested. It allows emulating `intersectionWith` efficiently, and adding a  `zipWithFilter` was not necessary for that, neither was returning a `PMaybe`. I also eliminated the mutual recursion, and the code now uses bools for the order. It is lower cost (on non-trivial maps at least), and introducing Scott-encoded bools helped even more.

However it turned out that the old `unionWith` can't be expressed using `zipWith`, at least not without cheating. Also, many things that used to be implemented in terms of `unionWith` now perform significantly worse, due to more `comibne` calls. Even if `unionWith` gets expressed using `zipWith` by cheating, the huge performance impact would stay.

`unionWith` can't be expressed by using `PNothing` default values and a combine function in terms of `PMaybe`, since the type of the combine function forces it to be able to handle both values being the default. It has no way to create a value from two `PNothing`s. Though one could cheat and error for `PNothing PNothing`, since it never actually gets called with that.

I believe `unionWith` is very useful, for purpose similar to combining values with `<>`. Also note that the Map in Haskell `containers` has `unionWith` as well that works exactly like our `unionWith` did. For operations like `(+)`, using `unionWith` instead of `zipMapsWith` would be much more efficient, since it doesn't call `combine` for many keys.

Luckily I was able to keep the `Semigroup` instance for `PMap` by bumping the contraints on the value type to `Monoid` and using `mempty` for the defaults.

Maybe it would be better pull the combine function itself onto the Haskell-level. It could decide on defaults internally, obviating the need for separate default parameters on `zipMapsWith`, and would return a Haskell `Maybe` to be able to drop keys. Downside is that the whole thing gets potientially hoisted in more variants. Upside is that this function could use `These`, allowing it to solve the above problem with expressing `unionWith`. It could also be used to produce more efficient code for combine operations like `(+)`, by just returning the one argument Term that wasn't missing. Not sure if it would be as efficient as a hand-written `unionWith` though.

cjay:
Type of the Haskell-level funciton would be:
```
combine :: 
  forall (s ::S) (k :: PType) (v ::Ptype). 
  These (Term s v) -> 
  Maybe (Term s s)
```

For intersection: 
```
combineIntersection combine' (These l r) = Just (combine' l r)
combine _ _ = Nothing
```

For old unionWith behavior:
```
combineUnion combine' these = 
  Just $ case these of
     These l r -> combine' l r
     This l -> l
     That r -> r
```

For simple zipping:
```
combineZip default combine' these = 
  Just $ case these of
     These l r -> combine' l r
     This l -> combine' l default
     That r -> combine' default r
```

And similar for left and right outer join eqivalents.

If `unionWith` is a too misleading name, maybe we could call it `unionResolvingCollisions` or something like that.

One thing the hand-written `unionWith` would do better though, is simply returning the rest of one of the lists, when the other one runs out.

L-as:
`combine` should definitely be entirely Haskell-level. You can create wrappers that turn it into a Plutarch-level function to enable more hoisting (if code size is an issue).

As for combining _values_ specifically, yes, when one side is empty, we can simply return the other side. You can however integrate this into your generalised function by allowing you to note whether the cases for the lhs and rhs are simply `id`. You can do this by having the user explicitly tell you whether this is the case, or, if you're up for the task, _check_ whether the function passed in is equivalent to `id`. You can do this because the `s` is rigid from the inside, i.e. `combine` _can't_ inspect the term its passed (unless it uses unsafe tricks). You can take a dummy value, pass it through, and check if you get the same dummy value back. This is of course more work, you might be more satisfied by just allowing the combine function to be something like
```haskell
data CombineBoth k v s = CombineBothConst | CombineBothFlipConst | CombineBoth (Term s k -> Term s v -> Term s v -> Term s v)
data CombineOne k v s = CombineOneId | CombineOneCustom (Term s k -> Term s v -> Term s v)
data Combine k v s = Combine (CombineBoth k v s) (CombineOne k v s) (CombineOne k v s)
```
(don't copy-paste the code verbatim, just take the idea)

L-as:
@cjay

cjay:
Alright this worked out nicely. Value tests with significant size that use `unionWithResolvingCollisions` use slightly more CPU and slightly less memory compared to master (but now respect the argument order). `pdifference` is now expressed using `pzipWith` too.
There is still one optimization opportunity: Keeping track of the argument order is not necessary for commutative operations. Should I add something for that? Making the `PSBool` optional in the recursion of the current `zipWith` infrastructure probably gets too messy, so I suppose I'd have to make a copy.

cjay:
I'm not quite sure why script sizes got so much worse though. Reintroducing the mutual recursion trick does't really help:
```
master:
unionWith.growing.17 {"exBudgetCPU":624137614,"exBudgetMemory":1530496,"scriptSizeBytes":645}
this branch:
unionResolvingCollisionsWith.growing.17 {"exBudgetCPU":629601906,"exBudgetMemory":1458796,"scriptSizeBytes":1157}
going back to mutual recursion:
unionResolvingCollisionsWith.growing.17 {"exBudgetCPU":837291906,"exBudgetMemory":2361796,"scriptSizeBytes":1006}
```

Script size gets slightly better, but memory gets really bad then.
In master one of the mutual recursive functions was hoisted, but this is not really possible now, unless we make `MergeHandler` a Plutarch argument, that would ruin both CPU and memory budget.

cjay:
Mitigated script size increase as far as I can:
```
after deduplicating zipMerge calls inside zipMergeInsert:
unionResolvingCollisionsWith.growing.17 {"exBudgetCPU":657753906,"exBudgetMemory":1581196,"scriptSizeBytes":997}

after pushing argOrder pmatching down into branches as far as possible:
unionResolvingCollisionsWith.growing.17 {"exBudgetCPU":657753906,"exBudgetMemory":1581196,"scriptSizeBytes":935}
```

Moral of the story is to at least `plet` locally if something can't be hoisted, and to push `pmatching` down into branches as far as possible, which doesn't even increase cost.

cjay:
Added commutativity flags, to allow picking the more efficient zipping loop that doesn't care about order. Most Value operations benefit.
A few tests have larger script size due to using both commutative and non-commutative ops.

L-as:
Thanks! Checking this now.

L-as (dismissed):
LGTM, but there is quite a lot of code that would take a long time to properly verify security off. @cjay If you have time would you mind making another PR that adds some property testing for each of these functions?

(waiting for CI to pass)

L-as:
You can run CI locally by just building the appropriate Nix derivations BTW. You can check which ones exist by opening `nix repl` doing `:lf .` and using tab-completion.

MangoIV:
https://paste.sr.ht/~mangoiv/5ff8222d80d705f75c8cbc05255f574c18fe556b

Perhaps this is useful (if you do this, you can run nix flake check when passing —impure and —allow-import-from-derivation)

L-as:
That is very useful indeed! @MangoIV you could probably use an environment variable instead. Might be worth putting into tooling.

L-as:
@cjay It failed again.

PR #620, opened 2023-11-17
Bumping to latest plutus(1.7.0.0)

SeungheonOh:
changes:
- Using CHaP instead of deprecated GHC next packages
- Bumping to GHC 963
- Adding pre-commit hooks: fourmolu, cabal-fmt, nixfmt, deadnix, statix, typos
- Locking cabal project index-state


plutarch-docs/Concepts/GenericProgramming.md
```diff
@@ -126,8 +126,8 @@ instance DerivePlutusType (MyType a b) where type DPTStrat _ = PlutusTypeScott
 -- If you instead want to use data encoding, you should derive 'PlutusType' and provide data strategy:
 
 data MyTypeD (a :: PType) (b :: PType) (s :: S)
-  = OneD (Term s (PDataRecord '[ "_0" ':= a ]))
-  | TwoD (Term s (PDataRecord '[ "_0" ':= b ]))
+  = OneD (Term s (PDataRecord '[ "_0" ' := a ]))
```
L-as, line 129: wtf?

L-as: Two possibilities:
- Fourmolu is misconfigured -> fix configuration
- You can't configure this -> remove leading `'`

SeungheonOh: It's not fourmolu problem. GHC gives warning if `':=` for some reason. 

```
 The suffix use of a ':<>:' might be repurposed as special syntax
      by a future language extension.
    Suggested fix: Add whitespace around ':<>:'.
```

I don't think turning off ticking infix tyop is a good idea either

MangoIV: That’s a GHC bug if you ask me

SeungheonOh: Yeah, yet another weird behavior of GHC parser. It should really enforce this whitespace rule after infix ticks so like `_':<>:_` not `'_:<>:_`

MangoIV: Just remove the tick though? Also perhaps report a GHC bug if you have time?

SeungheonOh: Ticks are quite useful, I don't think we should remove them.

MangoIV: I’m not working with Plutarch anymore so do whatever you fancy but it has been decided that Haskell will be a dependently typed language and GHC doesn’t even warn on missing ticks anymore, they’re merely an awkward way to tell the compiler whether you want the promoted constructor or the datatype in the presence of punning. I’d advocate for simply not using punning, wherever possible (also something where the Haskell ecosystem is decidedly headed btw)

SeungheonOh: > it has been decided that Haskell will be a dependently typed language
Hell, this sounds great. 

> I’d advocate for simply not using punning, wherever possible (also something where the Haskell ecosystem is decidedly headed btw)
This would be ideal, but even prelude itself have a ton of types with punning. At current stage, I don't think we are quite there to remove them entirely. More likely in the future. 

Made issue: https://gitlab.haskell.org/ghc/ghc/-/issues/24208

L-as: Revert change and disable warning or remove tick.

Plutarch/DataRepr/Internal.hs
```diff
@@ -192,7 +192,7 @@ instance
   where
   pshow' b xs = "[" <> pmatch xs go
     where
-      go :: PDataRecord ((label ':= x) : xs) s -> Term s PString
+      go :: PDataRecord ((label ':= x) ': xs) s -> Term s PString
```
L-as, line 195: You still have a lot of these changes. Can you maybe just rebase to before you did this change?
Another option is to remove the ' entirely everywhere, but we shouldn't be adding more of them at least.

L-as (changes requested):

Plutarch/DataRepr/Internal.hs
SeungheonOh: These are the places where ticks should've been in the first place for the consistency.

bladyjoker:
@SeungheonOh what's the hold up here?

bladyjoker (dismissed):

SeungheonOh:
I will add pre-commit hooks.

Plutarch/Api/V1.hs
```diff
@@ -116,7 +116,7 @@ redeemerHash :: Plutus.Redeemer -> Plutus.RedeemerHash
 redeemerHash = coerce . dataHash
 
 -- | Hash the data encoded representation of given argument.
-dataHash :: Plutus.ToData a => a -> Plutus.BuiltinByteString
+dataHash :: (Plutus.ToData a) => a -> Plutus.BuiltinByteString
 dataHash = hashData . Plutus.toData
```
bladyjoker, line 120: Add `single-constraint-parens: never` to fourmolu.yaml

flake.nix
```diff
@@ -11,60 +11,96 @@
...
+          flake = project.flake { };
+        in
+        {
+          pre-commit = {
+            settings = {
+              src = ./.;
+              settings = {
+                ormolu.defaultExtensions = [
+                  "TypeApplications"
+                  "PatternSynonyms"
```
bladyjoker, line 78: Use

```
          settings = {
            ormolu.cabalDefaultExtensions = true;
          };
```

flake.nix
```diff
@@ -11,60 +11,96 @@
...
+                nixpkgs-fmt.enable = true;
+                cabal-fmt.enable = true;
+                fourmolu = {
+                  enable = true;
+                  excludes = [ "\.lhs" ];
+                };
+                hlint.enable = false;
+                statix.enable = true;
+                deadnix.enable = true;
+              };
```
bladyjoker, line 92: I would rather extract the pre-commit stuff into `pre-commit.nix` to keep the flake.nix smaller as I did https://github.com/mlabs-haskell/lambdabuffers-cardano-demo/blob/main/pre-commit.nix.

Then you just add an import as https://github.com/mlabs-haskell/lambdabuffers-cardano-demo/blob/20e43a95ad5f8927985743187ab15d2648bebbfe/flake.nix#L17

Honestly, I would also extract Hercules stuff into hercules-ci.nix.

flake.nix
```diff
@@ -11,60 +11,96 @@
...
-        hercules-ci.github-pages.settings.contents = config.packages.combined-docs;
+              hooks = {
+                nixpkgs-fmt.enable = true;
+                cabal-fmt.enable = true;
+                fourmolu = {
+                  enable = true;
+                  excludes = [ "\.lhs" ];
+                };
+                hlint.enable = false;
+                statix.enable = true;
```
bladyjoker, line 90: please add 

```
typos.enable = true;
markdownlint.enable = true;
```

Mantra; leave no source unchecked

Plutarch/Api/V1.hs
SeungheonOh: Neat. Thanks

flake.nix
SeungheonOh: For small nix files like what we have here, I prefer to keep things together so someone else can get better idea of what's going on without reading a gazillion file. Especially here since Plutarch only uses Haskell, and, therefore, we won't really add much more to the current flake file

flake.nix
SeungheonOh: Didn't know Typos was a thing. This is great!

flake.nix
```diff
@@ -11,60 +11,72 @@
...
+            # TODO: Bring back CI, Bring back docs
+            packages = flake.packages;
+            devShells = flake.devShells;
+            checks.plutarch-test = pkgs.runCommand "plutarch-test"
+              {
+                nativeBuildInputs = [ flake.packages."plutarch-test:exe:plutarch-test" ];
+              } ''
+                plutarch-test
+                touch $out
+              '';
```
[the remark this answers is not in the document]
SeungheonOh: It won't actually run the tests without this

bladyjoker (approved):
Woohooo! Game on @SeungheonOh ! Good stuff! Let's merge!

L-as:
How did you find all the typos?

SeungheonOh:
Yes, `typos` pre-commit hook

PR #622, opened 2023-12-15
Lint and update nix files

peter-mlabs:
- add `hlint` to the precommit
- "adds" `apply-refact` to the dev shell, (but actually comments it out because it is broken)
- lints the codebase
- updates the `.envrc` to track the various nix files

### DIFF
--- a/Plutarch/Api/V1/AssocMap.hs
+++ b/Plutarch/Api/V1/AssocMap.hs
@@ -2,6 +2,9 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
+--  Some functions in this module seem to be sensitive to parentheses
+{-# HLINT ignore "Redundant bracket" #-}
+
 module Plutarch.Api.V1.AssocMap (
   PMap (PMap),
   KeyGuarantees (Unsorted, Sorted),
@@ -714,6 +717,7 @@ zipMergeInsertCommutative (MergeHandlerCommutative bothPresent onePresent) = unT
           PassOne -> pcons # x # xs'
           HandleOne handler ->
             List.pmap
+              {- HLINT ignore "Redundant bracket" -}
               # ( plam \pair ->
                     plet (pfstBuiltin # pair) \k ->
                       ppairDataBuiltin # k # handler k (psndBuiltin # pair)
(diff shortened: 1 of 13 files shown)


Plutarch/Api/V1/AssocMap.hs
```diff
@@ -714,6 +717,7 @@ zipMergeInsertCommutative (MergeHandlerCommutative bothPresent onePresent) = unT
           PassOne -> pcons # x # xs'
           HandleOne handler ->
             List.pmap
+              {- HLINT ignore "Redundant bracket" -}
```
bladyjoker, line 720: This is generally pretty safe, was there a problem with hls applied solution? Fyi, I usually visit this file and HLS provides an action that refactors the thing. I never had a semantic issue with it.

flake.nix
```diff
@@ -61,6 +61,14 @@
                 project.hsPkgs.hspec-discover.components.exes."hspec-discover"
                 project.hsPkgs.markdown-unlit.components.exes."markdown-unlit"
               ];
+              #   FIXME (peter-mlabs): This is currently broken.
```
bladyjoker, line 64: nit; Syntax is `FIXME(peter-mlabs): This is currently broken.` (without the space)

Also FIXME would indicate there's a behavior in need of fixing. Perhaps TODO is more appropriate? As we want to add a new behavior.

flake.nix
```diff
@@ -61,6 +61,14 @@
...
+              #   In the future you can do something like
+              #
+              #     find -name '*.hs' -not -path './dist-*/*' -exec hlint -j --refactor --refactor-options="--inplace" {} +
```
bladyjoker, line 67: well nice! We can make a pre-commit-hooks module that does that. Not sure if it's safe to do such refactors.

bladyjoker (approved):
Thanks @peter-mlabs

flake.nix
peter-mlabs: Wait, there's standard syntax around this?? I've always done this informally! 

What is the spec for this?

bladyjoker: Lol spec! The only thing close to that I've seen was https://peps.python.org/pep-0350/.

I'm hoping to adopt things like https://github.com/pgilad/leasot and have up to date reports with codetags.

`<TODO|FIX|WARN|HACK|WHATEVER>(<username|bug link>): <Description>`

That's what I've seen used in bigger codebases. Emacs magit-todos for example is excellent in representing this.

PR #633, opened 2024-01-16
Fix PRationalData comparison, tests

### LINKED ISSUE #618: POrd, PEq instances for PRationalData seems broken
```
0/1 #< 758566050/758566050 #&& 758566050/758566050 #<= 1/1
```

appears to evaluate to false. The instance is autoderived (while neither `PRational` nor `PlutusTx.Rational` are autoderived), so I'm assuming that either its broken entirely or assumes it is reduced.

Likewise, `1/1 #== 2/2` appears to evaluate to false.

kozross:
Fixes #618. In addition to fixing the issue, tests have been added to make sure comparison operators behave as expected, both on the reported failure cases and also more broadly.

I'm not sure why the failure in `PEq` happened: currently, this shouldn't be the case. It's possible we were working against a pre-normalization version of `Rational` in PlutusTx? Either way, I can't repro that issue, and the tests show no issues as far as I can tell.

@peter-mlabs - this should fix your issue, though I can't add you as a reviewer for some reason.

### DIFF
--- a/plutarch-extra/Plutarch/Extra/RationalData.hs
+++ b/plutarch-extra/Plutarch/Extra/RationalData.hs
@@ -7,7 +7,6 @@ module Plutarch.Extra.RationalData (
 ) where
 
 import Data.Bifunctor (first)
-
 import Plutarch.Builtin (pasConstr)
 import Plutarch.DataRepr (DerivePConstantViaData (DerivePConstantViaData), PDataFields)
 import Plutarch.Extra.TermCont (pguardC, pletC, pletFieldsC)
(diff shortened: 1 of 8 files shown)


plutarch-extra/Plutarch/Extra/RationalData.hs
```diff
@@ -61,3 +68,27 @@ prationalFromData = phoistAcyclic $
...
+      Term s' PRationalData ->
+      Term s' PBool
+    go l r = unTermCont $ do
+      l' <- pletFieldsC @'["numerator", "denominator"] l
+      r' <- pletFieldsC @'["numerator", "denominator"] r
+      let ln = pfromData $ getField @"numerator" l'
+      let ld = pfromData $ getField @"denominator" l'
+      let rn = pfromData $ getField @"numerator" r'
+      let rd = pfromData $ getField @"denominator" r'
+      pure $ f (ln * pto rd) (rn * pto ld)
```
SeungheonOh, line 94: shouldn't it divide not multiply?

SeungheonOh (approved):
LGTM

SeungheonOh:
I'm not sure why it wouldn't be reproducible though. Probably we want to construct the rational from Plutrach(and not from PlutusTx)?

kozross:
I'm not sure how we would construct it from Plutarch. As far as I can tell, you can only use `pconstant` to lift a `Rational`.

SeungheonOh:
Yes, I agree. I can't find any function exposed that allows creation of non-reduced Rational values. 

The changes made in POrd instance and test suite is good

Issue #640, opened 2024-01-17
Migrate plutus-ledger-api types and functions into independent module


SeungheonOh:
UPLC and Plutus Ledger API is independent of each other. UPLC defines the "instruction sets" for the smart contracts while Ledger API defines structure of data provided by nodes to each scripts. Different versions of Ledger API can be "run" on different versions of UPLC scripts. 

Separating Ledger API modules, currently embedded inside of Plutarch core codebase, into an independent library will make it easier to maintain and quicker to drop-in different versions of plutus-ledger-api if necessary.

SeungheonOh:
We have to note that different versions of ledger api will provide different builtin functions(**_need confirmation_**). Ideally, we define wrapper for these builtins in each plutarch-ledger-api library for each respective versions

SeungheonOh:
Note how it's done on aiken:
https://github.com/aiken-lang/stdlib
https://github.com/aiken-lang/prelude

Issue #642, opened 2024-01-18
`MergeHandlerXYZ` can be refactored with less abstraction 


SeungheonOh:
`MergeHandlerXYZ` and it's friends are abstracted and type-parameterized more then it needs to be: it attempts to support beyond Plutarch's terms. We don't need this and its kind signatures are very confusing. This can be refactored with less(or none perhaps) abstraction in mind.

PR #645, opened 2024-01-19
Create plutarch-api

kozross:
Initial repair for #640. Currently, this is minimal: nothing's been rewired yet, and only the exports from `Plutarch.Api.V2` are retained. This is designed to keep review surface minimal before we tackle additional porting and rewiring.

### DIFF
--- a/plutarch-api/src/Plutarch/Api.hs
+++ b/plutarch-api/src/Plutarch/Api.hs
@@ -1,0 +1,1275 @@
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{- | = Note
+
+The 'Value.PValue' and 'AssocMap.PMap'-related functionality can be found in
+other modules, as these clash with the Plutarch prelude. These should be
+imported qualified.
+-}
+module Plutarch.Api (
+  -- * Contexts
+  PScriptContext (..),
+  PTxInfo (..),
+  PScriptPurpose (..),
+
+  -- * Tx
+  PTxOutRef (..),
+  PTxOut (..),
+  PTxId (..),
+  PTxInInfo (..),
+  POutputDatum (..),
+
+  -- * Script
+
+  -- ** Types
+  PDatum (..),
+  PDatumHash (..),
+  PRedeemer (..),
+  PRedeemerHash (..),
+  PScriptHash (..),
+
+  -- ** Functions
+  scriptHash,
+  datumHash,
+  redeemerHash,
+  dataHash,
+
+  -- * Value
+  Value.PValue (..),
+  Value.AmountGuarantees (..),
+  Value.PCurrencySymbol (..),
+  Value.PTokenName (..),
+
+  -- * DCert
+  PDCert (..),
+
+  -- * Assoc map
+
+  -- ** Types
+  AssocMap.PMap (..),
+  AssocMap.KeyGuarantees (..),
+  AssocMap.Commutativity (..),
+
+  -- * Address
+  PCredential (..),
+  PStakingCredential (..),
+  PAddress (..),
+
+  -- * Time
+  PPosixTime (..),
+
+  -- * Interval
+  PInterval (..),
+  PLowerBound (..),
+  PUpperBound (..),
+  PExtended (..),
+
+  -- * Crypto
+
+  -- ** Types
+  PubKey (..),
+  PPubKeyHash (..),
+  pubKeyHash,
+
+  -- * Utilities
+
+  -- ** Types
+  PMaybeData (..),
+
+  -- ** Functions
+  ptuple,
+) where
+
+import Codec.Serialise (serialise)
+import Crypto.Hash (
+  Blake2b_224 (Blake2b_224),
+  Blake2b_256 (Blake2b_256),
+  hashWith,
+ )
+import Data.Bifunctor (first)
+import Data.ByteArray (convert)
+import Data.ByteString (ByteString, toStrict)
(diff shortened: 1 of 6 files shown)


plutarch-api/src/Plutarch/Api/AssocMap.hs
```diff
@@ -0,0 +1,818 @@
...
+              # (ppairDataBuiltin # (pfstBuiltin # x) # (f #$ psndBuiltin # x))
+              # (self # xs)
+        )
+        (const pnil)
+        # pto m
+
+-- Helpers
+
+-- TODO: Get rid of this whole mess.
+type SomeMergeHandler (k :: S -> Type) (v :: S -> Type) (s :: S) =
```
SeungheonOh, line 390: Yes please! #642

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -0,0 +1,1275 @@
...
+{- | Hash a script, appending the Plutus V2 prefix.
+
+@since 2.0.0
+-}
+scriptHash :: Script -> Plutus.ScriptHash
+scriptHash = hashScriptWithPrefix "\x02"
+
+-- | @since 2.0.0
+data PDCert (s :: S)
+  = PDCertDelegRegKey (Term s (PDataRecord '["_0" ':= PStakingCredential]))
```
SeungheonOh, line 486: Let's add documentation on each of these saying what each one does. This PR doesn't have to take care of it.

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -0,0 +1,1275 @@
...
+deriving via
+  (DerivePConstantViaData Plutus.TxInInfo PTxInInfo)
+  instance
+    PConstantDecl Plutus.TxInInfo
+
+-- | @since 2.0.0
+data POutputDatum (s :: S)
+  = PNoOutputDatum (Term s (PDataRecord '[]))
+  | POutputDatumHash (Term s (PDataRecord '["datumHash" ':= PDatumHash]))
+  | -- | Inline datum as per
```
SeungheonOh, line 345: Don't need this documentation about CIP. Rather, it needs to explain _**what inline-datum is**_. 

This is not under the scope of this PR.

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -0,0 +1,1275 @@
...
+  -- * Script
+
+  -- ** Types
+  PDatum (..),
+  PDatumHash (..),
+  PRedeemer (..),
+  PRedeemerHash (..),
+  PScriptHash (..),
+
+  -- ** Functions
```
SeungheonOh, line 34: There's two "Functions" sections. This needs to be cleaned later once everything lands here

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -0,0 +1,1275 @@
...
+    PConstantDecl Plutus.DatumHash
+
+-- | @since 2.0.0
+newtype PRedeemerHash (s :: S) = PRedeemerHash (Term s PByteString)
+
+{- | Hash a script, appending the Plutus V2 prefix.
+
+@since 2.0.0
+-}
+scriptHash :: Script -> Plutus.ScriptHash
```
SeungheonOh, line 481: Need opinion: Do we want to keep functions near their relevant types(e.g. hash functions near hash API data types) or do we want to keep these helper function in its own function away from all data type definitions?

Personally, I'd prefer latter; it's easier to find when it's like that

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -0,0 +1,1275 @@
...
+    unwrapped <- tcont . plet $ ptryFrom @(PAsData PByteString) opq snd
+    tcont $ \f ->
+      pif
+        (plengthBS # unwrapped #== 28)
+        (f ())
+        (ptraceError "ptryFrom(PPubKeyHash): must be 28 bytes long")
+    pure (punsafeCoerce opq, pcon . PPubKeyHash $ unwrapped)
+
+-- | @since 2.0.0
+newtype PubKey = PubKey
```
SeungheonOh, line 778: Why do we need this here? Feels a bit random and unrelated to Plutarch

plutarch-api/src/Plutarch/Api.hs
kozross: They're meant to be per-section.

plutarch-api/src/Plutarch/Api.hs
kozross: If a function is publically-exported, I like to keep it near its type(s), but this is sometimes hard to make good decisions around. For 'hidden helpers' which don't get exported, I agree, they should be grouped together.

plutarch-api/src/Plutarch/Api.hs
kozross: This is a direct port from how it was defined. It's because it acts as the Haskell-level equivalent to `PPubKey` I think.

plutarch-api/src/Plutarch/Api.hs
SeungheonOh: If function's type includes type that is defined, let's keep it near the type. If it doesn't have the exact type, let's keep it at the bottom. This means we move functions that is exported but not directly related to any of defined type to the bottom

SeungheonOh (approved):
LGTM

plutarch-api/src/Plutarch/Api/Utils.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE PolyKinds #-}
```
L-as, line 1: PolyKinds is included in GHC2021 FYI

PR #648, opened 2024-01-21
Port -extra into API

### LINKED ISSUE #641: Migrate API-related utilities from `plutarch-extra` into ledger version specific library(#640)
Can be done after, or concurrently with #640. 

Move extra utilities into `plutarch-ledger-api-v...` as currently some utilities from `plutrach-extra` is ledger-verison-specific, and currently only support V1 ledger API

kozross:
Closes #641 .

### DIFF
--- a/plutarch-api/src/Plutarch/Api.hs
+++ b/plutarch-api/src/Plutarch/Api.hs
@@ -1,0 +1,1275 @@
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{- | = Note
+
+The 'Value.PValue' and 'AssocMap.PMap'-related functionality can be found in
+other modules, as these clash with the Plutarch prelude. These should be
+imported qualified.
+-}
+module Plutarch.Api (
+  -- * Contexts
+  PScriptContext (..),
+  PTxInfo (..),
+  PScriptPurpose (..),
+
+  -- * Tx
+  PTxOutRef (..),
+  PTxOut (..),
+  PTxId (..),
+  PTxInInfo (..),
+  POutputDatum (..),
+
+  -- * Script
+
+  -- ** Types
+  PDatum (..),
+  PDatumHash (..),
+  PRedeemer (..),
+  PRedeemerHash (..),
+  PScriptHash (..),
+
+  -- ** Functions
+  scriptHash,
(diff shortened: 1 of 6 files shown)


SeungheonOh (dismissed):
Everything looks good to me as most of it is copy from old code. We can merge once we dealt with `PTuple`

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -1195,30 +1199,124 @@ dataHash = hashData . Plutus.toData
...
+    PNothing -> pcon PNothing
+    PJust datumTuple -> pcon . PJust . ptryFromData . pforgetData $ psndBuiltin # pfromData datumTuple
+  where
+    matches ::
+      forall (s' :: S).
+      Term s' (PDatumHash :--> PAsData (PBuiltinPair (PAsData PDatumHash) (PAsData PDatum)) :--> PBool)
+    matches = phoistAcyclic $ plam $ \a ab -> a #== pfromData (pfstBuiltin # pfromData ab)
+
+{-
+pparseDatum = phoistAcyclic $
```
SeungheonOh, line 1304: Clean this

plutarch-api/src/Plutarch/Api.hs
```diff
@@ -1273,3 +1371,10 @@ pmaybeLT whenBothNothing ltF = phoistAcyclic $
             $ pconstant whenBothNothing
         )
       $ pconstant True
+
+ptryFromData ::
+  forall (a :: S -> Type) (s :: S).
+  PTryFrom PData (PAsData a) =>
+  Term s PData ->
+  Term s (PAsData a)
+ptryFromData x = unTermCont $ fst <$> tcont (ptryFrom @(PAsData a) x)
```
SeungheonOh, line 1380: I don't think we need TermCont here

kozross: It's a direct copy. I'll rewrite it.

SeungheonOh: I don't think we need this helper really

SeungheonOh:
Also, changed need to be logged now. I thought it was meant to overtake 640, but now that it's independent PR

kozross:
> Also, changes need to be logged now. I thought it was meant to overtake 640, but now that it's independent PR

Happy to add those, but that means a version bump.

SeungheonOh:
Yes, we need version bumped since we just deprecated a type

SeungheonOh (approved):
LGTM

PR #651, opened 2024-01-23
Move TermCont utilities into `Plutarch` from plutarch-extra

SeungheonOh:


kozross (dismissed):

kozross (approved):

L-as:
TermCont has a foot gun with `pure` FYI. That's why `QualifiedDo` is preferred.

PR #652, opened 2024-01-24
Flesh out Plutarch.Api, port utilities

### LINKED ISSUE #636: Tests for V2 API spec
Currently, our tests only cover [V1](https://github.com/Plutonomicon/plutarch-plutus/blob/master/plutarch-test/tests/Plutarch/ApiSpec.hs). We need another such test for V2 (and V3 eventually as well).

kozross:
Closes #636. This is a fairly major set of changes: a lot of identifiers have moved, some have vanished altogether, and a few bits still remain outstanding. Every change has been logged, but a few are a bit more 'invisible'. 

# Guide to migration

Essentially, a lot of stuff that used to be in `plutarch-extra` is now part of `plutarch-api`. Specifically, `Plutarch.Extra.Map`, `Plutarch.Extra.Interval`, `Plutarch.Extra.Api` have been wholesale ported into `plutarch-api` (either a submodule, or `Plutarch.Api` itself), and any part of `Plutarch.Extra.Maybe` to do with `PMaybeData` is now in `Plutarch.Api` as well. Furthermore, `Plutarch.Api.*` from the main Plutarch repo is now gone: everything that was there was moved to `plutarch-api`.

Most notable change is that now, the API is V2 _only_. V1 stuff is gone, and all data types and functionality that differed between V1 and V2 now defaults to V2. This is a temporary measure, as we're moving to V3 in short order. This may mean some data types have changed a bit: if your old data types or functions were V1, you need to migrate to their V2 versions.

A few things have been removed altogether, as they either have no V1 equivalents, weren't useful (or covered by tests), or are otherwise no longer required:

* `pnever` (no V2 equivalent)
* `PTuple` (`type` synonym, `PTuple a b` is identical to `PBuiltinPair (PAsData a) (PAsData b)`)
* All the functionality around different `MergeHandler`s (hidden for now unless someone wants it, pending a full rethink)
* `ptuple`, `ptupleFromBuiltin`, `pbuiltinPairFromTuple` (no-ops anyway, now useless with `PTuple` being gone)
* `PValidator`, `PMintingPolicy`, `PStakingValidator` (all `type`s, just use their equivalents, which are `PData :--> PData :--> PScriptContext :--> POpaque`, `PData :--> PScriptContext :--> POpaque` and `PData :--> PScriptContext :--> POpaque` respectively)
* `IsList` instances for `PMap` (were broken and a bad idea to begin with, use `punsortedMapFromFoldable` or `psortedMapFromFoldable` instead of `fromList`)

It is possible some functionality for `PMap` or `PValue` slipped by us: if you need it, please let us know and we'll do our best to restore it!

# Outstanding tasks

In light of this PR landing, there are some remaining things to do:

1. Some of our tests are currently stubbed out, as they don't fit within V1 requirements, or are somehow broken more fundamentally. These need restoring, either with or without more broad test revisions.
2. Some functionality (particularly `PMap` functionality that used to be in `plutarch-extra`) has no test coverage. This should be dealt with, either with or without more broad test revisions.
3. V3 API support needs to be added. This is a key requirement for CIP-1694.
4. The rest of `plutarch-extra` needs to be subsumed into Plutarch itself, and the entire sublibrary needs to be eliminated.
5. Documentation for many of the identifiers in the new `plutarch-api` is quite lacklustre, and needs improvement. This is particularly important for stuff coming from the Plutus side (such as data types), as this is a _known_ confusion point. Examples of its use would also help drastically.
6. _All_ uses of `type` must be killed with fire. While efforts have been made to make `plutarch-api` free of that terrible idea, the rest of Plutarch still has rather a lot of it.


SeungheonOh:
Note about deprecation of `PTuple`: 
`PTuple a b` is identical to `PAsData (PBuiltinPair (PAsData a) (PAsData b))` and therefore `PBuiltinPair` will entirely replace `PTuple`. Utility functions of `PBuiltinPair` will take over the removed utility functions of `PTuple`. 

Note about deprecation of `IsList` instances:
`IsList` instances of various type have been removed due to their incompleteness in definition. The reason being, in order to lift Plutarch Term into Haskell value, one needs to evaluate the underlying value a term holds. As evaluation is not pure, Plutarch types were unable to implement the `toList` instance of `IsList` type class, and they were left to throw error all the time.  

Note about deprecation of V1 API:
V1 plutus-ledger-api is no longer supported as of this PR. It is because there's no reason for users to develop and maintain script on older ledger API. V2 is strict superset of V1(and V3 is likely to be strict superset of V2) and therefore, nothing incentivizes using V1 ledger API over V2. 

For existing codebase that is using V1(and V2, after V3 lands), they would need to bump their codebase, adapting to new api library. Since smart contract scripts are rather static objects, meaning they are hard to change easily nor frequently, discontinuation of old ledger API on Plutarch will not break existing and deployed onchain scripts. When replacing a old script, ledger version of the new script does not need to match that of old script; therefore, new ledger api can be used to build replacement script.

bladyjoker:
Woohooo staging!

bladyjoker (dismissed):
Thanks folks!

SeungheonOh (approved):
There are very few instances of commented out ledger api test cases, but we will eventually get better and separated test cases for ledger repo itself. 

LGTM

PR #669, opened 2024-05-04
Update Hashing.hs

peter-mlabs:


SeungheonOh (approved):
Either is fine, but maybe we want to expose this to haddock as well? So `--|` not `--`

peter-mlabs:
Oh, this isn't replacing the comment. Just adding some lines.

SeungheonOh:
Oh really?? Have I been using Haddock incorrectly since the beginning... I'll test it 

But tbh having | on every haddock comment line looks better.

SeungheonOh (approved):
Thanks!

PR #671, opened 2024-05-26
Add `evalTermNoEmit` and internal functions used by this function

SeungheonOh:
This is for @peter-mlabs to test if there are any notable performance difference between evaluating CEK with budgeting and trace logging vs. with everything turned off.

There will be some performance draw back for large scripts because is has to convert DeBruijn terms into named terms for evaluation and convert named terms back to DeBruijn to be re-integrated into Plutarch term.


peter-mlabs:
Haven't actually had the time to benchmark this, but I've found another function and now I'm wondering what the difference is :sob: https://cardano-ledger.cardano.intersectmbo.org/cardano-ledger-core/Cardano-Ledger-Plutus-Language.html#v:evaluatePlutusRunnable

PR #672, opened 2024-05-27
Log levels

kozross:
This adds the ability to specify a log level as well as a tracing mode, both at compile time for scripts, and also when injecting traces into generated code. By doing this, we can log 'different amounts' at the same time (for example, an error code at info level, but a much bigger message at debug level).

Currently, we don't test this new functionality: to ensure CI passes, deprecation warnings have been disabled for the relevant test module, but this needs to be rewritten if this functionality turns out to be beneficial.


Plutarch/Internal/Trace.hs
```diff
@@ -1,29 +1,73 @@
...
+  forall (a :: S -> Type) (s :: S).
+  Term s PString ->
+  Term s a ->
+  Term s a
+ptraceDebug msg x = pgetConfig $ \case
+  Tracing LogDebug _ -> pforce $ ptrace' # msg # pdelay x
+  _ -> x
+
+{-
 -- | Trace the given message before evaluating the argument.
```
SeungheonOh, line 69: We can remove this entirely

SeungheonOh (commented):
Looks good to me. I'm wondering, though, currently we only have two log levels--Debug and Info. Will we ever need extra levels of logging in the future?

I think it's no.

Plutarch/Internal/Trace.hs
SeungheonOh: nvm, we need to add this back with deprecation warning

kozross:
@SeungheonOh - I don't think so either, but @peter-mlabs is the best judge of this.

peter-mlabs:
Can we also add `pretty` and `eq` instances for `Config` here?

kozross:
`Eq` I can add, but `Pretty` isn't something we use very much in general in Plutarch.

Plutarch/Internal/Trace.hs
kozross: What do you mean by this?

SeungheonOh: old `ptrace` function is commented out at the end of the file

SeungheonOh (approved):
LGTM thanks

SeungheonOh (approved):

PR #677, opened 2024-06-03
Bump plutus 1.23.0.0

albertodvp:
Bumps plutus-core, depends on #672


kozross (approved):

SeungheonOh:
We should've bumped the major version since it is a dependency update

PR #679, opened 2024-06-13
V3 data types

### LINKED ISSUE #655: Move plutarch-api to V3
Follow-on for point 3 of #652 .

kozross:
This fixes #655. However, a few outstanding tasks remain:

1. Tests need to be written for all the added and changed types. This is part of our wider move to improve our test suite.
2. There are some functions in the V3 `Contexts` module on the Plutus side that we probably want Plutarch equivalents of, as some supercede existing functionality. These have been started, but not finished.
3. The documentation is now stale in places. I've fixed it so it compiles, but some of the prose no longer makes sense in light of these changes.

For my own sanity, I did some structural mirroring of the organization of the Plutus ledger API modules. We don't have to keep this if we don't want to.


plutarch-extra/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## 1.4.0 -- 07-06-2024
```
SeungheonOh, line 7: Last PR made it 1.7.0. So it should be 1.8.0?

kozross: Yes, it should be. Not sure what happened there.

SeungheonOh (approved):

PR #684, opened 2024-06-19
Orphans for V2

### LINKED ISSUE #681: Package for QuickCheck orphans
While we wait for [this issue](https://github.com/IntersectMBO/plutus/issues/6210), we can take the already existing orphans defined for various Plutus ledger types from [here](https://github.com/input-output-hk/xsy-liqwid-libs/blob/main/plutarch-quickcheck/src/Plutarch/Test/QuickCheck/Instances.hs) and make a sublibrary for them.

Some organizational questions in light of #680 are definitely a concern. My current plan is to make a `V1.Orphans`, `V2.Orphans` and `V3.Orphans`, mirroring the `plutus-ledger-api` data type exports from corresponding modules. We can, of course, share instances on the back end.

kozross:
Closes #681. This will need to be updated in PR #679 for V3 versions of the types in question.


plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+
+-- TODO: We should probably treat Ada values differently?
+
+{- | This is the most general possible instance for 'PLA.Value'. In particular,
+this can have zero values, and does not treat the Ada symbol or token name
+specially.
+
+@since 1.0.0
+-}
+instance Arbitrary PLA.Value where
```
SeungheonOh, line 423: Ideally, we would differentiate different types of `Value` as `PValue` does. 

Arbitrary `Value` coming from ledger is ensured to have all positive entries and sorted. However, `Value` can very well be constructed without such constraints. It would be useful to be able to generate and tests on such different kinds of `Value`s. 

For this PR, I don't really think we need additional types. I suggest making an issue for this and dealing with it on other PR

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+-- | @since 1.0.0
+deriving via PlutusTx.BuiltinData instance CoArbitrary PLA.Datum
+
+-- | @since 1.0.0
+instance Function PLA.Datum where
+  {-# INLINEABLE function #-}
+  function = functionMap coerce PLA.Datum
+
+-- Note: This works because TxId and DatumHash are the same hash type, with the
+-- same length.
```
SeungheonOh, line 620: I think it'd be better to have definition explicitly here. For the sake of clarity.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+                      PLA.StakingCredential
+                      ( Either
+                          (PLA.StakingCredential, PLA.PubKeyHash)
+                          ( Either (PLA.PubKeyHash, PLA.PubKeyHash) (PLA.PubKeyHash, Integer)
+                          )
+                      )
+                  )
+              )
+          )
+      into = \case
```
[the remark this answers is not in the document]
SeungheonOh: ~~Use `NS` :rofl:~~

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+
+{- | We provide an instance specialized to 'PLA.POSIXTime', rather than a more
+general one, as it doesn't make much sense to talk about 'PLA.Interval's of
+arbitrary types in general. Furthermore, this is the only instance we
+actually use, so there's no real loss there.
+
+This instance tries to make time intervals as fairly as possible, while also
+ensuring that they're sensibly formed. We work under the assumption of a
+32-bit epoch: while this is _technically_ not going to last much longer,
+we're safe until about 2030 on that basis, which should be enough for now.
```
SeungheonOh, line 858: Plutarch 2030 ??

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+instance Function PLA.RedeemerHash where
+  {-# INLINEABLE function #-}
+  function = functionMap coerce PLA.RedeemerHash
+
+{- | This is a very general instance, able to produce 'PLC.Data' of basically
+any shape. You probably want something more focused than this.
+
+@since 1.0.0
+-}
+instance Arbitrary PLC.Data where
```
SeungheonOh, line 1168: Any reason we have instances for both `Data` and `BuiltinData`

`BuiltinData` is just wrapper around `Data`

SeungheonOh (approved):
LGTM, some questions only

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: Agreed, this is a bit of a kludge. I've left a note specifying all this in case someone finds this sharp edge, but in the long run, we definitely want to make better instances.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: Sure, it is a bit cryptic as it stands.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: Not exactly. `Data` doesn't exist onchain, while `BuiltinData` does, so having both does make some sense. It's minor, but since the two instances are basically identical, it doesn't really hurt too much to have both.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+{- | Base-16 encoded bytestring.
+
+@since 1.0.0
+-}
+instance Arbitrary PLA.LedgerBytes where
+  {-# INLINEABLE arbitrary #-}
+  arbitrary = PLA.LedgerBytes . PlutusTx.toBuiltin @ByteString . Base16.encode <$> arbitrary
+  {-# INLINEABLE shrink #-}
+  -- We use the lenient strategy here, as it's not possible for us to ever end
+  -- up with anything that wasn't base-16 to begin with.
```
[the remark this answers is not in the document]
bladyjoker: https://github.com/IntersectMBO/plutus/blob/7dbbd7fe4e4ffd3393e6c5644a14fb610d92d524/plutus-ledger-api/src/PlutusLedgerApi/V1/Bytes.hs#L93

Basically you can ignore hex encoding anywhere, it's just bytes

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+-- | @since 1.0.0
+instance Function PLA.PubKeyHash where
+  {-# INLINEABLE function #-}
+  function = functionMap coerce PLA.PubKeyHash
+
+{- | BLAKE2b-244 hash. This does not shrink.
+
+@since 1.0.0
+-}
+deriving via PLA.PubKeyHash instance Arbitrary PLA.ScriptHash
```
bladyjoker, line 114: Consider having a Blake224Hash type and derive instances from there for PublicKeyHash and ScriptHash?

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+@since 1.0.0
+-}
+instance Arbitrary PLA.CurrencySymbol where
+  {-# INLINEABLE arbitrary #-}
+  arbitrary =
+    PLA.CurrencySymbol . PlutusTx.toBuiltin @ByteString
+      <$> frequency
+        [ (1, pure "")
+        , (maxBound, unSizedByteString @28 <$> arbitrary)
+        ]
```
bladyjoker, line 328: Consider using arbitrary of ScriptHash here?  Or BLake224Hash type?

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+limited to generating printable ASCII names, rather than the full UTF-8
+range. See 'PlutusLedgerApi.QuickCheck.Utils.NonAdaTokenName' documentation
+for an explanation of why we chose to do this. We also shrink into the Ada
+token name: bear this in mind.
+
+If you don't want to generate (or shrink into) the Ada token name, consider
+using 'PlutusLedgerApi.QuickCheck.Utils.NonAdaTokenName' instead.
+
+@since 1.0.0
+-}
```
bladyjoker, line 348: Ok, so you're choosing to have ascii token names? if so that's perfectly fine, just want to make sure it's known that TokenNames are also plain bytes.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+  function = functionMap AssocMap.toList AssocMap.unsafeFromList
+
+-- TODO: We should probably treat Ada values differently?
+
+{- | This is the most general possible instance for 'PLA.Value'. In particular,
+this can have zero values, and does not treat the Ada symbol or token name
+specially.
+
+@since 1.0.0
+-}
```
bladyjoker, line 422: This is where it gets tricky. UTxOValue must always have positive ADA, whereas MintValue must always have a 0 ADA.

So depends in which context you're generating for this might not be correct.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+      outOf (cred, scred) = PLA.Address cred scred
+
+{- | As this type is a hash, we don't shrink it.
+
+@since 1.0.0
+-}
+instance Arbitrary PLA.TxId where
+  {-# INLINEABLE arbitrary #-}
+  arbitrary =
+    PLA.TxId . PlutusTx.toBuiltin @ByteString . unSizedByteString @32 <$> arbitrary
```
bladyjoker, line 482: Document

```
{- | A transaction ID, i.e. the hash of a transaction. Hashed with BLAKE2b-256. 32 byte.
```

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
bladyjoker: Consider having a Blake2b256Hash from which these instances derive? Perhaps that clarifies things better

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
```diff
@@ -0,0 +1,1271 @@
...
+      into (PLA.TxOutRef txi ix) = (txi, ix)
+      outOf :: (PLA.TxId, Integer) -> PLA.TxOutRef
+      outOf (txi, ix) = PLA.TxOutRef txi ix
+
+-- Needed because otherwise, deriving an instance for TxInfo is impossible.
+-- Arguably this is actually a bug on IOG's side, as the corresponding Cardano
+-- type _does_ have an Ord instance.
+
+-- | @since 1.0.0
+deriving stock instance Ord PLA.ScriptPurpose
```
bladyjoker, line 674: https://github.com/IntersectMBO/plutus/blob/7dbbd7fe4e4ffd3393e6c5644a14fb610d92d524/plutus-ledger-api/src/PlutusLedgerApi/V2/Contexts.hs#L83

I just want y'all to appreciate this. How do they even construct the redeemers without the Ord?

bladyjoker (dismissed):
I'm not sure what happens when you generate some TxInfo fields

```
data TxInfo = TxInfo
    { txInfoInputs          :: [TxInInfo] -- ^ NON EMPTY Map TxOutRef TxInInfo (NOT A LIST)
    , txInfoReferenceInputs :: [TxInInfo] -- ^ Same as above
    , txInfoOutputs         :: [TxOut] -- ^ Anything goes but Value must have positive ADA
    , txInfoFee             :: Value -- ^ This MUST be ONLY ADA
    , txInfoMint            :: Value -- ^ This contains 0 ADA
    , txInfoDCert           :: [DCert] -- ^ Not sure but I suspect this is also a Map of sorts
    , txInfoWdrl            :: Map StakingCredential Integer
    , txInfoValidRange      :: POSIXTimeRange -- ^ End MUST be after begginning (shock and disbelief).
    , txInfoSignatories     :: [PubKeyHash] -- ^ This is NOT a LIST this is a Set PubKeyHash
    , txInfoRedeemers       :: Map ScriptPurpose Redeemer
    , txInfoData            :: Map DatumHash Datum
    , txInfoId              :: TxId
    } 
```

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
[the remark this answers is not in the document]
kozross: Good to know! That'll simplify the instance considerably.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: I felt that for testing purposes, it's better to generate sensible(ish) names. UTF-8 with a bounded byte length is hard to generate, but printable ASCII is easy, so that's what I went with.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: A mystery for the ages.

plutarch-orphanage/src/PlutusLedgerApi/V2/Orphans.hs
kozross: Given this, it suggests we want two `newtype` wrappers: `UTxOValue` and `MintValue`, both of which can be 'unwrapped' to make a `Value`. I'm a bit hesitant to define 'open' modifiers, but in this case, I think it's OK.

SeungheonOh (approved):

bladyjoker (approved):
TxInfo list/map/set invariants are missing right? You want to add these in some other PR?

PR #690, opened 2024-07-01
Restore V1 and V2 APIs

### LINKED ISSUE #680: Adding V1 and V2 types
As per @bladyjoker 's note, I found out some valid concerns about deprecating legacy ledger api types: old ledger types can used as datum to some data-providing scripts like oracles. In these cases, user might need to write their own types representing older ledger types.

To address, we need to add back V1 and V2 scripts to `plutrach-ledger-api`.

kozross:
Closes #680. The functions haven't been ported, but this was already the case before, now we're just more honest about it.


bladyjoker (approved):
Thank you this is great!

SeungheonOh (approved):
LGTM!

PR #693, opened 2024-07-03
PTryFrom tests for V3 types

### LINKED ISSUE #691: `PTryFrom` tests for V3 API types
These should be added to `plutarch-ledger-api`'s tests as laws.

kozross:
Closes #691.


SeungheonOh (dismissed):
LGTM, we should add V1 and V2 PTryFrom test as well. I made a separate ticket for it #694

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Contexts.hs
```diff
@@ -0,0 +1,59 @@
...
+import Plutarch.LedgerApi.Value qualified as Value
+import Plutarch.Lift (
+  PConstantDecl,
+  PUnsafeLiftDecl (PLifted),
+ )
+import Plutarch.Prelude
+import PlutusLedgerApi.V1 qualified as Plutus
+
+-- | @since 3.1.1
+data PScriptPurpose (s :: S)
```
SeungheonOh, line 22: Any particular reason why we moved this to its own module? 

I think having own module for this is better convention, but just wondering if there's other reason

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Contexts.hs
```diff
@@ -0,0 +1,59 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
```
SeungheonOh, line 1: I don't see any orphan here

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Contexts.hs
SeungheonOh: Ah nvm, V2 has to import this. Makes senes

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Tx.hs
```diff
@@ -0,0 +1,134 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
```
SeungheonOh, line 1: I don't see orphans here as well

SeungheonOh (approved):
LGTM, some questions

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Contexts.hs
kozross: Any `PConstantDecl` instance is an orphan, and we create an instance of it for `ScriptContext`.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Tx.hs
kozross: Similar here.

PR #699, opened 2024-07-10
applyArguments and doc fixes

kozross:
This PR does two things:

* Add `applyArguments` to `Plutarch.Evaluate`; and
* Fix the mistaken docs reference to this function (which IOG no longer provide)


SeungheonOh (approved):
LGTM

PR #700, opened 2024-07-10
PPubKeyHash documentation fixes

kozross:
This fixes incorrect documentation around `PPubKeyHash`.

### DIFF
--- a/plutarch-docs/Tricks/Representation of Plutarch type.md
+++ b/plutarch-docs/Tricks/Representation of Plutarch type.md
@@ -21,5 +21,5 @@ Now, let's discuss patterns of data declarations and what representation they _s
 - If it's a representationally equal wrapper (think Haskell `newtype`) to a term containing a Plutarch type - it should have the 
   same representation as that underlying Plutarch type.
 
-  e.g. `newtype PPubKeyHash (s :: S) = PPubKeyHash (Term s (PDataNewtype PByteString))` is just represented as `PByteString`. This is ensured 
+  e.g. `newtype PPositive (s :: S) = PPositive (Term s PInteger)` is just represented as `PInteger`. This is ensured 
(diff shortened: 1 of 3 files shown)


plutarch-docs/Tricks/Representation of Plutarch type.md
```diff
@@ -21,5 +21,5 @@ Now, let's discuss patterns of data declarations and what representation they _s
 - If it's a representationally equal wrapper (think Haskell `newtype`) to a term containing a Plutarch type - it should have the 
   same representation as that underlying Plutarch type.
 
-  e.g. `newtype PPubKeyHash (s :: S) = PPubKeyHash (Term s PByteString)` is just represented as `PByteString`. This is ensured 
+  e.g. `newtype PPubKeyHash (s :: S) = PPubKeyHash (Term s (PDataNewtype PByteString))` is just represented as `PByteString`. This is ensured
```
SeungheonOh, line 24: PPubkeyHash is not represented as `PByteString` as it is wrapped by `PDataNewtype`. 

It is better to give some new type 
like `newtype PMyType (s :: S) = PMyType (Term s PByteString)` which will actually have PByteString representation

plutarch-docs/Typeclasses/PConstant and PLift.md
```diff
@@ -91,7 +91,7 @@ This comes in three flavors:
 - Plutarch type represented **indirectly** by a builtin type that **is not** `Data` (`DefaultUniData`)
   ==> `DerivePConstantViaNewtype`
 
-  *Ex:* `PPubKeyHash` is a newtype to a `PByteString`, and `PByteString` is _directly_ represented as a
+  *Ex:* `PPubKeyHash` is a newtype to a `PDataNewtype PByteString`, and `PByteString` is _directly_ represented as a
```
SeungheonOh, line 94: Same thing here, I think using PByteString is no longer too straightforward due to `PDataNewtype`

SeungheonOh (changes requested):

PR #701, opened 2024-07-10
Explain trace levels

kozross:
This adds some documentation explaining trace levels and what they mean.

### DIFF
--- a/plutarch-docs/Usage/Tracing.md
+++ b/plutarch-docs/Usage/Tracing.md
@@ -31,7 +31,7 @@ in development mode, these functions basically do nothing.
 
 # Important note
 
-Use of `PShow` is strongly discouraged in any tracing code (and any code in
-general). This uses a lot of resources onchain, and can easily exhaust script
+Use of `PShow` is strongly discouraged in any production on-chain scripts. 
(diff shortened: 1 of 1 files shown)


plutarch-docs/Usage/Tracing.md
```diff
@@ -12,8 +12,26 @@ import Plutarch.Prelude ()
...
+certain conditions. See those modules, and their documentation, for more
+details.
+
+If you have the `development` flag for `plutarch` turned on - you'll see the 
+trace messages appear in the trace log during script evaluation. When not 
+in development mode, these functions basically do nothing.
+
+# Important note
+
+Use of `PShow` is strongly discouraged in any tracing code (and any code in
```
L-as, line 34: But you wouldn't compile this when going on-chain?

kozross: True. This might not be the best place for this warning.

SeungheonOh: I think this note is in lightof https://github.com/Plutonomicon/plutarch-plutus/issues/697

If we correctly handle pruning of pshow according to the trace config, we don't need warning

SeungheonOh (approved):

PR #703, opened 2024-07-10
Fix PShow not to emit show code when NoTracing is on

### LINKED ISSUE #697: Tracing functions leave some computation even when compiled with `NoTrace` option
Reported by @bladyjoker , we want tracing function to erase all computation related to that trace message, i.e. `pshow`, when we are compiling with `NoTrace` option. This seems to be not the case currently. Need to investigate and fix if it is not working as we desire.

kozross:
Fixes #697 .

### DIFF
--- a/Plutarch/Show.hs
+++ b/Plutarch/Show.hs
@@ -31,8 +31,10 @@ import Plutarch.Bool (PBool, PEq, pif, pif', (#<), (#==))
 import Plutarch.ByteString (PByteString, pconsBS, pindexBS, plengthBS, psliceBS)
 import Plutarch.Integer (PInteger, PIntegral (pquot, prem))
 import Plutarch.Internal (
+  Config (NoTracing),
   Term,
   perror,
+  pgetConfig,
   phoistAcyclic,
   plet,
   punsafeCoerce,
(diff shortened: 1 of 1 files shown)


Plutarch/Show.hs
```diff
@@ -58,9 +60,19 @@ class PShow t where
...
+Use this in preference to @pshow'@ whenever possible. We use some additional
+logic to avoid compiling (potentially very expensive) 'PShow' code into
+traces when tracing is disabled in this function, which will not be used if
+you write code using @pshow'@ instead.
+-}
 pshow :: PShow a => Term s a -> Term s PString
-pshow = pshow' False
+pshow t = pgetConfig $ \case
+  NoTracing -> pconstant ""
+  _ -> pshow' False t
```
SeungheonOh, line 75: I don't think this is a correct solution. This will break other valid use cases for pshow. 

Rather, I think we should try to fix it on trace functions to correctly prune branches on certain config

kozross: Yeah, upon closer inspection I agree - just because tracing isn't on doesn't mean we want to block all `PShow` use. However, the kind of pruning you describe seems rather hard to do.

L-as: Why would you ever want to show on-chain? What are the valid use cases?

SeungheonOh:
irrelevant

Issue #705, opened 2024-07-13
Bug in `pcontains`


peter-mlabs:
It appears that `pcontains` is not working properly.

```hs
ghci> i1
Interval {ivFrom = LowerBound (Finite (POSIXTime {getPOSIXTime = 120})) True, ivTo = UpperBound (Finite (POSIXTime {getPOSIXTime = 129})) False}
ghci> i2
Interval {ivFrom = LowerBound (Finite (POSIXTime {getPOSIXTime = 120})) True, ivTo = UpperBound (Finite (POSIXTime {getPOSIXTime = 129})) True}
ghci> contains i2 i1
True
ghci> plift $ pcontains # pconstant i2 # pconstant i1
False
```

As far as I can tell, it works when the upper bound for both are closed, but not when one is open. We should have the $[120, 129) \subseteq [120, 129]$, but we don't.

peter-mlabs:
Note that this is particularly an issue because of [this](https://github.com/Plutonomicon/cardano-transaction-lib/issues/1041#issuecomment-1280868255). 

Also, we should probably indicate that "gotcha" in the Interval module and haddock comments.

L-as:
Following that comment, if it isn't supported in Plutus, why support it in Plutarch at all?

peter-mlabs:
@L-as Its support in `plutus-ledger-api`, so I think it does make sense for compatibility reasons. 

But I still think that we should have a separate library that actually does what PLA "means" and not what it "does"

L-as:
> `plutus-ledger-api`

Fair reason, but arguably even then you should never use it. You could mark it as deprecated but still fix it.

PR #706, opened 2024-07-15
Interval fixes

kozross:
Partially addresses #705:

* We clearly specify that only certain intervals are seen as acceptable, and provide `pcheckInterval` to ensure that this is preserved (or blow up).
* The corresponding generator from `plutarch-orphanage` has been updated
* `pinterval` has been deprecated, as it generates an invalid interval
* `pintervalOpenEnd` has been added as a replacement to minimize breakages


plutarch-ledger-api/src/Plutarch/LedgerApi/Interval.hs
```diff
@@ -1,7 +1,29 @@
+{-# LANGUAGE PolyKinds #-}
```
L-as, line 1: This smells wrong, isn’t this the default nowadays?

Issue #707, opened 2024-07-17
`FromJSON` instance missing for `Config`


jaredponn:
In the latest `staging` branch, we can see that we have the `ToJSON Config` instance implemented

https://github.com/Plutonomicon/plutarch-plutus/blob/staging/Plutarch/Internal.hs#L325-L345

but there is no `FromJSON Config` implemented.

Indeed, the `CHANGELOG.md` suggests the `FromJSON Config` instance was [implemented](https://github.com/Plutonomicon/plutarch-plutus/blob/346202aa0d25625bf8045a1556fa11835758c547/CHANGELOG.md?plain=1#L37), but this is evidently not the case.

I don't think I have push permission, but this patch should implement the instance (it's missing roundtrip tests though!):

```diff
diff --git a/Plutarch/Internal.hs b/Plutarch/Internal.hs
index 54eff22bd..7dfe6a600 100644
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -53,7 +53,9 @@ import Data.Aeson (
   object,
   pairs,
   withText,
+  withObject,
   (.=),
+  (.:),
  )
 import Data.ByteString qualified as BS
 import Data.Kind (Type)
@@ -344,6 +346,14 @@ instance ToJSON Config where
           <> ("logLevel" .= ll)
           <> ("tracingMode" .= tm)
 
+-- | @since 1.6.0
+instance FromJSON Config where
+  parseJSON = withObject "Config" $ \v ->
+    v .: "tag" >>= \(tag :: Int) -> case tag of
+        0 -> return NoTracing
+        1 -> Tracing <$> v .: "logLevel" <*> v .: "tracingMode"
+        _ -> fail "Invalid tag"
+
 {- | If the config indicates that we want to trace, get its mode.
 
 @since 1.6.0
````

PR #708, opened 2024-07-17
FromJSON instance missing for Config #707

jaredponn:
This PR:
- [x] Adds the `FromJSON Config` instance
- [ ] Roundtrip tests (probably will be left for another issue)

### DIFF
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -52,7 +52,9 @@ import Data.Aeson (
   ToJSON (toEncoding, toJSON),
   object,
   pairs,
+  withObject,
   withText,
+  (.:),
   (.=),
  )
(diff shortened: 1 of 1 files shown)


Plutarch/Internal.hs
```diff
@@ -344,6 +346,14 @@ instance ToJSON Config where
           <> ("logLevel" .= ll)
           <> ("tracingMode" .= tm)
 
+-- | @since 1.6.0
```
jaredponn, line 349: Not sure if this `@since` is right?

kozross: Does the changelog disagree? If so, it should match the changelog.

kozross (approved):

PR #709, opened 2024-07-18
Add trace message codifier

SeungheonOh:
Small utility function to turn long messages into code table.

### DIFF
--- a/plutarch-extra/Plutarch/Extra/ErrorCode.hs
+++ b/plutarch-extra/Plutarch/Extra/ErrorCode.hs
@@ -1,0 +1,71 @@
+module Plutarch.Extra.ErrorCode (errorTagReplacer, codifyStrings) where
+
+import Data.List (nub)
+import Data.Maybe (fromMaybe)
+import Data.Text (Text)
+import Data.Text qualified as T
+import Data.Tuple (swap)
+import Plutarch.Script (Script (Script))
+import PlutusCore (DefaultUni (..))
+import Universe (Some (Some), ValueOf (ValueOf))
+import UntypedPlutusCore qualified as UPLC
+
+-- | Replace any string that is tagged with "ERR:" prefix will get replaced by a unique integer id
+errorTagReplacer :: Int -> Text -> Maybe Text
+errorTagReplacer n s =
+  if T.isPrefixOf "ERR:" s
+    then Just $ "CODE:" <> T.pack (show n)
+    else Nothing
+
+-- | Replace strings satisfying the predicate with provided replacement
+codifyStrings :: (Int -> Text -> Maybe Text) -> Script -> (Script, [(Text, Text)])
+codifyStrings codify (Script (UPLC.Program ann version terms)) =
+  let
+    immap :: forall a b. (Int -> a -> Maybe b) -> [a] -> [(a, b)]
+    immap f xs = go 0 xs
+      where
+        go _ [] = []
+        go n (x : xs) =
+          case f n x of
+            Just x' -> (x, x') : go (n + 1) xs
+            Nothing -> go (n + 1) xs
+
+    messagesIds = immap codify $ nub $ collectErrorMessages terms
+   in
+    (Script (UPLC.Program ann version (replaceMessages messagesIds terms)), swap <$> messagesIds)
+  where
+    -- \| Scripts all Strings form the script
+    collectErrorMessages :: UPLC.Term UPLC.DeBruijn UPLC.DefaultUni UPLC.DefaultFun () -> [Text]
+    collectErrorMessages (UPLC.Var _ _) = []
+    collectErrorMessages (UPLC.LamAbs _ _ uterm) = collectErrorMessages uterm
+    collectErrorMessages (UPLC.Apply _ uterm1 uterm2) =
+      collectErrorMessages uterm1 <> collectErrorMessages uterm2
+    collectErrorMessages (UPLC.Force _ uterm) = collectErrorMessages uterm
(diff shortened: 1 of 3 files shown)


plutarch-extra/plutarch-extra.cabal
```diff
@@ -65,19 +66,23 @@ common c
     -Wincomplete-record-updates -Wmissing-deriving-strategies
     -Wno-name-shadowing -Wunused-foralls -fprint-explicit-foralls
     -fprint-explicit-kinds -Wno-unrecognised-warning-flags
+    -Wno-missing-export-lists
```
bladyjoker, line 69: Why?

plutarch-extra/Plutarch/Extra/ErrorCode.hs
```diff
@@ -0,0 +1,71 @@
...
+import Plutarch.Script (Script (Script))
+import PlutusCore (DefaultUni (..))
+import Universe (Some (Some), ValueOf (ValueOf))
+import UntypedPlutusCore qualified as UPLC
+
+-- | Replace any string that is tagged with "ERR:" prefix will get replaced by a unique integer id
+errorTagReplacer :: Int -> Text -> Maybe Text
+errorTagReplacer n s =
+  if T.isPrefixOf "ERR:" s
+    then Just $ "CODE:" <> T.pack (show n)
```
bladyjoker, line 17: Maybe also `ERR:1` rather than `CODE:1`?

plutarch-extra/Plutarch/Extra/ErrorCode.hs
```diff
@@ -0,0 +1,71 @@
...
+    else Nothing
+
+-- | Replace strings satisfying the predicate with provided replacement
+codifyStrings :: (Int -> Text -> Maybe Text) -> Script -> (Script, [(Text, Text)])
+codifyStrings codify (Script (UPLC.Program ann version terms)) =
```
bladyjoker, line 22: This should return a `Map Text Text` not a list?

plutarch-extra/Plutarch/Extra/ErrorCode.hs
```diff
@@ -0,0 +1,71 @@
...
+    collectErrorMessages (UPLC.Constant _ (Some (ValueOf DefaultUniString str))) = [str]
+    collectErrorMessages (UPLC.Constant _ _) = []
+    collectErrorMessages (UPLC.Builtin _ _) = []
+    collectErrorMessages (UPLC.Error _) = []
+    collectErrorMessages (UPLC.Constr _ _ uterms) = foldMap collectErrorMessages uterms
+    collectErrorMessages (UPLC.Case _ uterm uterms) =
+      collectErrorMessages uterm <> foldMap collectErrorMessages uterms
+
+    replaceMessages ::
+      [(Text, Text)] ->
```
bladyjoker, line 54: Also `Map Text Text`

bladyjoker (approved):
Thank you @SeungheonOh , I would really like to see some tests covering this so the users don't discover any issues themselves.

plutarch-extra/plutarch-extra.cabal
kozross: This is a weird leftover from earlier Plutarch days. Should be removed I think.

plutarch-extra/Plutarch/Extra/ErrorCode.hs
```diff
@@ -0,0 +1,71 @@
...
+  if T.isPrefixOf "ERR:" s
+    then Just $ "CODE:" <> T.pack (show n)
+    else Nothing
+
+-- | Replace strings satisfying the predicate with provided replacement
+codifyStrings :: (Int -> Text -> Maybe Text) -> Script -> (Script, [(Text, Text)])
+codifyStrings codify (Script (UPLC.Program ann version terms)) =
+  let
+    immap :: forall a b. (Int -> a -> Maybe b) -> [a] -> [(a, b)]
+    immap f xs = go 0 xs
```
kozross, line 25: Were you looking for [this](https://hackage.haskell.org/package/witherable-0.5/docs/Witherable.html#t:WitherableWithIndex) perhaps?

kozross (changes requested):
Basically I think @bladyjoker beat me to most of the review.

L-as:
My thoughts:
- I can't see why you want this TBF, when is this useful? You don't want strings on-chain anyway?
- If you did want error codes on-chain, wouldn't it make more sense to use a more compact representation that uses the full extent of UTF-8?
- Rather than doing this in a post-processing step, shouldn't `ptrace` etc. do this at generation time? You could add another alternative between the no tracing and full tracing options.

SeungheonOh:
This is per @bladyjoker 's request. I see this being very useful. IMO, not having any trace for mainnet script was a huge mistake, at least for Liqwid. Anytime script threw up, it was massive pain to recreate that with scripts with tracing to actually figure out what went wrong. Codifying error messages would save a lot of efforts without not much cost since it will be reference script anyways. 

As for UTF-8 codes, I think that is a good idea. I change it to support UTF-8 tags.

Also, for having this on Plutarch's compile time, this was made to be external tool so that it can be used on already existing scripts as well. Also, deep integration to Plutarch was too much effort; this is small and portable. Maybe if we can get more funding on Plutarch, we can tackle this.

SeungheonOh:
Not really, in case of Liqwid, we'd need to deploy extra sets of scripts which will cost few hundred Ada and make changes to offchain in some annoying way. So, rather, we have this utility that extracts CBOR and input UTXOs in their resolved form into JSON. We import that to Haskell project and run the CEK manually. It is less painful for sure, but still immensely annoying. Not to mention, this is very offchain framework dependent.

PR #712, opened 2024-10-01
Integrate plutarch-extra

### LINKED ISSUE #656: Subsume rest of plutarch-extra into Plutarch itself
Follow-on from point 4 of #652 .

kozross:
Closes #656. I have also moved all tests for the functionality that used to be in `plutarch-extra` to Plutarch itself, rather than `plutarch-test`. I'm saving reorganization of the test stanza and its code for another PR.


test/Main.hs
```diff
@@ -133,3 +196,53 @@ traces ::
...
   Right (_, _, logs) -> assertEqual "" expected logs
+
+isTrue ::
+  (forall (s :: S). Term s PBool) ->
+  IO ()
+isTrue comp = case evalTerm NoTracing comp of
+  Left err -> assertFailure $ "Did not compile: " <> Text.unpack err
+  Right (res, _, _) -> case res of
+    Left err -> assertFailure $ "Execution errored: " <> show err
+    Right t -> unless (plift t) (assertFailure "is false")
```
SeungheonOh, line 207: Instead of these, you can use `plift'` which gives you error directly

SeungheonOh: same thing for below

SeungheonOh (approved):
Looks good, mostly moving around + test

test/Main.hs
kozross: I need to throw an `AssertionError` specifically for `tasty-hunit` to work here.

Issue #704, opened 2024-07-11
`PPOSIXTime` must be able to store negative value


SeungheonOh:
`PPOSIXTime` should be able to store negative integer to store time before 1/1/1970. Currently, instances of `PPOSIXTime` forbids its value going negative

L-as:
Why do you need this?

SeungheonOh:
I don't know @peter-mlabs asked. 

Forbidding negative number for PPOSIXTime was over constraint to why not

kozross:
Are we sure negative times are allowed onchain as such?

SeungheonOh:
Yeah, it seems to work as expected. There's not special invariant on PlutusLedgerApi

L-as:
It's almost always invalid on-chain.

Nevertheless, I don't think there's any good reason to maintain this invariant in Plutarch.

PR #713, opened 2024-10-01
PCountable and PEnumerable

kozross:
This adds two type classes, providing successor and predecessor operations. This is groundwork for fixing #705 . We still need further work on fixing `PPosixTime` to support negative values as well (as per #704), but I'll send this in a separate PR.


Plutarch/Enum.hs
```diff
@@ -0,0 +1,120 @@
...
+import Plutarch.Internal.PLam (plam)
+import Plutarch.Positive (PPositive)
+
+{- | A notion of \'next\' value. More formally, instances of this type class are
+discrete linear orders with no maximal element.
+
+= Laws
+
+1. @x /= psuccessor x@
+2. @y < x@ @=@ @x >= psuccessor y@
```
SeungheonOh, line 28: @y<x@ @=@ @psuccessor y <= x@ reads better

Plutarch/Enum.hs
```diff
@@ -0,0 +1,120 @@
...
+import Plutarch.Internal.Other (pfix, pto)
+import Plutarch.Internal.PLam (plam)
+import Plutarch.Positive (PPositive)
+
+{- | A notion of \'next\' value. More formally, instances of this type class are
+discrete linear orders with no maximal element.
+
+= Laws
+
+1. @x /= psuccessor x@
```
L-as, line 27: What if there’s only one element

kozross: Then you can't be an instance.

Plutarch/Enum.hs
```diff
@@ -0,0 +1,120 @@
...
+  -- | @since WIP
+  {-# INLINEABLE psuccessorN #-}
+  psuccessorN :: forall (s :: S). Term s (PPositive :--> a :--> a)
+  psuccessorN = phoistAcyclic $ plam $ \n x -> go n # (psuccessor # x) # 1
+    where
+      go ::
+        forall (s' :: S).
+        Term s' PPositive ->
+        Term s' (a :--> PPositive :--> a)
+      go limit = pfix #$ plam $ \self acc count ->
```
SeungheonOh, line 56: This is getting recursed on script. Is this more favorable to having it generate script that just runs `psuccessor` n times? 

Latter will make script size grow linearly on the argument `n` but with smaller execution cost.

Plutarch/Enum.hs
```diff
@@ -0,0 +1,120 @@
...
+  -- | @since WIP
+  {-# INLINEABLE ppredecessorN #-}
+  ppredecessorN :: forall (s :: S). Term s (PPositive :--> a :--> a)
+  ppredecessorN = phoistAcyclic $ plam $ \n x -> go n # (ppredecessor # x) # 1
+    where
+      go ::
+        forall (s' :: S).
+        Term s' PPositive ->
+        Term s' (a :--> PPositive :--> a)
+      go limit = pfix #$ plam $ \self acc count ->
```
SeungheonOh, line 109: Similar observation here about recursion

Plutarch/Enum.hs
SeungheonOh: Maybe it's better to just not provide default instance here as many types like `PInteger` `PPositive` have much better way of dealing with `psuccessorN` than what's provided as default

SeungheonOh (approved):
Some questions, versions needs to be correctly appended. Rest looks good to me. You can merge once versioning is addressed

Plutarch/Enum.hs
kozross: Running `psuccessor` N times would require N to be a constant, which is a bit limiting. 

I reckon that a default instance, assuming we note that it's not particularly efficient, isn't a problem. All of our existing instances are hand-rolled anyway.

SeungheonOh (dismissed):

t4ccer (approved):
I'm not sure if we should `phoistAcyclic` things like `psuccessor = phoistAcyclic $ plam (+)` as I get that `plam (+)` would get inlined as a single builtin otherwise, but you probably need to have benchmarks to support such claim. Good for know I guess, there was also some recent discussion on Plutonomicon discord on excessive use of `phoistAcyclic` so I guess it's better not to block over this and batch review it later.

PR #714, opened 2024-10-02
Allow negative PPosixTime

### LINKED ISSUE #704: `PPOSIXTime` must be able to store negative value

kozross:
Closes #704. The orphanage instance for `POSIXTime` has been updated to have the option to generate negative values; we left the `Interval POSIXTime` instance alone (namely, kept the semantics as they were), as in the only place we use it (`TxInfo`) it would make no sense to have negative endpoints.


plutarch-ledger-api/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+* `pposixTime` and `unPPosixTime`, to construct and deconstruct `PPosixTime`
+  more conveniently
+
+### Changed
+
+* `PPosixTime` now permits negative values as a result of its operations
+
+### Fixed
+
+* Bug in `PNum` instance for `PPosixTime`, where `#*` was implemented as `#+
```
L-as, line 20: Yikes! TBF this shouldn't even be implemented though. Haskell's way of overloading operators leaves much to be desired.

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Time.hs
```diff
@@ -42,68 +47,35 @@ newtype PPosixTime (s :: S) = PPosixTime (Term s (PDataNewtype PInteger))
...
 instance PIntegral PPosixTime where
+  {-# INLINEABLE pdiv #-}
   pdiv = phoistAcyclic $ plam $ \t1 t2 ->
-    pmatch t1 $ \(PPosixTime t1') ->
-      pmatch t2 $ \(PPosixTime t2') ->
-        pmatch t1' $ \(PDataNewtype t1'') ->
-          pmatch t2' $ \(PDataNewtype t2'') ->
-            pcon . PPosixTime . pcon . PDataNewtype . pdata $ pdiv # pfromData t1'' # pfromData t2''
+    pposixTime (pdiv # unPPosixTime t1 # unPPosixTime t2)
+  {-# INLINEABLE pmod #-}
```
L-as, line 53: Why are you inlining this stuff

plutarch-ledger-api/CHANGELOG.md
kozross: Yes, big yikes.

L-as: I feel like the better solution would be to drop all the instances completely, make a new prelude where all operators are "untyped", e.g.
```haskell
class OperatorPlus a b c | a b -> c , a c -> b, b c -> a where
  (+) :: a -> b -> c

instance OperatorPlus PPosixTime PInteger PPosixTime where ...
```

Then we wouldn't have all these arguably incorrect instances.

plutarch-ledger-api/src/Plutarch/LedgerApi/V1/Time.hs
SeungheonOh: I don't think this really matters. It's in-lining Haskell functions anyways. Might as well have better performing haskell functions

SeungheonOh (approved):

Issue #710, opened 2024-07-26
Run plutus optimization pipeline on compile via `Config` option


dragospe:
See https://github.com/IntersectMBO/plutus/blob/c7eb24c72f3be0e9fd218dc340d297d7c9356530/plutus-core/executables/uplc/Main.hs#L254-L265

These could be run on the raw UPLC after compiling.

L-as:
Why not just do this manually? `Config` is essentially just a `Reader` you can use when generating the code to switch on compile-time options (notably, tracing).

PR #715, opened 2024-10-07
Script optimization option

kozross:
Solves #710. The `Config` option was added carefully so as not to break any current setups: the only thing that might happen is an incomplete pattern error. Additionally, we chose _not_ to optimize by default, just in case it does something strange.


Plutarch/Internal.hs
```diff
@@ -745,7 +831,25 @@ compile' t =
 -- | Compile a (closed) Plutus Term to a usable script
 compile :: Config -> ClosedTerm a -> Either Text Script
 compile config t = case asClosedRawTerm t of
-  TermMonad (ReaderT t') -> Script . UPLC.Program () uplcVersion . compile' <$> t' config
+  TermMonad (ReaderT t') -> do
+    configured <- t' config
+    let compiled = compile' configured
+    case optimization config of
+      Optimization -> do
+        case go compiled of
```
L-as, line 839: I don't think this is what you want to do. The configuration is for when you want to check whether it's enabled in user code.
Would user code ever do something different depending on whether an optimization flag is set or not? If no, this is the wrong place to set it (as I assume it is).

Plutarch/Internal.hs
```diff
@@ -745,7 +831,25 @@ compile' t =
 -- | Compile a (closed) Plutus Term to a usable script
 compile :: Config -> ClosedTerm a -> Either Text Script
```
L-as, line 832: You probably just want to add a `compileOptimized` function.

Plutarch/Internal.hs
SeungheonOh: @kozross This is a good point. I don't think we really need optimization configurable since there will rarely be an occasion where code needs to be change depending on it does optimization or not

kozross: I believe we discussed this, and chose to add this option in case optimizations on IOG's end do something undesirable. I agree that a simpler approach would be to optimize whenever `NoTracing` is requested, and not optimize otherwise; alternatively, _always_ optimize regardless of what tracing is requested. Happy to do either of these.

Plutarch/Internal.hs
```diff
@@ -747,6 +751,32 @@ compile :: Config -> ClosedTerm a -> Either Text Script
...
   TermMonad (ReaderT t') -> Script . UPLC.Program () uplcVersion . compile' <$> t' config
 
+{- | As 'compile', but performs UPLC optimizations. Furthermore, this will
+always elide tracing (as if with 'NoTracing').
+
+@since WIP
+-}
+compileOptimized ::
+  forall (a :: S -> Type).
+  (forall (s :: S). Term s a) ->
```
SeungheonOh, line 761: I know it's a GHC misfeature, but `ClosedTerm a` would be better for consistency

SeungheonOh (approved):
We could definitely use some tests. Even simple ones that just runs one script compiled regularly, and other with simplification and see if they functions identically

PR #716, opened 2024-10-07
CIP-122 ops over bytes

kozross:
Adds proper support in `Plutarch.ByteString` for the byte-oriented functionality of CIP-122. Specifically, we support AND, OR and XOR, as well as replication, with some safety wrappers to help it make more sense. As part of this, a limited `PByte` type is introduced.

This also removes a deprecated identifier.


SeungheonOh:
Some of the operations overlaps with https://github.com/Plutonomicon/plutarch-plutus/blob/staging/Plutarch/Bitwise.hs. I think we can move ones from `Bitwise` to this module

t4ccer (approved):

PR #720, opened 2024-10-16
Conversion primitives

kozross:
This provides clearer and safer wrappers around the CIP-121 conversion primops. It also removes the existing (loose) wrappers for the same from `Plutarch.Bitwise`.


Plutarch/Convert.hs
```diff
@@ -0,0 +1,116 @@
...
+a result of the minimal size required: if you want to specify a size, use
+'punsafeIntegerToByteStringSized'. For details, see
+[CIP-121](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0121#builtinintegertobytestring).
+
+= Note
+
+This conversion is unsafe, as it will error when given a non-negative
+integer.
+-}
+punsafeIntegerToByteString ::
```
SeungheonOh, line 98: We don't need unsafe in the name, this function isn't messing around with type/representation of underlying values.

Plutarch/Bitwise.hs
```diff
@@ -21,39 +19,6 @@ import Plutarch.Integer (PInteger)
...
-[CIP-121](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0121).
-The first argument indicates the endianness of the conversion and the third
-argument is the integer to be converted, which must be non-negative.  The
-second argument must also be non-negative and it indicates the required width
-of the output.  If the width is zero then the output is the smallest
-bytestring which can contain the converted input (and in this case, the
-integer 0 encodes to the empty bytestring).  If the width is nonzero then the
-output bytestring will be padded to the required width with 0x00 bytes (on
-the left for big-endian conversions and on the right for little-endian
-conversions); if the input integer is too big to fit into a bytestring of the
```
SeungheonOh, line 38: Documentation about which side the padding will go is gone. Though it's kind of obvious, I think it's good to have.

Plutarch/Convert.hs
```diff
@@ -0,0 +1,116 @@
+{- | Conversions between various types.
+
+@since WIP
+-}
+module Plutarch.Convert (
```
SeungheonOh, line 5: Is there any reason we are not putting bytestring conversion functions with it's friends at Plutarch.ByteString?

Plutarch/Convert.hs
```diff
@@ -0,0 +1,116 @@
...
+
+{- | Convert a 'PByteString' into a 'PInteger', as per
+[CIP-121](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0121#builtinbytestringtointeger).
+
+@since WIP
+-}
+pbyteStringToInteger ::
+  forall (s :: S).
+  Term s (PEndianness :--> PByteString :--> PInteger)
+pbyteStringToInteger = plam $ \e bs -> pmatch e $ \(PEndianness e') ->
```
SeungheonOh, line 85: since this is a uplc application, this is less efficient. We should pass in the endianess flag as Haskell value

Plutarch/Convert.hs
kozross: Should this also apply to the corresponding `PByte` conversion then?

Plutarch/Convert.hs
kozross: `PPositive` would form a cyclic dependency then.

Plutarch/Convert.hs
kozross: What should the resulting signature for `pbyteStringToInteger` be?

SeungheonOh (approved):
Thanks!

PR #722, opened 2024-10-17
Remove Bitwise module, add BitString and operations

kozross:
This finishes properly supporting all CIP-122 and CIP-123 operations. We also eliminate the `Bitwise` module, as it is no longer needed. `pexpModInteger` has to be moved somewhere else for now.


Plutarch/BitString.hs
```diff
@@ -0,0 +1,147 @@
...
+-- | @since WIP
+instance DerivePlutusType PBitString where
+  type DPTStrat _ = PlutusTypeNewtype
+
+{- | Bit access operation, as defined in [CIP-122](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0122#readbit).
+
+@since WIP
+-}
+preadBit :: forall (s :: S). Term s (PBitString :--> PInteger :--> PBool)
+preadBit = plam $ \bs i -> punsafeBuiltin PLC.ReadBit # pto bs # i
```
SeungheonOh, line 78: We don't need `plam` here and every places below. `punsafeBuiltin` will take care of it. This is actually less efficient

Plutarch/BitString.hs
```diff
@@ -0,0 +1,147 @@
...
+-}
+psetBits :: forall (s :: S). Term s (PBitString :--> PBuiltinList PInteger :--> PBitString)
+psetBits = plam $ \bs ixes -> punsafeBuiltin PLC.WriteBits # pto bs # ixes # pconstant True
+
+{- | As 'psetBits', but clears the bits instead.
+
+@since WIP
+-}
+pclearBits :: forall (s :: S). Term s (PBitString :--> PBuiltinList PInteger :--> PBitString)
+pclearBits = plam $ \bs ixes -> punsafeBuiltin PLC.WriteBits # pto bs # ixes # pconstant False
```
SeungheonOh, line 98: What would be rational to have this separately as functions? As above, just `punsafeBuiltin`ing is more efficient. We still get more typelevel assurance as it can still have `PBitString` type.

SeungheonOh (commented):
Some comments about efficiency. 

We need to have some test before releasing this WIP updates

SeungheonOh (dismissed):
Thanks!

SeungheonOh (approved):
Thanks

PR #727, opened 2024-10-24
Fix zero-ada behaviour of mint value generators

t4ccer:
- Fix previous specialized value generators to include zero Ada entry in V1 and V2
- Add new mint value generator for V3 that does not include zero Ada entry as this invariant [changed in V3](https://github.com/IntersectMBO/plutus/blob/384997b4f395c984685a10a8317c7f729c37ff26/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L483-L487)
- Fix mint value guarantee in V1 to `NoGuarantees` from `NonZero` (apparently it was always wrong)
- Fix mint value guarantee in V3 to `NonZero` from `NoGuarantees`

### DIFF
--- a/plutarch-ledger-api/src/Plutarch/LedgerApi/V1.hs
+++ b/plutarch-ledger-api/src/Plutarch/LedgerApi/V1.hs
@@ -182,7 +182,7 @@ deriving via
 -- | @since 3.1.1
 instance PTryFrom PData (PAsData PTxInInfo)
 
--- | @since 3.1.1
+-- | @since WIP
 newtype PTxInfo (s :: S)
   = PTxInfo
       ( Term
--- a/plutarch-ledger-api/src/Plutarch/LedgerApi/V1.hs
+++ b/plutarch-ledger-api/src/Plutarch/LedgerApi/V1.hs
@@ -191,7 +191,7 @@ newtype PTxInfo (s :: S)
               '[ "inputs" ':= PBuiltinList (PAsData PTxInInfo)
                , "outputs" ':= PBuiltinList (PAsData PTxOut)
                , "fee" ':= Value.PValue 'AssocMap.Sorted 'Value.Positive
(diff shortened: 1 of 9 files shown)


plutarch-ledger-api/src/Plutarch/LedgerApi/V1.hs
```diff
@@ -191,7 +191,7 @@ newtype PTxInfo (s :: S)
               '[ "inputs" ':= PBuiltinList (PAsData PTxInInfo)
                , "outputs" ':= PBuiltinList (PAsData PTxOut)
                , "fee" ':= Value.PValue 'AssocMap.Sorted 'Value.Positive
-               , "mint" ':= Value.PValue 'AssocMap.Sorted 'Value.NonZero -- value minted by transaction
+               , "mint" ':= Value.PValue 'AssocMap.Sorted 'Value.NoGuarantees -- value minted by transaction
```
kozross, line 194: The `@since` on this data type needs to be `WIP` since we've changed the type.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Contexts.hs
```diff
@@ -828,7 +828,7 @@ newtype PTxInfo (s :: S)
                , "referenceInputs" ':= PBuiltinList (PAsData PTxInInfo)
                , "outputs" ':= PBuiltinList (PAsData PTxOut)
                , "fee" ':= Value.PLovelace
-               , "mint" ':= Value.PValue 'AssocMap.Sorted 'Value.NoGuarantees -- value minted by transaction
+               , "mint" ':= Value.PValue 'AssocMap.Sorted 'Value.NonZero -- value minted by transaction
```
kozross, line 831: Same here.

kozross (changes requested):

kozross (approved):

Issue #643, opened 2024-01-18
LambdaBuffers Plutarch Runtime cleanup


bladyjoker:
Some items for Plutarch if someone can take a look at that (NOTE: I don't use `Data` anywhere in naming because everything is `Data`, and on that note, imo `Scott` should have been explicitly named not `Data` which is much more prevalent)

1. Missing PAssetClass

https://github.com/mlabs-haskell/lambda-buffers/blob/2181c9493ef7a15ca396f4a8bb9ed03dbf7c4c5d/runtimes/haskell/lbr-plutarch/src/LambdaBuffers/Runtime/Plutarch.hs#L60

2. Missing PEither

https://github.com/mlabs-haskell/lambda-buffers/blob/2181c9493ef7a15ca396f4a8bb9ed03dbf7c4c5d/runtimes/haskell/lbr-plutarch/src/LambdaBuffers/Runtime/Plutarch.hs#L75

3. Messed up PMaybe

https://github.com/mlabs-haskell/lambda-buffers/blob/2181c9493ef7a15ca396f4a8bb9ed03dbf7c4c5d/runtimes/haskell/lbr-plutarch/src/LambdaBuffers/Runtime/Plutarch.hs#L82

4. Had to redefined PList as well

https://github.com/mlabs-haskell/lambda-buffers/blob/2181c9493ef7a15ca396f4a8bb9ed03dbf7c4c5d/runtimes/haskell/lbr-plutarch/src/LambdaBuffers/Runtime/Plutarch.hs#L668

5. A bunch of missing instances throughout this module

SeungheonOh:
I don't think we'll add `PList` here since it is a bit too specific in LB. Other things we will try to add in new, cleaned up prelude/ledger api

bladyjoker:
> I don't think we'll add `PList` here since it is a bit too specific in LB. Other things we will try to add in new, cleaned up prelude/ledger api

```
{- | PList because PBuiltinList misses `PAsData` on its constituents which causes type errors when used.
TODO(bladyjoker): Upstream these changes or fix PBuiltinList.
-}
```
How's that specific to LB? There's a bunch of types that appear to escape the mold for some reason that other types use. For example, some stuff requires `PLift` which is out of place. We should do a pass on all the types in the refactor and make sure they are aligned properly, and devoid of non-core classes.

SeungheonOh:
[...]

PR #728, opened 2024-10-24
PEitherData

kozross:
This adds a `Data`-encoded counterpart to `PEither`, along with some helper functions. Aims to address #643 .


Plutarch/Either.hs
```diff
@@ -9,13 +33,170 @@ import Plutarch (
...
+
+instance DerivePlutusType (PEither a b) where
+  type DPTStrat _ = PlutusTypeScott
+
+{- | @Data@-encoded 'Either'.
+
+@since WIP
+-}
+data PEitherData (a :: S -> Type) (b :: S -> Type) (s :: S)
+  = PDLeft (Term s (PDataRecord '["_0" ':= a]))
```
bladyjoker, line 77: Why not just 'a'? I'd avoid PDataRecord if possible

bladyjoker (commented):
I'd really appreciate if PDataRecord could be removed in PEitherData if possible.

It's convenient for deriving stuff, but honestly on the use site it's pretty bad.

bladyjoker (approved):
❤️

PR #732, opened 2024-10-29
Fix pinterval and phull

t4ccer:
They assigned invalid open/close ends of the intervals. Issue was that the same function was used to
compute `min/max` of lower and upper bounds, but they actually have different implementations.

Also, that somehow improved performance


t4ccer:
Closes https://github.com/Plutonomicon/plutarch-plutus/issues/725, now we have full coverage of all Interval functions

kozross (approved):

L-as:
Aren't these security vulnerabilities?

PR #734, opened 2024-10-30
Value representation rework

### LINKED ISSUE #630: UPLC SOP support
Add support for UPLC SOP prim-op added in UPLC v1.1.0

SeungheonOh:
nothing much yet

closes #630 and probably bunch more

### DIFF
--- a/Plutarch/Internal.hs
+++ b/Plutarch/Internal.hs
@@ -70,6 +70,7 @@ import Data.Set qualified as S
 import Data.String (fromString)
 import Data.Text (Text)
 import Data.Text qualified as Text
+import Data.Vector qualified as V
 import Flat.Run qualified as F
 import GHC.Stack (HasCallStack, callStack, prettyCallStack)
 import GHC.Word (Word64)
@@ -116,6 +117,8 @@ data RawTerm
   | RError
   | RHoisted HoistedTerm
   | RPlaceHolder Dig
+  | RConstr Word64 [RawTerm]
+  | RCase RawTerm [RawTerm]
   deriving stock (Show)
 
 addHashIndex :: forall alg. HashAlgorithm alg => Integer -> Context alg -> Context alg
@@ -148,6 +151,10 @@ hashRawTerm' RError = addHashIndex 7
 hashRawTerm' (RHoisted (HoistedTerm hash _)) = addHashIndex 8 . flip hashUpdate hash
 hashRawTerm' (RCompiled code) = addHashIndex 9 . flip hashUpdate (hashUTerm @alg code hashInit)
 hashRawTerm' (RPlaceHolder hash) = addHashIndex 10 . flip hashUpdate hash
+hashRawTerm' (RConstr x y) =
+  addHashIndex 11 . flip hashUpdate (F.flat (fromIntegral x :: Integer)) . flip (foldl' $ flip hashRawTerm') y
+hashRawTerm' (RCase x y) =
+  addHashIndex 12 . hashRawTerm' x . flip (foldl' $ flip hashRawTerm') y
 
 hashRawTerm :: RawTerm -> Dig
 hashRawTerm t = hashFinalize . hashRawTerm' t $ hashInit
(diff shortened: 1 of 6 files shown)


L-as:
What's your idea?

SeungheonOh:
Have `PStruct` and `PRec` represent any kind of composite data types, and have `PDataRec`, `PScottRec`, ... to represent their encoded format. This will give types that encodes same structure differently for free and also conversion from and to different encoding for free.

SeungheonOh:
Having to deal with 
```
type MyPMaybeRepr a = UnTermStruct (MyPMaybe S' a)
```
is kind of annoying : / Any suggestions are welcome

Plutarch/Internal.hs
```diff
@@ -144,7 +146,8 @@ hashRawTerm' (RConstant x) = addHashIndex 5 . flip hashUpdate (F.flat x)
 hashRawTerm' (RBuiltin x) = addHashIndex 6 . flip hashUpdate (F.flat x)
 hashRawTerm' RError = addHashIndex 7
 hashRawTerm' (RHoisted (HoistedTerm hash _)) = addHashIndex 8 . flip hashUpdate hash
-hashRawTerm' (RCompiled code) = addHashIndex 9 . flip hashUpdate (hashUTerm @alg code hashInit)
+hashRawTerm' (RPlaceHolder hash) = addHashIndex 9 . flip hashUpdate hash
+hashRawTerm' (RCompiled code) = addHashIndex 10 . flip hashUpdate (hashUTerm @alg code hashInit)
```
kozross, line 150: I'd rather we kept existing hashing choices the same. It probably doesn't matter, but I'd rather the hash index of `RCompiled` stayed as `9`.

Plutarch/Rework.hs
```diff
@@ -0,0 +1,335 @@
...
+data PDataStruct (struct :: [[S -> Type]]) (s :: S) where
+  PDataStruct :: (All2 PDataRepresentable struct, All2 PIsData struct) => PStruct struct s -> PDataStruct struct s
+
+data PDataRec (struct :: [S -> Type]) (s :: S) where
+  PDataRec :: (All PDataRepresentable struct, All PIsData struct) => PRec struct s -> PDataRec struct s
+
+class PDataRepresentable (a :: S -> Type)
+instance PDataRepresentable (PDataStruct struct)
+instance PDataRepresentable PInteger
+instance PDataRepresentable PByteString
```
kozross, line 104: What does this type class actually mean?

L-as:
Why does `MyPMaybeRepr` exist at all? I think there is a way of making the functions generic over it too BTW.

I also think you can do this without having it in the types at all.

SeungheonOh:
> I think there is a way of making the functions generic over it too.
any examples? I don't understand. 

Also, I'm somewhat sure this can be done without embedding these into types, but I don't want to use Haskell's generics, they are pain to deal with. Also don't want TH for similar reasons

L-as:
You can change the ast a bit such that operations over data types are more abstract. Everything could be SOP-like in the AST. When you lower it to UPLC you could choose at each site what is most convenient, and according to user preference.

There are multiple strategies here, some optimize size, some performance, and some of the problems are in NP.

Plutarch/Rework.hs
```diff
@@ -0,0 +1,491 @@
...
+    go _ = False
+
+  pure $ f (go rt)
+
+createPlaceholder :: Term s a -> TermCont s (Dig, Term s a)
+createPlaceholder t = do
+  (hash :: Dig) <- hashOpenTerm t
+  pure (hash, Term $ const $ pure $ TermResult (RPlaceHolder hash) [])
+
+pletSmart :: Term s a -> (Term s a -> Term s b) -> Term s b
```
L-as, line 81: Does this account for stuff like

```
let x = ... in map (\_ -> x) list
```

Note that UPLC is strict.

SeungheonOh: yeah, it won't be able to figure out if bound term will be used multiple times or not

L-as: The solution is to cover a small set of cases that are useful and happen in practice and are easy enough to detect.
You can replace it with checking whether the bound variable ever appears under a `delay` or `lam`. If it does not, that should mean it's always only evaluated exactly once, hence you can inline it without changing _almost_ anything.

Three properties to uphold:
- Algorithmic complexities must not change
- What errs without optimizations must not err with optimizations.
- What succeeds without optimizations must succeed with optimizations.

The bottom two are a very unfortunate side effect of UPLC's design. I made a CIP <https://github.com/cardano-foundation/CIPs/pull/469> to fix this, but it wasn't very popular.

L-as: The topmost one wouldn't be affect by doing laziness like Haskell though, since sharing doesn't happen under lambdas. It does though on HVM and stuff like that AFAIU.

SeungheonOh:
Waiting on module restructuring

Issue #735, opened 2024-10-30
Please expose internal variants that are not safety opinionated


colll78:
https://github.com/Plutonomicon/plutarch-plutus/blob/93833546d61d7a463d3ec25e8fbd16ba1abd0ec1/Plutarch/ByteString.hs#L240C1-L291C43

The unique value proposition of Plutarch over other languages targeting PlutusCore is that it is not opinionated. It gives the developer full low-level control over everything from strictness to inlining vs hoisting to binding and so forth. This is important because if you are very knowledgeable in the domain of Cardano smart contracts, you can write extremely efficient code that you cannot write in the other languages where the compilers are more opinionated (ie the compiler makes choices on your behalf regarding strictness or whether something is inlined or not, etc). 

Please don't include any opinioned logic in builtin internal functions, if you truly want such logic please include it a builtin external module so that developers who don't want that logic can still use the internal module (which I expect is at-least 95% of the developers using Plutarch since it is a hyper-efficiency focused DSL).

For example,
```
preplicateBS :: forall (s :: S). Term s (PInteger :--> PByte :--> PByteString)
preplicateBS = phoistAcyclic $ plam go
  where
    go :: forall (s' :: S). Term s' PInteger -> Term s' PByte -> Term s' PByteString
    go len w8 =
      punsafeBuiltin PLC.ReplicateByte
        # pif (len #< 0) 0 len
        # punsafeCoerce @_ @_ @PInteger w8
        
pzeroesBS :: forall (s :: S). Term s (PInteger :--> PByteString)
pzeroesBS = ...

ponesBS = ...
```

All of these functions have this `pif (len #< 0) 0 len` to check that the value is non-negative. There is no reason for this check. Optional types and so forth for delayed failure are an anti-pattern in onchain code, why would you ever want to pass this a negative integer and not have it error? I'd imagine this might even cause some vulnerabilities if the caller is not expecting negative integers to be passed. Even if you do want this check in some weird edge-cases, why should you make that the default behavior? 

Please just expose all PLC.ANYTHINGHERE directly as a Plutarch binding with no additional opinionated logic.

L-as:
@colll78 you can still use `punsafeBuiltin` directly I think, but obviously this code is wrong. Missed in review I suppose.

L-as:
You could always make a PR.

colll78:
> You could always make a PR.

[...]

PR #738, opened 2024-11-04
Reorganize module hierarchy

kozross:
This is the first attempt to try and make some sense of the module hierarchy. The overall goals were:

1. Users should need to only import `Plutarch.Prelude` and get everything they need
2. Plutarch itself shouldn't import `Plutarch.Prelude`, only things in the `Plutarch.Internal.*` namespace
3. `plutarch-ledger-api` and `plutarch-testlib` should rely minimally on imports from `Plutarch.Internal.*`

Additionally, there were a couple of changes:

1. All builtins should have a `pbuiltinMyBuiltinName` in `Plutarch.Internal.Builtin`, allowing them to be called without having to wonder what their signatures are.
2. `Plutarch.*` has been significantly emptied.
3. `Plutarch.Internal.Builtin` contains (most of) the types that we _have_ to have because they're mandated by Plutus.
4. Some modules got merged.

This is not the final outcome, I just wanted to get _something_ reviewed and merged, or the diffs would get too huge.


src/Plutarch/Internal/Builtin.hs
```diff
@@ -0,0 +1,810 @@
+{-# LANGUAGE FlexibleInstances #-}
```
SeungheonOh, line 1: This is quite massive. More so than I thought, would it be better to make `Plutarch.Builtins.XYZ` modules instead?

kozross: Yeah, I agree.

SeungheonOh:
Will also resolve #735

SeungheonOh:
Notes: this is closed because of nasty conflict

PR #739, opened 2024-11-06
Initial PLiftable draft

kozross:
This implements `PLiftable`, designed to unify and replace `PConstantDecl` and `PUnsafeLiftDecl`. This avoids orphans, as well as having a much clearer design. This PR doesn't include instances: we can add them later.


plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+result in an `Integer` (or an error), and `Integer`s are part of the Plutus
+default universe. Indeed, if you attempt to use this helper with a Haskell type
+that _isn't_ part of the default Plutus universe, you will get a compile error.
+
+The type is defined as follows: its implementation is not important.
+
+```haskell
+newtype DeriveBuiltinPLiftable (a :: S -> Type) (h :: Type) (s :: S)
+  = DeriveBuiltinPLiftable (a s)
+``
```
SeungheonOh, line 119: missing "`" here

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+
+We can see that this `newtype` has two type arguments (besides the `s` to make
+it a Plutarch type): one for the Plutarch type for which we want to derive the
+instance, and one for a Haskell type that is meant to be its `AsHaskell`
+equivalent. To use the derivation helper, you write something like the
+following, substituting `a` and `h` as appropriate:
```
SeungheonOh, line 125: example like how we derive PLiftable for `PInteger` here would be nice

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+deriving via (DeriveDataPLiftable a h) instance PLiftable a
+```
+
+There are more stringent requirements of `a` and `h` for this to work. Aside
+from `a` being an instance of `PlutusType`, we also must ensure that:
+
+* `PInner a` is `PData` (namely, construction and matching is actually carried
+  out in a computation involving `Data`)
+* `h` is an instance of both `ToData` and `FromData` (namely, it has a `Data`
+  encoding that we can decode from and encode into)
```
SeungheonOh, line 163: Some example here as well. Some ledger type i guess

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+We can see that this `newtype` has two type arguments (besides the `s` to make
+it a Plutarch type): one for the Plutarch type for which we want to derive the
+instance, and one for a Haskell type that is meant to be its `AsHaskell`
+equivalent. To use the derivation helper, you write something like the
+following, substituting `a` and `h` as appropriate:
+
+```haskell
+-- ensure that a is an instance of Plutus type, and h is in the default universe
+
+deriving via (DeriveBuiltinPLiftable a h) instance PLiftable a
```
SeungheonOh, line 130: This feels like a poor example.

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+  = DeriveDataPLiftable (a s)
+```
+
+Similarly to `DeriveBuiltinPLiftable`, we have two relevant type arguments: 
+one for the Plutarch type for which we want to derive an instance, and 
+another for its Haskell-level equivalent. Much as with `DeriveBuiltinPLiftable`,
+we use `DeriveDataPLiftable` as so, substituting `a` and `h` appropriately:
+
+```haskell
+deriving via (DeriveDataPLiftable a h) instance PLiftable a
```
SeungheonOh, line 154: Again, poor example

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+```
+1. fromPlutarch . toPlutarch = Right
+2. fmap toPlutarch . fromPlutarch = Right
+```
+
+If you use either of `DeriveBuiltinPLiftable` or `DeriveDataPLiftable` with
+`via` derivation, these laws will automatically be satisfied. If you write an
+instance manually, you will have to ensure this yourself. We provide a helper
+for testing that these laws hold in `plutarch-testlib`, in the
+`Plutarch.Test.Laws` module, called `checkPLiftableLaws`, which uses QuickCheck
```
SeungheonOh, line 187: Let's show how to define test case for the law. Feel like this needs more emphasis

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
...
+    AsHaskell a
+```
+
+The type signatures more-or-less speak for themselves: `pconstant` is the
+Haskell-to-Plutarch direction, while `plift` is the Plutarch-to-Haskell
+direction. There are three minor caveats to their use:
+
+* `pconstant` is technically ambiguous, as many different Plutarch types can
+  share the same choice for `AsHaskell`. Thus, you may need to use type
+  arguments when calling `pconstant` to avoid ambiguity errors from the
```
SeungheonOh, line 220: Show something like how there can be both
`PLiftable PInteger` and `PLiftable (PAsData PInteger)` both from Haskell `Integer` in which case user would need to `pconstant @(PAsData PInteger)` or `pconstant @PInteger`

plutarch-docs/src/Typeclasses/PLiftable.md
```diff
@@ -0,0 +1,231 @@
+# `PLiftable`
```
SeungheonOh, line 1: We can technically make this into a literate haskell where we can have snippet that will be compiled to check validity. But this document looks good without it. We can probably reserve that for other guides more oriented for usages.

Plutarch/Internal/Lift.hs
```diff
@@ -0,0 +1,263 @@
...
+  PLifted a s
+unsafeUniToPLifted =
+  PLifted . punsafeConstantInternal . PLC.someValue @h @PLC.DefaultUni
+
+{- | Helper for writing 'PLiftable' instances. This is /highly/ unsafe, and only
+suitable for internal use!
+
+@since WIP
+-}
+unsafePLiftedToUni ::
```
SeungheonOh, line 250: Nitpick. We have `[to|from]Plutarch` but this is "to" both ways. It'd be nice to keep naming consistent

Plutarch/Internal/Lift.hs
```diff
@@ -0,0 +1,263 @@
...
+
+  -- ** Via-helpers
+  DeriveBuiltinPLiftable (..),
+  DeriveDataPLiftable (..),
+
+  -- ** Manual instance helpers
+  PLifted (..),
+  unPLifted,
+  LiftError (..),
+  unsafeUniToPLifted,
```
SeungheonOh, line 25: One that goes other way is not exported.

SeungheonOh (dismissed):
Some suggestions on documentation. Implementation looks good to me

Plutarch/Internal/Lift.hs
kozross: So `unsafePLiftedToUni` and `unsafePLiftedFromUni`?

SeungheonOh (approved):

PR #753, opened 2024-11-20
`PLiftable` rework

### LINKED ISSUE #749: Convert to PLiftable
We need to replace all uses, and instances, of `PUnsafeLiftDecl` and `PConstantDecl` with `PLiftable`. This has to be done 'all at once', and may require additional definitions or bits. Instances will be the primary work, because `pconstant` and `plift` have the same semantics (and signatures) modulo minor definition changes.

t4ccer:
Closes #749


Plutarch/Internal/Lift.hs
```diff
@@ -79,58 +87,112 @@ Plutus by extension), and we have the ability to move between them.
...
-   \'pull\' terms out.
-2. 'fromPlutarch' involves compilation and evaluation. When doing so, ensure
-   that you use the largest possible budgets to avoid weird problems.
-3. When choosing a type for 'AsHaskell', /any/ value of that type /must/ be
+1. You still shouldn't write every method by hand, there are helpers
+   @fromPlutarch*@ and @toPlutarch*@ to cover common cases like types
+   in Plutus universe or Scott encoding
+2. If defining 'toPlutarchRepr' and 'fromPlutarchRepr' you will need to define
+   an associated 'PlutusRepr' type, this is a Hasekll level type that is included
+   in 'DefaultUni' - universe of Plutus
```
kozross, line 108: Just say 'is included in the Plutus default universe'.

Plutarch/Internal/Lift.hs
```diff
@@ -79,58 +87,112 @@ Plutus by extension), and we have the ability to move between them.
...
 
 = Laws
 
-1. @'fromPlutarch' '.' 'toPlutarch'@ @=@ @'Right'@
-2. @'fmap' 'toPlutarch' '.' 'fromPlutarch'@ @=@ @'Right'@
+1. @'fromPlutarchRepr' '.' 'toPlutarchRepr'@ @=@ @'Just'@
+2. @'fmap' 'toPlutarchRepr' '.' 'fromPlutarchRepr'@ @=@ @'Just'@
+3. @'fromPlutarch' '.' 'toPlutarch'@ @=@ @'Right'@
+4. @'fmap' 'toPlutarch' '.' 'fromPlutarch'@ @=@ @'Right'@
+5. @'plift' '.' 'pconstant'@ @=@ @'id'@
```
kozross, line 121: Those functions aren't type class methods, so law 5 is technically redundant.

Plutarch/Internal/Lift.hs
```diff
@@ -214,51 +285,83 @@ instance DerivePlutusType (DeriveDataPLiftable a h) where
...
-suitable for internal use!
+  {-# INLINEABLE fromPlutarchRepr #-}
+  fromPlutarchRepr = coerce . fromPlutarchRepr @inner
+
+  {-# INLINEABLE fromPlutarch #-}
+  fromPlutarch p = coerce . fromPlutarch @inner $ punsafeCoercePLifted p
+
+{- | Similar to 'Identity', but at the level of Plutarch. Only needed when
+writing manual instances of 'PLiftable', or if you want to use 'toPlutarch'
+and 'fromPlutarch' directly.
```
SeungheonOh, line 343: This is strictly for `coerce`ing Haskell terms. That is running `pcoerceUnsafe :: Term s a -> Term s b` in Haskell level `coerce :: PLifted s a -> PLifted s b`. It'd be nice to note that maybe

SeungheonOh: Actually, I think I can remove this and still make everything work. I'll try this afternoon

Plutarch/Internal/Lift.hs
```diff
@@ -214,51 +285,83 @@ instance DerivePlutusType (DeriveDataPLiftable a h) where
...
+
+-- | @since WIP
+mkPLifted :: Term s a -> PLifted s a
+mkPLifted t = PLifted (popaque t)
+
+{- |  Use this as 'PlutusRepr' when defining 'PLiftable' instance for Scott encoded type
+
+@since WIP
+-}
+newtype PLiftedScott (a :: S -> Type) = PLiftedScott {unPLiftedClosed :: forall (s :: S). PLifted s a}
```
SeungheonOh, line 367: This should be just `PClosedLifted` It's not only for scott. We'll probably have SOP support this

kozross (changes requested):

SeungheonOh (commented):
some comments that I might forget latter

Plutarch/Internal/Lift.hs
SeungheonOh: Yeah, SOP's implementation should be very similar to scott one so this function should be named more generally

Plutarch/Internal/Lift.hs
t4ccer: I've tried but `coerce` yelled at me during `via` deriving

kozross (approved):

PR #756, opened 2024-11-20
Initial reorganization

kozross:
This does a few things:

* `Plutarch.Internal` is now `Plutarch.Internal.Term`, to better reflect its actual contents
* `Plutarch` module is gone, as it was confusing and arguably didn't do anything
* `PType` synonym is gone

To anticipate future stability guarantees, the `Plutarch.Prelude` now explicitly specifies from which module any given identifier comes. This will help us track this information and ensure we don't accidentally export too much or too little.


Plutarch/DataRepr/Internal.hs
```diff
@@ -108,29 +80,61 @@ import Plutarch.DataRepr.Internal.HList (
...
+  pupcast,
+ )
 import Plutarch.Unit (PUnit (PUnit))
 import Plutarch.Unsafe (punsafeCoerce)
 
 {- | A "record" of `exists a. PAsData a`. The underlying representation is
  `PBuiltinList PData`.
 -}
-data PDataRecord (as :: [PLabeledType]) (s :: S) where
+data PDataRecord (as :: [PLabeledType]) (s :: P.S) where
```
SeungheonOh, line 137: I prefer no quantification for `S` as per other modules

SeungheonOh (approved):
looks good to me, no changes on any core functions, just moving around

Plutarch/DataRepr/Internal.hs
kozross: Unavoidable here unfortunately - this module uses the `S` from `GHC.Generics` as well.

PR #758, opened 2024-11-21
Break apart Plutarch.Bool

kozross:
This does several things:

* `PSBool` is now in `Plutarch.LedgerApi.Utils`, as it's not used anywhere else
* `PBool` is now housed in `Plutarch.Builtin.Bool`, along with its major functionality
* A new binding for the sole boolean builtin (with the right type) exists
* `PEq` now resides in its own module
* `PPartialOrd` and `POrd` now reside in their own modules
* `pif'` is now exported from the Prelude


SeungheonOh (approved):
LGTM. Thanks

PR #762, opened 2024-11-25
Optimize term inlining

t4ccer:


plutarch-testlib/goldens/unit.uplc.golden
```diff
@@ -1,5 +1,5 @@
 pcon program 1.0.0 ()
-pmatch program 1.0.0 ((\!0 -> True) ())
-compare.== program 1.0.0 ((\!0 -> (\!0 -> True) ()) ())
-compare.< program 1.0.0 ((\!0 -> (\!0 -> False) ()) ())
-compare.<= program 1.0.0 ((\!0 -> (\!0 -> True) ()) ())
+pmatch program 1.0.0 True
+compare.== program 1.0.0 True
+compare.< program 1.0.0 False
+compare.<= program 1.0.0 True
```
t4ccer, line 5: Is this correct though?

SeungheonOh (changes requested):
This pretty confusing. Changes made in `punsafeConstantInternal` should not change golden outputs because same terms will get inlined as no conditions have changes.

plutarch-testlib/goldens/either.uplc.golden
```diff
@@ -4,37 +4,37 @@ eq.true.left program
       !2
         (\!0 -> !2 (\!0 -> equalsInteger !2 !1) (\!0 -> False))
         (\!0 -> !2 (\!0 -> False) (\!0 -> equalsInteger !2 !1)))
-     ((\!0 !0 !0 -> !2 !3) 42)
```
SeungheonOh, line 7: These shouldn't also change as 42 is under 256 :thinking:

plutarch-testlib/goldens/pisdata.lte.uplc.golden
```diff
@@ -378,19 +378,19 @@ PCredential.pmatch-pdatarecord.false program
                         (\!0 ->
                            force
                              (!4
-                                (equalsInteger 1 !2)
```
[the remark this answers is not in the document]
t4ccer: Looks like it just changed places? https://github.com/Plutonomicon/plutarch-plutus/pull/762/files#diff-498fa1577eac293c414f4df8eece64c7fad5d6f75478e3c95c5ebad5d21d5853L330-R330

SeungheonOh (dismissed):
ignore my previous review. 

Changes are:
- All constant terms are first created as Hoisted term
- inlineable constants are inlined when compiling

This fixes issue of constanted term not getting inlined when they are applied to a function right after

SeungheonOh (approved):

Plutarch/Internal/Term.hs
```diff
@@ -638,14 +638,8 @@ punsafeConstant = punsafeConstantInternal
 
 punsafeConstantInternal :: Some (ValueOf PLC.DefaultUni) -> Term s a
 punsafeConstantInternal c = Term \_ ->
-  pure $ case c of
-    -- These constants are smaller than variable references.
```
L-as, line 642: Why?

Plutarch/Internal/Term.hs
```diff
@@ -718,12 +719,15 @@ compile' t =
...
+      hoistedTermRaw :: HoistedTerm -> RawTerm
+      hoistedTermRaw (HoistedTerm _ t) = t
+
       toInline :: S.Set Dig
       toInline =
         S.fromList
           . fmap (\(HoistedTerm hash _) -> hash)
           . (head <$>)
-          . filter ((== 1) . length)
+          . filter (\terms -> length terms == 1 || smallEnoughToInline (hoistedTermRaw $ head terms))
```
L-as, line 730: I see, but why make this change? Does it fix anything?

SeungheonOh: It's slightly more efficient when constanted value is pletted. Previously, plet stays, but with this, plet of constanted value will get automatically inlined. 

It's not ideal, because it wont' inline application of inlinable constant values, but still strictly better than original one

SeungheonOh: ex;

```hs
pcon $ PJust $ pconstant 10

--previously: (\x -> (\a b -> a x)) 10
--new: (\a b -> a 10)
```

t4ccer: It inlined constants too early so they were applied to lambdas, this instead inlines inside

PR #765, opened 2024-11-26
Separate PInteger and PNum

kozross:
This refactors both `PNum` and `PIntegral` into a new `Plutarch.Internal.Numeric` module, as well as moving `PInteger` to a module in the `Plutarch.Builtin` hierarchy. I've also reversed the deps for `PEq`, `PPartialOrd` and `POrd`, similarly to `PBool`, for `PInteger`.


SeungheonOh (approved):
Note: Strict restructuring without performance changes. 

Looks good to me.

PR #768, opened 2024-11-27
Remove PPartialOrd

### LINKED ISSUE #766: Kill `PPartialOrd`
`PPartialOrd` is solely for `PValue`. Everything else have total ordering. 

Instead, `PValue` will get specialized functions as it is on Plutus Ledger Api

kozross:
Fixes #766. As part of this, there were a few other small changes:

* `#>` is now a function instead of a method. Since it's just an argument rearrangement, it's not possible for it to be implemented more efficiently per instance anyway.
* `PValue` gained new functions to replace its old `PPartialOrd` instances. As these are meant to mirror functionality from `plutus-ledger-api`, we make mention of that.
* The default implementations for `pmax` and `pmin` now use `#<=` instead of `#<`, which gives us slight perf bumps.

Things that would be nice (and need follow-up tickets):

* Checks for `POrd` actually being a total order (namely, that the laws for its mandatory methods hold)
* Goldens/benches/whatever to compare whether a custom implementation of a `POrd` method is actually better than the default, for each optional method
* Checks that `pltPositive` and similar functions mirror their `plutus-ledger-api` equivalents

### DIFF
--- a/Plutarch/Internal/Ord.hs
+++ b/Plutarch/Internal/Ord.hs
@@ -1,9 +1,8 @@
 module Plutarch.Internal.Ord (
-  PPartialOrd (..),
   POrd (..),
 ) where
 
-import Plutarch.Builtin.Bool (PBool, pif', pnot)
+import Plutarch.Builtin.Bool (PBool, pand', pif', pnot, por')
 import Plutarch.Builtin.Integer (PInteger)
 import Plutarch.Internal.Eq (PEq)
 import Plutarch.Internal.Lift (pconstant)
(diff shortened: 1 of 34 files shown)


Plutarch/Internal/Ord.hs
```diff
@@ -18,98 +19,96 @@ import Plutarch.Internal.Term (
...
+  -- | @since WIP
+  {-# INLINEABLE (#<) #-}
   (#<) :: Term s t -> Term s t -> Term s PBool
   default (#<) :: POrd (PInner t) => Term s t -> Term s t -> Term s PBool
   x #< y = pto x #< pto y
 
   -- | @since WIP
+  {-# INLINEABLE (#>=) #-}
   (#>=) :: forall (s :: S). Term s t -> Term s t -> Term s PBool
   x #>= y = pnot #$ x #< y
```
SeungheonOh, line 73: Can't this also be argument re arrangement of #<= ?

kozross: Hmm, yeah, good point.

SeungheonOh (approved):
Thanks!

PR #774, opened 2024-11-28
Add more `PValue` tests

t4ccer:
More tests, fix `pltNonZero`

### DIFF
--- a/plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
+++ b/plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
@@ -38,6 +38,7 @@ module Plutarch.LedgerApi.Value (
 
   -- ** Transformation
   passertPositive,
+  passertNonZero,
   passertSorted,
   pforgetPositive,
   pforgetSorted,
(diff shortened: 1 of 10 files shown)


plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
```diff
@@ -275,8 +276,7 @@ pltNonZero ::
   Term s (PValue 'AssocMap.Sorted 'NonZero) ->
   Term s (PValue 'AssocMap.Sorted 'NonZero) ->
   Term s PBool
-pltNonZero t1 t2 =
-  phoistAcyclic (pcheckBinRel #$ phoistAcyclic $ plam (#<)) # t1 # t2
+pltNonZero t1 t2 = pleqNonZero t1 t2 #&& (pnot # (t1 #== t2))
```
t4ccer, line 279: There should be a smarter way but I don't know how. `pcheckBinRel #<` does not work cause it defaults to `0` in missing case and then there are issues with empty maps, etc.

kozross (approved):

PR #783, opened 2024-12-02
Roll in Plutarch.Convert

### LINKED ISSUE #778: Roll Convert module into ByteString module
The original separation was required due to a dependency cycle, which no longer exists.

kozross:
Closes #778. Also fixes a few issues in the ByteString module.


Plutarch/Builtin/ByteString.hs
```diff
@@ -123,15 +189,17 @@ consisting entirely of zero bytes.
 @since WIP
 -}
 pzeroesBS :: forall (s :: S). Term s (PInteger :--> PByteString)
-pzeroesBS = punsafeBuiltin PLC.ReplicateByte
+pzeroesBS = phoistAcyclic $ plam $ \len ->
+  preplicateBS # len #$ pintegerToByte # pconstantInteger 0
```
SeungheonOh, line 193: We don't want to have these here. Plutarch.Builtin.* must stay without any "additional checks"

SeungheonOh: If we want to have "safe" version, that should go somewhere else other than builtin modules. We only keep "raw" versions here

kozross: We can probably just remove these. They don't do anything you can't do with builtins anyway.

Plutarch/Builtin/ByteString.hs
```diff
@@ -218,6 +265,57 @@ phexByteStr = punsafeConstantInternal . PLC.someValue . BS.pack . f
...
+[CIP-121](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0121#builtinbytestringtointeger).
+
+@since WIP
+-}
+pbyteStringToInteger ::
+  forall (s :: S).
+  Term s PEndianness ->
+  Term s (PByteString :--> PInteger)
+pbyteStringToInteger e = plam $ \bs ->
+  punsafeBuiltin PLC.ByteStringToInteger # punsafeCoerce @PBool e # bs
```
t4ccer, line 278:
```suggestion
  punsafeBuiltin PLC.ByteStringToInteger # pto e # bs
```
does this work?

t4ccer (approved):

Plutarch/Builtin/ByteString.hs
SeungheonOh: just `pbyteStringToInteger = punsafeBuiltin PLC.ByteStringToInteger` is enough

Plutarch/Builtin/ByteString.hs
```diff
@@ -47,6 +83,33 @@ instance Semigroup (Term s PByteString) where
...
+-}
+pmostSignificantFirst :: forall (s :: S). Term s PEndianness
+pmostSignificantFirst = punsafeCoerce ptrue
+
+{- | Indicates the conversion should be most-significant-last.
+
+@since WIP
+-}
+pmostSignificantLast :: forall (s :: S). Term s PEndianness
+pmostSignificantLast = punsafeCoerce pfalse
```
t4ccer, line 111:
```suggestion
pmostSignificantFirst :: forall (s :: S). Term s PEndianness
pmostSignificantFirst = pcon $ PEndianness ptrue

{- | Indicates the conversion should be most-significant-last.

@since WIP
-}
pmostSignificantLast :: forall (s :: S). Term s PEndianness
pmostSignificantLast = pcon $ PEndianness pfalse
```
No need for unsafe

SeungheonOh (dismissed):
LGTM. Thanks

SeungheonOh (approved):

Plutarch/Builtin/ByteString.hs
kozross: Dependency cycle.

PR #786, opened 2024-12-04
Add `PMaybeSoP` and documentation on SoP encoding

t4ccer:


Plutarch/Maybe.hs
```diff
@@ -192,3 +202,53 @@ pmapMaybe = phoistAcyclic $
   plam $ \f mv -> pmatch mv $ \case
     PJust v -> pjust # (f # v)
     PNothing -> pnothing
+
+-- | @since WIP
+data PMaybeSoP (a :: S -> Type) (s :: S)
```
SeungheonOh, line 207: We can just change `PMaybe` to use SOP instead of having multiple types

SeungheonOh:
Also, please add haddock documentations to each derivation strategies on `Plutarch.Repr.*`.

Plutarch/Maybe.hs
t4ccer: I think we can do it in separate PR, confirm that SoP improves PRational and probably drop Scott encoding altogether

kozross (approved):

PR #788, opened 2024-12-08
More PRational optimizations

kozross:
This PR does several things:

* Where possible, avoid using `#<` on `PInteger` over `#<=`, as it turns out it's more efficient
* Try to eliminate redundant operations (of which there are many)
* Refactor some common operations into helpers

This is an improvement, in some cases by quite a bit.

### DIFF
--- a/Plutarch/Rational.hs
+++ b/Plutarch/Rational.hs
@@ -144,7 +144,7 @@ instance Fractional (Term s PRational) where
               ( plet (xn * pto yd) $ \numm ->
                   preduce
                     #$ pif
-                      (denm #< 0)
+                      (denm #<= 0)
                       (pcon $ PRational (pnegate #$ numm) (punsafeCoerce $ pnegate # denm))
                       (pcon $ PRational numm (punsafeCoerce denm))
               )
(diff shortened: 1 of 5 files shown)


Plutarch/Rational.hs
```diff
@@ -340,3 +314,23 @@ pproperFraction = phoistAcyclic $
...
+
+-- Assumes d is positive
+preduce' :: forall (s :: S). Term s (PInteger :--> PInteger :--> PRational)
+preduce' = phoistAcyclic $ plam $ \n d' ->
+  plet d' $ \d ->
+    plet (pgcd # n # d) $ \r ->
+      pif
+        (d #<= 0)
+        (pcon $ PRational (pnegate # (pdiv # n # r)) $ punsafeDowncast (pnegate # (pdiv # d # r)))
+        (pcon $ PRational (pdiv # n # r) $ punsafeDowncast (pdiv # d # r))
```
t4ccer, line 336: > Assumes d is positive

But it doesn't right? It branches and negates if non-positive

kozross: Huh, I am obviously blind. In that case, I can wring out even more.

t4ccer (approved):

PR #790, opened 2024-12-11
New numerical hierarchy

kozross:
This PR replaces both `PNum` and `PIntegral` with a new, more principled hierarchy:

* `PAdditiveSemigroup`, which provides `#+` and `pscalePositive`, allowing us to scale different types by `PPositive` (similar to multiplication)
* `PAdditiveMonoid`, which provides a polymorphic `pzero` as an identity for `#+`
* `PAdditiveGroup`, which provides `#-` and `pnegate`, as well as `pscaleInteger`, allowing us to scale by `PInteger`s (similar to multiplication)
* `PMultiplicativeSemigroup`, which provides `#*` and `ppowPositive`, allowing us to exponentiate (efficiently, using exponentiation-by-squaring) various types by `PPositive`
* `PMultiplicativeMonoid`, which provides a polymorphic `pone` as an identity for `#*`
* `PRing`, which now houses `pfromInteger` (with actual laws this time!)
* `PIntegralDomain`, for `pabs` and `psignum`

These type classes all have laws (but not tests for these yet). As part of these changes, the following was also done:

* `PNum` has been completely removed; its only remaining functionality was to 'bridge' to `Num`, which is now done by way of `PIntegralDomain`
* `PPosixTime` now doesn't support the full slew of arithmetic (as it never made too much sense to). Currently, it is a `PAdditiveSemigroup`, `PAdditiveMonoid` and `PAdditiveGroup`, but nothing else.
* In light of the change to `PPosixTime`, `PIntegral` is now gone too; `pquot`, `prem`, `pdiv` and `pmod` remain as backwards-compatible functions on `PInteger`, which arguably is the only place they were really used.
* BLS12-381 elements are now able to do some numerical stuff: they are now `PAdditiveSemigroup`, `PAdditiveMonoid` and `PAdditiveGroup`, but nothing else.
* `PPositive` is now exported by the Prelude. It's strange how we didn't do this before.

Things considered, but not done:

* `PNatural` fits reasonably well into this hierarchy, but wasn't added as this PR is already big enough
* It is theoretically possible to restore a more general version of `pquot` (or `pdiv`, depending on how you look at it) where the divisor is `PPositive`. This is law-abiding, but unfortunately doesn't _quite_ fit the hierarchy as-current, as BLS12-381 elements can't really have such an operation on them.
* In theory, we could re-generalize Euclidean division by way of this hierarchy, but that raises some annoying questions, and ultimately, we can't really have more than one useful instance.
* Law tests and benchmarks; these should come in a separate PR.

There is one slightly odd regression in present benchmarks, which I am unable to account for or fix.


plutarch-testlib/goldens/rational.bench.golden
```diff
@@ -23,4 +23,4 @@ data.id.2 {"exBudgetCPU":16100,"exBudgetMemory":200,"scriptSizeBytes":12}
 data.id.11/3 {"exBudgetCPU":6235110,"exBudgetMemory":26827,"scriptSizeBytes":239}
 div by 0.1/0 {"exBudgetCPU":278933,"exBudgetMemory":136,"scriptSizeBytes":239}
 div by 0.recip 0 {"exBudgetCPU":187980,"exBudgetMemory":134,"scriptSizeBytes":123}
-div by 0.1/(1-1) {"exBudgetCPU":4724786,"exBudgetMemory":20156,"scriptSizeBytes":286}
+div by 0.1/(1-1) {"exBudgetCPU":4825994,"exBudgetMemory":20158,"scriptSizeBytes":286}
```
t4ccer, line 26: :crying_cat_face: Do we know why?

t4ccer (approved):
LGTM

plutarch-testlib/goldens/rational.bench.golden
kozross: I wish I did, but AFAICT, nothing changed in a way that would trigger _this specific case_.

PR #792, opened 2024-12-19
Fix substitution to work with `UPLC.Case` to unbork SOP encoding

t4ccer:


kozross (approved):

Plutarch/Internal/Term.hs
```diff
@@ -673,10 +673,16 @@ subst idx x (UPLC.Apply () yx yy) = UPLC.Apply () (subst idx x yx) (subst idx x
...
 subst idx x (UPLC.Force () y) = UPLC.Force () (subst idx x y)
-subst idx x (UPLC.Var () (DeBruijn (Index idx'))) | idx == idx' = x idx
-subst idx _ y@(UPLC.Var () (DeBruijn (Index idx'))) | idx > idx' = y
-subst idx _ (UPLC.Var () (DeBruijn (Index idx'))) | idx < idx' = UPLC.Var () (DeBruijn . Index $ idx' - 1)
-subst _ _ y = y
+subst idx x y@(UPLC.Var () (DeBruijn (Index idx'))) =
+  case compare idx idx' of
+    EQ -> x idx
+    GT -> y
+    LT -> UPLC.Var () (DeBruijn . Index $ idx' - 1)
```
SeungheonOh, line 680: This assumes we are erasing a lambda. I don't think we should do this

PR #793, opened 2024-12-19
Convert PMaybe to SOP

kozross:
Changes `PMaybe` to use SOP instead of Scott encoding. This overall doesn't change performance in any one direction: some things get much better, others get slightly worse. Overall the impact seems positive though.


t4ccer (approved):

plutarch-testlib/goldens/list.bench.golden
```diff
@@ -16,8 +16,8 @@ pfoldl.empty-primed {"exBudgetCPU":1033427,"exBudgetMemory":5433,"scriptSizeByte
 elemAt.elemAt_3_[1..10] {"exBudgetCPU":12945655,"exBudgetMemory":56708,"scriptSizeBytes":157}
 elemAt.elemAt_0_[1..10] {"exBudgetCPU":10192885,"exBudgetMemory":43796,"scriptSizeBytes":157}
 elemAt.elemAt_9_[1..10] {"exBudgetCPU":18451195,"exBudgetMemory":82532,"scriptSizeBytes":157}
-find.find_(==3)_[1..4] {"exBudgetCPU":6365468,"exBudgetMemory":29622,"scriptSizeBytes":100}
-find.find_(==5)_[1..4] {"exBudgetCPU":7549850,"exBudgetMemory":36224,"scriptSizeBytes":100}
+find.find_(==3)_[1..4] {"exBudgetCPU":6381468,"exBudgetMemory":29722,"scriptSizeBytes":97}
```
SeungheonOh, line 19: it's somehow more expensive?

t4ccer: Creating `PMaybe` is more expensive if it's SoP encoded cause that's actually doing something vs Scott encoding that's just constructing a lambda but that's good in the actual program cause consuming SoP encoded PMaybe is cheaper so it's still an improvement anyway (you'll always consume it cause at the end you need to collapse to PUnit or perror)

SeungheonOh: oops, should've read what function actually does. I forget `pfind` returns PMaybe

PR #794, opened 2024-12-19
Use SOP more

kozross:
This converts `PEither`, `PPair` and `PList` into SOP encodings. We have some losses in perf, but overall, huge gains.


Plutarch/List.hs
```diff
@@ -58,15 +57,34 @@ import Plutarch.Internal.Term (
...
+-- | @since WIP
+deriving via
+  DeriveAsSOPStruct (PList a)
+  instance
+    PlutusType (PList a)
+
+{-
 instance DerivePlutusType (PList a) where
   type DPTStrat _ = PlutusTypeScott
+-}
```
t4ccer, line 87: Can be removed

SeungheonOh (approved):
Thanks

PR #814, opened 2025-02-09
Restore PLiftable (DeriveAsTag a)

t4ccer:
Depends on https://github.com/Plutonomicon/plutarch-plutus/pull/813


SeungheonOh:
Ah, thank you.

PR #820, opened 2025-02-24
Bump required plutus-core version to 1.40

t4ccer:


SeungheonOh (approved):
This requires major version bump

PR #825, opened 2025-03-26
Add compilation option to set threshold for constant term to be hoisted

SeungheonOh:
`ConstantHoist Nothing` will prevent hoisting anything. `ConstantHoist $ Just xyz` will hoist any terms that are larger than integer `xyz`


Plutarch/Internal/Term.hs
```diff
@@ -320,24 +322,28 @@ we want to trace, and if so, under what log level and mode.
...
       Show
     )
 
+-- | @since WIP
+instance Semigroup Config where
+  (Config a b) <> (Config a' b') = Config (a <> a') (b <> b')
+
+-- | @since WIP
+instance Monoid Config where
+  mempty = Config mempty (Last $ Just 20)
```
t4ccer, line 342: I'm not sure about `Last $ Just 20` here. Now we have `NoTracing <> mempty /= NoTracing`. It'd need to be a bit more elaborate, like `InliningConfig = AlwaysInline | InlineIfSmaller Integer` and when you actually want to extract the config from Last then you can do the default with `fromMaybe (InlineIfSmaller 20) constHoist'config`. Then you can just put `mempty` here to get lawful monoid back.

TLDR; `Nothing` cause `Last` was not set must be different than "no inline threshold"

t4ccer (changes requested):

Plutarch/Internal/Term.hs
SeungheonOh: Ah, yeah. I originally had `mempty` as `Last Nothing`, which would've worked. I'll fixed this

SeungheonOh: I was not feeling like rolling up a whole new type for just this. I just made `consthoist'config :: Last Integer` and made `-1` represent "never hoist"

t4ccer: Magic values :disappointed:

SeungheonOh: ehh I think this is pretty sensible

t4ccer (approved):
I won't block it over the magic `-1`

PR #826, opened 2025-03-26
Fix script prefix in V3 scriptHash

szg251:


t4ccer:
Some goldens depend on it, the easiest way is to nuke goldens folder and re-run `cabal test` from inside `plutarch-testlib` directory. There's also a flag for tasty for that but I don't remember it

t4ccer:
Wow this is so bad

t4ccer (approved):

Issue #851, opened 2025-08-18
Missing `PTryFrom` instances for V3 types


kozross:
These need to be reintroduced. They were removed as part of dealing with `PDataRecord`, but at least one downstream consumer (LambdaBuffers) needs them to exist.

Only instances of the form `PTryFrom PData (PAsData a)` need to be added.

PR #855, opened 2025-08-19
Fix missing PTryFrom instances

szg251:
Closes: #849, #850, #851

### DIFF
--- a/plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap.hs
+++ b/plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap.hs
@@ -165,7 +165,7 @@ instance PEq (PMap 'Sorted k v) where
         Term s (PMap 'Sorted k v :--> PMap 'Sorted k v :--> PBool)
       peqViaData = phoistAcyclic $ plam $ \m0 m1 -> pdata m0 #== pdata m1
 
--- | @since 3.3.1
+-- | @since 3.4.0
 instance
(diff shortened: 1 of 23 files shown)


plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap.hs
```diff
@@ -165,6 +165,40 @@ instance PEq (PMap 'Sorted k v) where
...
+    unwrapped <- tcont . plet $ PPrelude.pmap # ptryFromPair # opq'
+    pure (pdata . pcon . PMap $ unwrapped, ())
+    where
+      ptryFromPair ::
+        forall (s :: S).
+        Term s (PBuiltinPair PData PData :--> PBuiltinPair (PAsData k) (PAsData v))
+      ptryFromPair = plam $ \p ->
+        ppairDataBuiltin
+          # ptryFrom (pfstBuiltin # p) fst
+          # ptryFrom (psndBuiltin # p) fst
```
kozross, line 186: This should probably be an instance for `PBuiltinPair` rather than a helper, but I won't block a merge over it.

szg251: I couldn't implement this because implementing such an instance causes a reduction stack overflow.

kozross: OK, we can leave it for now.

kozross (approved):

Issue #833, opened 2025-04-29
Add tests for `DeriveTagPlutusType` and `DeriveTagPLiftable`


t4ccer:

kozross:
This ticket is a bit unclear. What we actually need to do is verify that derivations of `PlutusType` and `PLiftable` via the `DeriveAsTag` helper newtype work as we expect. In particular:

* The `PlutusType` and `PLiftable` laws hold for such a derivation; and
* The integer values we encode into are in declaration order.

PR #867, opened 2025-09-01
DeriveAsTag PLiftable tests

szg251:
Couldn't figure out how to test PlutusType yet, so I decided to push this as a separate PR. Issue: #833

### DIFF
--- a/plutarch-testlib/Plutarch/Test/Laws.hs
+++ b/plutarch-testlib/Plutarch/Test/Laws.hs
@@ -315,7 +315,10 @@ checkPLiftableLaws =
       plift (pconstant @a x) `prettyEquals` x
   ]
 
--- | @since 3.5.0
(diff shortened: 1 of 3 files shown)


plutarch-testlib/Plutarch/Test/Laws.hs
```diff
@@ -311,6 +313,25 @@ checkPLiftableLaws =
       plift (pconstant @a x) `prettyEquals` x
   ]
 
+class AsTagTest a where
+  tagNum :: a -> Integer
```
kozross, line 317: I see what you were going for there, but I don't think it's the best idea. A type class like this, only for testing purposes, won't work terribly well, because any type someone would want to test would need to be an instance.

It'd be better to do this using `SOP.Generic`, since we need an instance of `SOP.Generic` for Plutarch types anyway.

kozross (changes requested):

PR #871, opened 2025-09-05
Choener/explicit eq no digest

choener:
In this branch, I have replaced the digest-based system of keeping track of `HoistedTerm`s with, essentially, `(Hash HoistedTerm, RawTerm)` and storage in a `HashMap`. Collisions are now handled first checking the cheap hash, then the full term.

In addition, `phoistAcyclic` checks have been made optional (though this should maybe be a runtime option, not a compilation option as is now).

Finally, I added `pfindAllPlaceholders` which is more cheap than repeatedly traversing the tree with `pfindPlaceholder`.

Apart from compilation performance-related changes, I re-added some functions that Djed uses.

===

The second commit changes the plutus-core version to 1.45, which works out fine in all tests and corresponds to mainnet.

### DIFF
--- a/Plutarch/Internal/Term.hs
+++ b/Plutarch/Internal/Term.hs
@@ -1,6 +1,9 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE NoPartialTypeSignatures #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
 
 module Plutarch.Internal.Term (
   -- | \$hoisted
@@ -28,9 +31,6 @@ module Plutarch.Internal.Term (
   compile',
   optimizeTerm,
   ClosedTerm,
-  Dig,
-  hashTerm,
-  hashRawTerm,
   RawTerm (..),
   HoistedTerm (..),
   TermResult (TermResult, getDeps, getTerm),
@@ -68,6 +68,8 @@ import Data.Aeson (
  )
 import Data.ByteString qualified as BS
 import Data.Default (def)
+import Data.HashMap.Strict qualified as HM
+import Data.Hashable (Hashable (..), defaultHashWithSalt)
 import Data.Kind (Type)
 import Data.List (foldl', groupBy, sortOn)
 import Data.Map.Lazy qualified as M
@@ -86,7 +88,9 @@ import PlutusCore (Some (Some), ValueOf (ValueOf))
 import PlutusCore qualified as PLC
 import PlutusCore.DeBruijn (DeBruijn (DeBruijn), Index (Index))
 import Prettyprinter (Pretty (pretty), (<+>))
+import System.IO.Unsafe (unsafePerformIO)
 import UntypedPlutusCore qualified as UPLC
+import UntypedPlutusCore.Transform.Simplifier qualified as UPLC
 
 {- $hoisted
  __Explanation for hoisted terms:__
@@ -103,10 +107,31 @@ import UntypedPlutusCore qualified as UPLC
  though the name is relative to the current level.
 -}
 
-type Dig = Digest Blake2b_160
-
-data HoistedTerm = HoistedTerm Dig RawTerm
+data HoistedTerm = HoistedTerm {htHash :: Int, htRawTerm :: RawTerm}
   deriving stock (Show)
+
+{- | Hoisted term carry their own non-cryptographic hash, making it incredibly
+cheap to hold these in hashmaps.
+-}
+instance Hashable HoistedTerm where
+  hashWithSalt = defaultHashWithSalt
+  {-# INLINE hashWithSalt #-}
+
+  -- The instance uses that this is basically 'Hashed a'.
+  hash = htHash
+  {-# INLINE hash #-}
+
+{- | Equality of hoisted terms is first checked via their non-cryptographic
+hash, then via term equality. Inequality, which should happen much more
+often is thus cheap.
+
+Note that we could just as well derive this instance...
+-}
+instance Eq HoistedTerm where
+  l == r =
+    htHash l == htHash r
+      && htRawTerm l == htRawTerm r
+  {-# INLINE (==) #-}
 
 type UTerm = UPLC.Term UPLC.DeBruijn UPLC.DefaultUni UPLC.DefaultFun ()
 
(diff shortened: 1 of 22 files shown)


Plutarch/List.hs
```diff
@@ -221,3 +240,137 @@ pelemAt' = phoistAcyclic $
       (n #== 0)
       (phead # xs)
       (self # (n - 1) #$ ptail # xs)
+
+{- | Match first N elements from the list. It's is better to use @pmatchList@ if number of elements that needs to be
```
SeungheonOh, line 244: Did you wrote this? I swear I remember writing this myself.

SeungheonOh: Oh, I did: https://github.com/Plutonomicon/plutarch-plutus/pull/824

idk why this got purged. It'd be nice if you can back port some of the golden tests I wrote in this PR back here. Feel free not to do it if you find this too laborious.

Plutarch/List.hs
```diff
@@ -221,3 +240,137 @@ pelemAt' = phoistAcyclic $
...
+  Replicate 0 _ = '[]
+  Replicate x a = a ': Replicate (x - 1) a
+
+-- This is very dangerous, be extra careful when using it
+-- Essentially, this will construct NP of specified structure from the given list of term
+-- There is no type assurances whatsoever and each term will be coerced.
+class UnsafeConstrNP xs where
+  constrNP :: forall a s. [Term s a] -> Maybe (NP (Term s) xs)
+
+instance UnsafeConstrNP xs => UnsafeConstrNP (x ': xs) where
```
SeungheonOh, line 369: I'm not proud of this

plutarch-orphanage/plutarch-orphanage.cabal
```diff
@@ -73,7 +73,7 @@ library
   build-depends:
     , bytestring
     , containers
-    , plutus-core           >=1.51.0.0
+    , plutus-core           >=1.45.0.0
```
SeungheonOh, line 76: What is the reason for downgrading? I remember @kozross bumping version for something he's doing.

Plutarch/Internal/Term.hs
```diff
@@ -1,6 +1,9 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE NoPartialTypeSignatures #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
SeungheonOh, line 6: These need to go

plutarch.cabal
```diff
@@ -8,6 +8,13 @@ license:            MIT
 extra-source-files: README.md
 tested-with:        GHC ==9.6.4
 
+flag check-phoist
+  description:
+    Checks every single phoistAcyclic. This is fairly costly for large scripts!
+
+  default:     False
```
SeungheonOh, line 15: I think it's more sensible to have it turned on by default. 

Also, I'd prefer for this flag to be passed in via config type we already have: https://github.com/Plutonomicon/plutarch-plutus/blob/427e49b0cbfd2ff97cdc91522a09e06edd7a7fa6/Plutarch/Internal/Term.hs#L325

SeungheonOh (changes requested):
Thanks

Plutarch/Internal/Term.hs
kozross: Agreed.

Plutarch/Internal/Term.hs
```diff
@@ -68,6 +68,8 @@ import Data.Aeson (
  )
 import Data.ByteString qualified as BS
 import Data.Default (def)
+import Data.HashMap.Strict qualified as HM
+import Data.Hashable (Hashable (..), defaultHashWithSalt)
```
kozross, line 72: Please explicitly import type class methods.

Plutarch/Internal/Term.hs
```diff
@@ -770,6 +784,7 @@ smallEnoughToInline = \case
   _ -> False
 
 -- The logic is mostly for hoisting
+{-
```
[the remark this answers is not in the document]
kozross: Yeah, we don't want to leave commented code like that.

Plutarch/Internal/TermCont.hs
```diff
@@ -94,3 +96,25 @@ pfindPlaceholder idx x = TermCont $ \f -> Term $ \i -> do
...
+    findPlaceholder (RDelay x) = findPlaceholder x
+    findPlaceholder (RHoisted (HoistedTerm _ x)) = findPlaceholder x
+    findPlaceholder (RPlaceHolder idx) = [idx]
+    findPlaceholder (RConstr _ xs) = foldMap findPlaceholder xs
+    findPlaceholder (RCase x xs) = findPlaceholder x <> foldMap findPlaceholder xs
+    findPlaceholder (RVar _) = []
+    findPlaceholder (RConstant _) = []
+    findPlaceholder (RBuiltin _) = []
+    findPlaceholder (RCompiled _) = []
+    findPlaceholder RError = []
```
kozross, line 118: This would be better as a `where` bind using a lambda-case I think, but I won't insist.

Plutarch/Prelude.hs
```diff
@@ -109,6 +109,10 @@ module Plutarch.Prelude (
   PLiftable (..),
   DeriveDataPLiftable,
   DeriveNewtypePLiftable,
+  DeriveAsDataRec (..),
+  DeriveAsDataStruct (..),
+  DeriveAsSOPRec (..),
+  DeriveAsSOPStruct (..),
```
kozross, line 115: Good catch.

plutarch-orphanage/plutarch-orphanage.cabal
kozross: Yeah, this is _way_ low by current standards.

plutarch.cabal
kozross: Agreed.

kozross (changes requested):
This is a great contribution - thank you! However, there's a few things that need addressing:

* Retarget this PR to `treasury-milestone-2`, as this is where current maintenance work is being done.
* Address all of Seungheon's feedback.
* Some stylistic changes.

Once these are resolved, I'd be happy to merge.

Plutarch/Internal/Term.hs
```diff
@@ -682,23 +689,30 @@ punsafeConstant = punsafeConstantInternal
...
 asClosedRawTerm t = asRawTerm t 0
 
 -- FIXME: Give proper error message when mutually recursive.
 phoistAcyclic :: HasCallStack => ClosedTerm a -> Term s a
+-- TODO: (choener) Instead of just disabling compilation, measure what memoizing hoisted terms brings.
 phoistAcyclic t = Term \_ ->
   asRawTerm t 0 >>= \case
     -- Built-ins are smaller than variable references
     t'@(getTerm -> RBuiltin _) -> pure t'
+#ifdef CHECKPHOIST
```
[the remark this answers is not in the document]
choener: An option to run evalScript (or not) is now part of the InternalConfig, which I also expose via variants of the compile functions.
I went with this to be backward-compatible.

Implemented in the master-with-additions, which will have everything except the change to the Hashmap-based internal compilation pipeline.

Plutarch/List.hs
choener: I have re-added the missing functionality.

plutarch-orphanage/plutarch-orphanage.cabal
choener: I changed everything back to 1.51.0.0.
Is there a particular reason to not want to lower the version? I had chosen 1.45 as what is being used in current mainnet.

plutarch.cabal
choener: This can be ignored. We now have an InternalConfig option, which is indeed much more nice.

choener:
Please find all changes that do not related to the change to the internal compilation pipeline here:
https://github.com/Plutonomicon/plutarch-plutus/pull/874
I tried to keep the commits modular in have only "uncontroversial" things in there.

choener:
> @choener - I've merged PR 874. Can this one be closed, or are there other changes in here that stack atop it?

@kozross
@SeungheonOh
There is an important commit in this PR. The previous `compile'` pipeline and `HoistedTerm` digest are replaced with a HashMap-based system.
This provides, for large scripts, another huge improvement in compilation times. I separated it out for two reasons:

- Changing the whole compilation pipeline should receive additional scrutiny. Though both the Plutarch tests all pass, and we use this in production where it passes our tests.
- The second part is the observed benchmark regression for the `intervalutils` test. Though I am now fairly certain that this is an artifact of the way terms are now ordered (as previously they were essentially unordered due to the use of a digest function). This can be seen in the UPLC, but I'd be grateful for a quick from your side.

plutarch-orphanage/plutarch-orphanage.cabal
kozross: There's an upcoming hardfork, and we want to track Plutus versions as close as we can to current where possible. This isn't actually the most current version, so it'll likely bump even higher.

Plutarch/Internal/Term.hs
```diff
@@ -102,9 +95,34 @@ import UntypedPlutusCore qualified as UPLC
...
+
+Note that we could just as well derive this instance...
+-}
+instance Eq HoistedTerm where
+  l == r =
+    htHash l == htHash r
+      && htRawTerm l == htRawTerm r
+  {-# INLINE (==) #-}
+
+type UTerm = UPLC.Term UPLC.DeBruijn UPLC.DefaultUni UPLC.DefaultFun ()
```
kozross, line 124: I'd rather keep this expanded in places we need it. Yes, it's longer, but I'd rather _know_ that we're referring to some horrid upstream pile and not potentially a type wrapper which I have to retain additional context about.

kozross (changes requested):

PR #874, opened 2025-09-09
Re-adds plistMatch and allows disabling intermediate eval calls in phoistAcyclic

choener:
* Re-adds plistMatch
* Allows disabling the (costly) intermediate eval calls in phoistAcyclic
* Layered the individual commits to (try to) simplify cherry-picking, if needed.


kozross:
Thanks for the contribution!

PR #889, opened 2025-11-12
Deprecate the old style of derivations

### LINKED ISSUE #885: Deprecate `Plutarch.DataRepr` and the `DPT` mechanism
These have been entirely superceded at this point, and removing these from the codebase should significantly reduce confusion and improve maintanability.

kozross:
Closes #885 . Also removes an unnecessary module `Plutarch.Repr.Newtype`, as all this does is duplicate functionality already found in `Plutarch.Internal.PlutusType`.


errfrom (changes requested):
`DeriveNewtypePlutusType` is declared identically in the `Plutarch.Repr.Newtype` and `Plutarch.Internal.PlutusType` modules. I think we should address this duplication before merging

errfrom (approved):
Looks good!

PR #894, opened 2025-11-14
Deprecate and replace `pfstBuiltin` and `psndBuiltin`

### LINKED ISSUE #893: Replace all uses of `pfstBuiltin` and `psndBuiltin` with pattern matching
Given that we have pattern matching directly on pairs now (and optimized to use `Case` at that), using these is no longer optimal and should be fixed.

kozross:
Closes #893 . This improves performance everywhere except the `fields` benchmarks, and those are part of a subsystem due for removal anyway (and are deprecated in #889 anyway).


plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap/Zip.hs
```diff
@@ -175,17 +175,19 @@ zipWorker mergeHandler =
...
               # mapR
           HandleOrDropOne handler ->
             pmapDropNothing
               # plam
                 ( \entry -> P.do
-                    k <- plet $ pfstBuiltin # entry
-                    v <- plet $ psndBuiltin # entry
-                    pmapMaybe # plam (\x -> ppairDataBuiltin # k # x) # (handler # k # v)
+                    PBuiltinPair x y <- pmatch entry
+                    k <- plet x
```
errfrom: Why is this `plet` necessary after a `match`? Could we pre-evaluate the bindings instead?

plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap/Zip.hs
```diff
@@ -175,17 +175,19 @@ zipWorker mergeHandler =
             PPrelude.pmap
               # plam
                 ( \entry -> P.do
-                    k <- plet $ pfstBuiltin # entry
-                    ppairDataBuiltin # k #$ handler # k #$ psndBuiltin # entry
+                    PBuiltinPair x y <- pmatch entry
+                    k <- plet x
```
errfrom: Should we also add `v <- plet y` here?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 * Bump `plutus-core` to 1.54.0.0
 * `pif'` is now deprecated, as builtin casing on `Bool` outperforms it
+* `pheadBuiltin` and `ptailBuiltin` are now deprecated, as pattern matching on
```
errfrom:
```suggestion
* `pfstBuiltin` and `psndBuiltin` are now deprecated, as pattern matching on
```

Plutarch/Internal/TryFrom.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
+-- Note (Koz, 14/11/2025: Needed for the PIsData constraints on the PBuiltinPair
```
errfrom: Missing closing parenthesis in the note

plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
```diff
@@ -582,10 +582,12 @@ plovelaceValueOf = phoistAcyclic $
...
     pmatch (pto $ pto $ pto value) $ \case
       PNil -> 0
-      PCons x _ ->
+      PCons x _ -> pmatch x $ \(PBuiltinPair xFst xSnd) ->
         pif
-          (pfstBuiltin # x #== padaSymbolData)
-          (pfromData (psndBuiltin #$ phead #$ pto $ pto $ pfromData $ psndBuiltin # x))
+          (xFst #== padaSymbolData)
+          ( pmatch (phead #$ pto $ pto $ pfromData xSnd) $ \(PBuiltinPair _ thing) ->
+              pfromData thing
```
errfrom:
```suggestion
          ( pmatch (phead #$ pto $ pto $ pfromData xSnd) $ \(PBuiltinPair _ quantity) ->
              pfromData quantity
```

plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
```diff
@@ -602,7 +604,14 @@ pisAdaOnlyValue = phoistAcyclic $
...
     pmatch (pto $ pto $ pto value) $ \case
       PNil -> pcon PTrue
-      PCons x xs -> pand' # (pnull # xs) # (pfstBuiltin # x #== padaSymbolData)
+      PCons x xs ->
+        pand'
+          # (pnull # xs)
+          # pmatch
+            x
+            ( \(PBuiltinPair thing _) ->
+                thing #== padaSymbolData
```
errfrom:
```suggestion
            ( \(PBuiltinPair cs _) ->
                cs #== padaSymbolData
```

plutarch-ledger-api/src/Plutarch/LedgerApi/Value.hs
```diff
@@ -629,9 +638,9 @@ pinsertAdaEntry =
     plam $ \value ->
       pmatch (pto $ pto $ pto value) $ \case
         PNil -> psingletonSortedValue # padaSymbol # padaToken # 0
-        PCons x xs ->
+        PCons x xs -> pmatch x $ \(PBuiltinPair xFst _) ->
           pif
-            (pfstBuiltin # x #== padaSymbolData)
+            (xFst #== padaSymbolData)
```
errfrom:
```suggestion
        PCons x xs -> pmatch x $ \(PBuiltinPair cs _) ->
pif
            (cs #== padaSymbolData)
```

errfrom (changes requested):

plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap/Zip.hs
kozross: Turns out it isn't! The reason for this is because `pmatch` against a `PBuiltinPair` uses the `Case` construct of `Term`, which effectively means that `k <- plet x` here is performing a lambda transform for no reason, as `x` is already a variable in the generated code. Going to fix this.

plutarch-ledger-api/src/Plutarch/LedgerApi/AssocMap/Zip.hs
kozross: No, and even this `plet` is unnecessary, as per my comment above.

errfrom (approved):
LGTM!

PR #897, opened 2025-11-17
Integer casing to improve PValidateData on DeriveAsTag

### LINKED ISSUE #891: Use casing on `Integer` to optimize derived `PValidateData`
Follow-on from #888 . We can use a similar trick for `PValidateData` to improve its performance. Given that we over 3Xed exunits and 2Xed memory use, this might be worth it.

kozross:
Closes #891.


Plutarch/Internal/Parse.hs
```diff
@@ -222,17 +224,12 @@ instance PValidateData a => PValidateData (PAsData a) where
...
-            perror
-            ( pif
-                (i #>= plen)
-                perror
-                x
-            )
+     in Term $ \level -> do
+          TermResult rawI depsI <- asRawTerm (pasInt # opq) level
+          TermResult rawX depsX <- asRawTerm x level
+          pure . TermResult (RCase rawI . replicate len $ rawX) $ depsI <> depsX
```
errfrom, line 234: Is it safe / performant for data types with many constructors?

kozross: Good point. I'll write some extra benchmarks for this.

kozross:
```
pparseData 1 {"exBudgetCPU":195442,"exBudgetMemory":1096,"scriptSizeBytes":15}
pparseData 2 {"exBudgetCPU":195442,"exBudgetMemory":1096,"scriptSizeBytes":20}
pparseData 3 {"exBudgetCPU":195442,"exBudgetMemory":1096,"scriptSizeBytes":24}
pparseData 4 {"exBudgetCPU":195442,"exBudgetMemory":1096,"scriptSizeBytes":29}
pparseData 5 {"exBudgetCPU":195442,"exBudgetMemory":1096,"scriptSizeBytes":33}

```

This shows basically:

* Exunit and memory costs are identical no matter how many 'arms' the enum type has
* Script size increases linearly with the case count (unsurprisingly)

Given the scale of the improvement relative using a pair of `pif`s with two builtin calls, I think we should still keep this.

errfrom (approved):

PR #930, opened 2025-12-11
PShow instance for PNatural, expose PShow method

kozross:
This adds an instance of `PShow` for `PNatural`, which for some reason didn't have one before. We also expose the `pshow'` method of `PShow` from the Plutarch prelude.


errfrom (approved):
Looks good

PR #931, opened 2026-01-07
Fix typoes in PInterval fields

kozross:
This fixes the (rather embarrassing) typoes in the field names for `PInterval`.

### DIFF
--- a/plutarch-ledger-api/src/Plutarch/LedgerApi/Interval.hs
+++ b/plutarch-ledger-api/src/Plutarch/LedgerApi/Interval.hs
@@ -41,8 +41,10 @@ import PlutusLedgerApi.V3 qualified as Plutus
 
 -- | @since 2.0.0
 data PInterval (a :: S -> Type) (s :: S) = PInterval
-  { pinteral'from :: Term s (PLowerBound a)
-  , pinteral'to :: Term s (PUpperBound a)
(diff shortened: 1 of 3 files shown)


plutarch-ledger-api/src/Plutarch/LedgerApi/Interval.hs
```diff
@@ -41,8 +41,10 @@ import PlutusLedgerApi.V3 qualified as Plutus
 
 -- | @since 2.0.0
 data PInterval (a :: S -> Type) (s :: S) = PInterval
-  { pinteral'from :: Term s (PLowerBound a)
-  , pinteral'to :: Term s (PUpperBound a)
+  { pinterval'from :: Term s (PLowerBound a)
+  -- ^ @since 3.6.0
+  , pinterval'to :: Term s (PUpperBound a)
+  -- ^ @since 3.6.0
```
errfrom: @kozross Should we open an issue to update all remaining `@since wip` annotations as well?

errfrom (approved):

plutarch-ledger-api/src/Plutarch/LedgerApi/Interval.hs
kozross: That's actually a mistake on my part - I should only have done this as part of release prep. I'll fix.

PR #935, opened 2026-04-02
Treasury milestone 3

### LINKED ISSUE #882: Release prep for Treasury Milestone 3
This includes:

* Video showing changes (if any)
* Changelogging
* Merge to `master`
* Documentation changes or fixes

kozross:
This acts as a release for Milestone 3. Closes #882.


plutarch-docs/src/Concepts/Term.md
```diff
@@ -0,0 +1,106 @@
...
+## Prerequisites
+
+[TODO: Fill in]
+
+## Anatomy of a `Term`
+
+[TODO: Lead-in with `Term`'s tyvars]
+
+As Plutarch is an eDSL for generating UPLC scripts, we fundamentally need a
+_code generation monad_ to be available. This effectively acts as an
```
gnumonik: Ok so this makes sense *to me* because you've already explained to me how "term is morally a monad", but I think this might be somewhat misleading to someone who hasn't received that explanation. 

Unless I missed the instance, `Term` isn't *actually* a monad so I think if you're going to say that it effectively *is* the term generation monad, you need to give enough of that explanation for that to make sense.

plutarch-docs/src/Concepts/Term.md
```diff
@@ -0,0 +1,106 @@
...
+anything in-between. [TODO: More?]
+
+`Term` fills this role in Plutarch. Put simply, `Term s a` is a description of
+a UPLC computation which, when executed[^1], does one of the following:
+
+1. Crashes with an error.
+2. Succeeds producing the UPLC representation of `a`.
+
+Thus, whenever you see a `Term`, you should imagine it as a segment of UPLC
+code. The nearest analogy to this in 'basic' Haskell would be the `Q` monad used
```
gnumonik: I am not sure that "It's like the Q Monad, kind of" is all that helpful without a bit more explanation. (I'm _actually_ not sure, I would think that even many experienced Haskellers, even ones who have used some TH here or there, may not have a great intuition for what `Q` actually is / how it works / why it's needed. I could be wrong though.)

plutarch-docs/src/Concepts/Term.md
```diff
@@ -0,0 +1,106 @@
...
+for Template Haskell, although this is only a conceptual similarity[^2].
+
+## The difference between `->` and `:-->` with `Term`
```
gnumonik: Personally I would have, in either its own section or somewhere in here, a more gentle introduction to the distinction between "Haskell-level" and "Plutarch-level". This section is part of that explanation, I suppose, but I think the relevant bits are a bit _too_ technically correct (esp lines 45-50, there are a lot of parenthetical qualifiers there which I think make it hard to follow). 

I'd start by saying something that's incomplete but communicates the gist, like: 

"The regular function arrow `->` works how you would expect: A function of type `Term s a -> Term s b` takes a term and produces a new term. You can view Haskell functions over terms as combinators for constructing more complex terms out of simpler component terms. The Plutarch-level function arrow `:--:>`, by contrast, indicates a UPLC-level function, which is itself a complete term. You cannot apply a Plutarch-level function directly, but you can construct a term that represents the function applied to some (Plutarch-level) argument by using the Plutarch application combinator `#`." 

I know it seems really basic but something like that would have probably helped me and certainly would have helped some of the other people I had to explain Plutarch to.

plutarch-docs/src/Introduction/DelayAndForce.md
```diff
@@ -3,6 +3,7 @@
 <p>
 
 ```haskell
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
gnumonik, line 6: Maybe note somewhere what this is for so we can know whether it's ever safe to remove in the future?

plutarch-docs/src/Typeclasses/PlutusType.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Purpose of the class
+
+## Laws
+
+## Interface
+
+The two 'directions' of `PlutusType` give rise to two parallel interfaces: one
+to sequence and lift computations in `Term`s, the other to convert from `a` to
+`PInner a` at (essentially) no cost. For convenience, there are helpers
```
gnumonik: I assume that this will be covered in some of the not-yet-written stuff, but it would probably be helpful to explain why `PInner a` is needed/useful in the first place. It's not very helpful to know that `PlutusType` is an interface for converting between `a` and `PInner a` if you don't have any idea why you'd ever want/need to do that.

plutarch-ledger-api/src/Plutarch/LedgerApi/Value/TokenName.hs
```diff
@@ -52,6 +52,19 @@ instance PLiftable PTokenName where
...
+'PByteString' must not exceed 32 bytes.
+
+@since 3.6.0
+-}
+instance PValidateData PTokenName where
+  pwithValidated opq x =
+    plet (plengthBS #$ pasByteStr # opq) $ \bsSize ->
+      pif (bsSize #<= ptokenNameByteSizeLimit) x perror
+
+ptokenNameByteSizeLimit :: forall (s :: S). Term s PInteger
```
gnumonik, line 65: Can this ever actually change? If not, why not hardcode the value when it's needed?

CHANGELOG.md
```diff
@@ -1,6 +1,50 @@
...
+* `PShow` instance for `PNatural`
+* `punsafeCase` is now exported from `Plutarch.Unsafe`.
+
+## Changed
+
+* Bump `plutus-core` to 1.60.0.0
+* Add `KnownBuiltinType` constraint for various `PLiftable` instances as per the
+  new requirements from `plutus-core`. This led to some redundant `Includes`
+  constraints being removed as well.
+* `pif'` is now deprecated, as builtin casing on `Bool` outperforms it
```
gnumonik, line 23: Why deprecate it instead of changing it to use the builtin case on `Bool` under the hood?

gnumonik (changes requested):
Mostly comments on the docs, a few small things aside from that I had questions about. Code-wise it's fine as far as I can tell. I didn't review the goldens.

CHANGELOG.md
kozross: Mostly back-compatibility. If people expect strictness, they'd be _very_ surprised if it was non-strict. We're likely gonna remove it later anyway.

plutarch-ledger-api/src/Plutarch/LedgerApi/Value/TokenName.hs
kozross: I _believe_ it's a parameter set by voting or something?

plutarch-docs/src/Introduction/DelayAndForce.md
kozross: It's for the `PCon` and `PMatch` type classes, which are now (deprecated) synonyms. They should be removed, but this whole thing needs rewriting anyway.

gnumonik (approved):
Most of my comments ended up being irrelevant since addressed at doc stuff that shouldn't have been there, things should be good now.

PR #944, opened 2026-04-06
CIP-133 primitive bindings

### LINKED ISSUE #942: Support for CIP-133 multi-scalar multiplication for BLS primitives
This would require adding a 'primitive binding' to the relevant builtins. It can't really impact any existing type class instances (being a niche operation).

kozross:
Closes #942.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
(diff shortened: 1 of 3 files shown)


gnumonik (approved):
There's barely anything that _could_ even be wrong here, approved.

PR #945, opened 2026-04-07
PBuiltinValue

### LINKED ISSUE #943: Add bindings to new `Value` type and primops
These will be similar to their equivalents in `Plutarch.Builtin`.

kozross:
Closes #943 . Unfortunately, due to the somewhat odd set of primitive operations, only `PlutusType`, `PLiftable` and `PIsData` instances made sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # UNRELEASED
 
+## Added
+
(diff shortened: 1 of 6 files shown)


gnumonik (approved):
I asked you the only question I had on Slack. In light of your response everything here is fine (I mean modulo the fact that the BuiltinValue stuff itself seems dubiously useful at best).

PR #947, opened 2026-04-08
'Connect up' `PBuiltinValue` with existing ledger types

### LINKED ISSUE #946: Add conversions from existing `PValue` subtypes to new `PBuiltinValue`
In general, a script receives any `Value`-esque thing as a `Data` encoding, whether it's from an input or the `ScriptContext`. These currently have a range of data types that describe them:

* `PRawValue`
* `PSortedValue`
* `PLedgerValue`
* `PMintValue` (which is different in V1 and V3!)

Each of these varies by way of what they guarantee: whether we allow zero amounts, whether they're sorted and deduplicated, etc. Constructing a `PBuiltinValue` at runtime is typically done using `punValueData`, which itself has specific expectations. Thus, the only 'safe' conversion is `PMintValue`, as all the others might fail.

To help with this, we need to provide at least a conversion of type `Term s (PAsData PMintValue) -> PBuiltinValue` that uses `pforgetData` together with `punValueData`. We also need to document _quite specifically_ what we can expect if we try to (unsafely) convert any other `P*Value`. We might (or might not) want to provide helper conversions for these.

kozross:
This adds some interface logic between `PBuiltinValue` and existing ledger types. The interface is fairly conservative, but if we end up needing more, we can always add more. There's also some helpers for making `PBuiltinValue` a bit nicer to work with.

Closes #946 , depends on #943 .


plutarch-testlib/test/Plutarch/Test/Suite/PlutarchLedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,60 @@
...
+import PlutusLedgerApi.V3.Orphans ()
+import Test.QuickCheck (Property, arbitrary, forAllShrinkShow, shrink)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.QuickCheck (testProperty)
+
+tests :: TestTree
+tests =
+  testGroup
+    "PBuiltinValue"
+    [ testProperty "PMintValues convert to PBuiltinValues without error" propMintToBuiltin
```
errfrom, line 65: A tiny suggestion: we should explicitly mention that this only applies to V3 PMintValues to reduce cognitive load for contributors.
```suggestion
    [ testProperty "PMintValues (V3) convert to PBuiltinValues without error" propMintToBuiltin
```

errfrom (approved):
Looks good

I think we should add some tests for the `PBuiltinValue` functionality itself (not only the conversion to and from ledger Value types). For example, to ensure that `plookupCoin` behaves as expected, etc.

Otherwise, I like the interface - it's simple and intuitive.

I'd also personally like to see the "verify first then convert" functions we discussed earlier. They may not be generally efficient, but I can imagine that for certain use cases it will be safer and more time-efficient to have these functions rather than having to reinvent them myself.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,428 @@
...
+@since wip
+-}
+pfromRawValue ::
+  forall (r :: S -> Type) (s :: S).
+  Term s PRawValue ->
+  Term s r ->
+  Term s (PBuiltinValue :--> r) ->
+  Term s r
+pfromRawValue x whenFail whenSucceed =
+  pif (pcanBeBuiltinValue # x) whenFail (whenSucceed #$ punValueData # pforgetData (pdata x))
```
errfrom: The `pif` arguments seem to be flipped, or am I missing something?
```suggestion
  pif (pcanBeBuiltinValue # x) (whenSucceed #$ punValueData # pforgetData (pdata x)) whenFail
```

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,428 @@
...
+@since wip
+-}
+pfromSortedValue ::
+  forall (r :: S -> Type) (s :: S).
+  Term s PSortedValue ->
+  Term s r ->
+  Term s (PBuiltinValue :--> r) ->
+  Term s r
+pfromSortedValue x whenFail whenSucceed =
+  pif (pnoZeroAmounts # x) whenFail (whenSucceed #$ punValueData # pforgetData (pdata x))
```
errfrom: here as well

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,428 @@
...
+
+{- | As 'punsafeFromRawValue', except for 'PSortedValue's instead. The same
+caveats apply.
+
+@since wip
+-}
+punsafeFromSortedValue :: forall (s :: S). Term s (PAsData PRawValue) -> Term s PBuiltinValue
```
errfrom: But it accepts `PRawValue`, not `PSortedValue`

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,428 @@
...
+@since wip
+-}
+ptoLedgerValue ::
+  forall (r :: S -> Type) (s :: S).
+  Term s PBuiltinValue ->
+  Term s r ->
+  Term s (PAsData PLedgerValue :--> r) ->
+  Term s r
+ptoLedgerValue v whenInvalid whenValid =
+  pif (plovelaceValueOf # v #== 0) whenInvalid (whenValid # punsafeCoerce (pvalueData # v))
```
errfrom, line 226: I think it'd be better to add a zero ADA entry instead

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -0,0 +1,428 @@
...
+
+@since wip
+-}
+punsafeToLedgerValue ::
+  forall (s :: S).
+  Term s PBuiltinValue -> Term s (PAsData PLedgerValue)
+punsafeToLedgerValue v = punsafeCoerce $ pvalueData # v
+
+{- | Convert a 'PBuiltinValue' into a 'PAsData' 'PLedgerValue' without checking
+anything. This will error if the input contains an Ada entry.
```
errfrom: > This will error if the input contains an Ada entry.

I don't think it will error immediately, just the type assumptions will be broken - we're not performing any validation here

errfrom (changes requested):

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: Yes, you are correct. Good catch!

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: Again, good catch, thank you!

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: Fixed.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: I think both 'do something different' and 'add a zero and carry on' should be options. Will make another function with the behaviour you describe. The reason I didn't do this by default is that we cannot add zero entries using the `PBuiltinValue` interface (as this does nothing), so it would have to be converted, _then_ added, and I don't think this is a choice that should be made implicitly without an alternative.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: Good point. Fixed both.

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
```diff
@@ -219,6 +225,24 @@ ptoLedgerValue ::
...
+
+@since wip
+-}
+ptoLedgerValue' ::
+  forall (s :: S).
+  Term s PBuiltinValue ->
+  Term s (PAsData PLedgerValue)
+ptoLedgerValue' v = plet (punsafeCoerce (pvalueData # v)) $ \asLedgerValue ->
+  pif
+    (plovelaceValueOf # v #== 0)
```
errfrom: `pinsertAdaEntry` already checks internally whether the `PSortedValue` has an ADA entry, so this check seems redundant

errfrom (approved):
We should add more tests in the future, but otherwise it looks good to me

plutarch-ledger-api/src/Plutarch/LedgerApi/V3/Value.hs
kozross: Good point, will remove it.

PR #949, opened 2026-04-09
Use Name for prettyprinting

### LINKED ISSUE #948: Change `printScript` to use `layoutSmart` instead of `Show`
`show` defaults to `layoutPretty`, which isn't optimal for the kind of output we expect. `layoutSmart` is designed specifically [to deal with these kinds of issues](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:layoutSmart), and we should really be using it.

kozross:
Closes #948 . This also affects other issues, as it seems that named DeBruijn and nameless DeBruijn are both equally unreadable. We should use `Name`s instead.


gnumonik (approved):
Seems alright as a temporary hack. If I understand your examples on slack, it is definitely an improvement over what we have.

PR #952, opened 2026-04-14
Change Term to be a proper ReaderT stack

### LINKED ISSUE #951: Modify `Term` to be more explicit about its use of a transformer stack
Currently, this is done implicitly, which isn't very amenable to refactoring or future extensions. Given our goal to switch to `Name`s, this is something we _will_ need to do, so we should set ourselves up for success as much as possible.

kozross:
Closes #951 .


gnumonik (approved):
Aside from the thing I asked you about on slack this looks good.

PR #961, opened 2026-04-25
Sean/938

### LINKED ISSUE #938: Explicit `let` in `Term`
One big reason why `Term` prettyprints are so illegible is that 

```
plet foo $ \x -> bar
```

is rewritten to the same thing as 

```
(plam $ \x -> bar) # foo
```

To make matters _more_ confusing, in some cases, `plet` performs direct inlining. This is even more confusing, and technically the wrong place for this to happen: optimization should happen during compilation, not during code construction!

While we won't change our compilation strategy for `plet`, having `let` be explicit in `Term`s would allow for much clearer prettyprinting from the point of view of intent, as well as 'pushing' optimization to the right time. Together with #939 , we could also 'mark' `let`-bound variables differently to regular function arguments, which would greatly aid readability of prettyprinted `Term`s.

gnumonik:
This should close #938 modulo small golden/benchmark mismatches. See the slack channel for a more detailed explanation of why those mismatches remain & why I think it's not really possible to ameliorate them. 

Marking this as a draft for now because need to confer with @kozross about the goldens. 

@kozross I'm tagging you for review only so you can have a quick look on (your) Monday if you start before me. There's no need for a thorough review yet, just a sanity check is fine.

### DIFF
--- a/Plutarch/Internal/Term.hs
+++ b/Plutarch/Internal/Term.hs
@@ -141,9 +141,8 @@ data RawTerm
   | RConstr Word64 [RawTerm]
   | RCase RawTerm [RawTerm]
   | -- @since WIP
-    -- Left is a plam'd function, right is an inlined result (we shouldn't need the fn to pretty print since it's already been applied, I think? ugh)
-    -- We need to avoid looking at the fn part if we don't need to, it seems that's what breaks things unexpectedly
-    RLet RawTerm (Either RawTerm RawTerm)
+    -- Let x (\x' -> ...)
+    RLet RawTerm RawTerm
   deriving stock (Show, Eq)
 
 -- | A very cheap hash which cheapens equality, but is also needed for using an unordered container.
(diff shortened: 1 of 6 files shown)


Plutarch/Internal/Term.hs
```diff
@@ -83,6 +83,8 @@ import PlutusCore.DeBruijn (DeBruijn (DeBruijn), Index (Index))
 import Prettyprinter (Pretty (pretty), (<+>))
 import UntypedPlutusCore qualified as UPLC
 
+-- import Debug.Trace (traceM)
+
```
kozross: Remove these.

Plutarch/Internal/Term.hs
```diff
@@ -138,6 +140,9 @@ data RawTerm
   | RPlaceHolder Integer
   | RConstr Word64 [RawTerm]
   | RCase RawTerm [RawTerm]
+  | -- @since WIP
+    -- Let x (\x' -> ...)
```
kozross: The `@since` isn't needed here, because this is internal stuff anyway. Even if it was, this isn't a Haddock, so it wouldn't render.

Plutarch/Internal/Term.hs
```diff
@@ -650,6 +659,29 @@ plet v f = Term $ do
       RHoisted _ -> asRawTerm (f v)
       -- Do the lambda transform as normal
       _ -> asRawTerm (papp (plam' f) v)
+-}
```
kozross: We can remove this whole block I think. Also, we want the docs flush with the signature line.

Plutarch/TermCont.hs
```diff
@@ -7,7 +7,6 @@ module Plutarch.TermCont (
   TC.runTermCont,
   TC.unTermCont,
   TC.tcont,
-  TC.pfindPlaceholder,
```
kozross, line 10: This probably needs a `CHANGELOG` note under `Removed` for the `Unreleased` section, as it's technically part of the public API.

kozross (changes requested):

kozross (approved):

PR #964, opened 2026-05-11
Remove Plutarch deprecations

### LINKED ISSUE #962: Remove deprecations in Plutarch
Currently, the following identifiers are marked as deprecated:

* `pif'`
* `pfstBuiltin`
* `psndBuiltin`
* `PlutusTypeStrat`
* `DerivePlutusType`
* `pletFieldsC`

`pif'`, `pfstBuiltin` and `psndBuiltin` are no longer useful now that direct matching on `PBool` and pairs is possible. `PlutusTypeStrat` and `DerivePlutusType` supported old-style derivations, and the correct mechanism has already existed for several releases, and should be used instead. `pletFieldsC` once again relates to the old `PDataRecord` mechanism and thus also should no longer be used.

Furthermore, several modules are marked deprecated, as they entirely relate to the old-style derivations:

* `Plutarch.DataRepr`
* `Plutarch.DataRepr.Internal`
* `Plutarch.DataRepr.Internal.Field`
* `Plutarch.Internal.Newtype`
* `Plutarch.Internal.ScottEncoding`

These should all be removed.

kozross:
Closes #962 . Some documentation changes were also required.


plutarch-docs/src/Typeclasses/PTryFrom.md
```diff
@@ -35,41 +34,7 @@ recoverListFromPData = unTermCont . fmap fst . tcont . ptryFrom @(PAsData (PBuil
...
-recoverAB :: Term s (PAsData PAB)
-recoverAB = unTermCont $ fst <$> tcont (ptryFrom sampleABdata)
-
-```
-
-> Note: There are other valid implementations for recovering your datatype from `PData`, in some cases you might,
-> for example, want to include additional checks, think making sure that some `PNatural` is indeed positive.
-> In this case you will have to hand-roll the implementation of `PTryFrom`. For some examples, see `plutarch-test`'s 
-> `PTryFromSpec.hs`
+[TODO: Fill in]
```
gnumonik, line 37: Is this _supposed_ to be here or did you mean to fill it in for this PR? I guess it's fine if you intend to clean up later but we want to make sure that we have something here because `PTryFrom` is not the most straightforward type class.

CHANGELOG.md
```diff
@@ -11,9 +11,14 @@
 
 * `Term` prettyprinter now uses `Name`s instead of `DeBruijn`s for readability
 
-## Removed 
+## Removed
 
-* Removed `pfindPlaceholder` from `Internal.TermCont`. It isn't used for anything (though the `findAllPlaceholders` version is)
+* `pif'`, as there's no reason to use it any more
```
gnumonik, line 16: I'm confused. You say you removed it but it's still there in some of the source files, and you refer to it in some of the docs?

If you changed it then say how.

plutarch-docs/src/Typeclasses/PTryFrom.md
kozross: This would require a rewrite that's substantial enough to be its own PR. Furthermore, `PTryFrom` is headed for deprecation, so spending a lot of effort here wouldn't really be that worthwhile.

CHANGELOG.md
kozross: I'm unsure which source files still use `pif'`, or which docs reference it. If so, those should be gone too.

PR #965, opened 2026-05-11
Remove plutarch-ledger-api deprecations

### LINKED ISSUE #963: Remove deprecations in `plutarch-ledger-api`
Two versions of `passertSorted` are marked deprecated, as their behaviour is not obvious. They have been replaced with more usefully-named functions, and thus, we no longer need to keep them around.

kozross:
Closes #963


gnumonik (approved):
Too small for anything other than a "looks good to me"

PR #967, opened 2026-05-18
Refactor stack for `Term`

### LINKED ISSUE #966: Change `Term` stack to `RWST` instead of `ReaderT`
Useful step leading to #939 .

kozross:
Closes #966 . This also reorders the `Except` from the inside to the outside: the stack now is `ExceptT` over `RWS`.

### DIFF
--- a/Plutarch/Internal/Term.hs
+++ b/Plutarch/Internal/Term.hs
@@ -44,14 +44,9 @@ module Plutarch.Internal.Term (
   (#$),
 ) where
 
-import Control.Monad.Reader (
-  ReaderT,
-  ask,
-  asks,
-  lift,
-  local,
-  runReaderT,
- )
+import Control.Monad.Except (Except, runExcept, throwError)
+import Control.Monad.RWS.CPS (RWST, evalRWST)
+import Control.Monad.Reader (ask, asks, lift, local)
 import Control.Monad.State.Strict (evalStateT)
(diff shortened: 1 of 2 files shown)


Plutarch/Internal/Term.hs
```diff
@@ -435,7 +434,7 @@ data TermEnv = TermEnv
  instantiated with.
 -}
 newtype Term (s :: S) (a :: S -> Type) = Term
-  { asRawTerm :: ReaderT TermEnv (Either Text) TermResult
+  { asRawTerm :: ExceptT Text (RWS TermEnv () ()) TermResult
```
gnumonik, line 437: Can we do better than `Text` for the error type? I get that we probably can't do better NOW but could we do better _eventually_? If we can maybe add a note so we remember that we can.

gnumonik (approved):
I think this is fine, though I am curious as to whether we can have a better error type (or whether that'd be useful)

Plutarch/Internal/Term.hs
kozross: We could, but there wouldn't be much point. The main reason why we have that `ExceptT Text` there is CTE errors from UPLC evaluation, and those errors _must_ be `Text`. Thus, there's not much of anything useful we could derive here structurally without parsing Plutus error messages.

Issue #939, opened 2026-04-05
Switch to `Name` over `DeBruijn` in `RawTerm`


kozross:
Currently, Plutarch's compilation strategy targets 'nameless' DeBruijn. This representation is required for compilation, as `serialiseUPLC` expects it. However, it is _very_ unhelpful from the perspective of prettyprinting.

Consider the following sample Plutarch:

```haskell
pcon @(PPair PInteger PString) (PPair 42 "Hello") #== pcon (PPair 42 "Hello")
```

When using 'nameless' DeBruijns, the UPLC this compiles to prettyprints as follows:

```
(\!0 ->
      (\!0 !0 ->
         case
           !2
           [ (\!0 !0 ->
                case
                  !3
                  [ (\!0 !0 ->
                       force
                         ((\!0 !0 -> delay (case !2 [!2, (force !1)]))
                            (equalsInteger !4 !2)
                            (delay (equalsString !3 !1)))) ]) ])
        (constr 0 [42, !1])
        (constr 0 [42, !1]))
     "Hello"
```

This is difficult to read, due to an intersection of two issues:

* The UPLC prettyprinter [coalesces adjacent lambdas and applications](https://github.com/IntersectMBO/plutus/blob/master/plutus-core/plutus-core/src/PlutusCore/Pretty/Readable.hs#L142-L164); and
* _All_ function parameters are labelled `!0`, except when used as arguments, in which case they use (_one_-based) DeBruijn indexes.

Thus, figuring out how any given function's bound variable(s) get used and where becomes a tedious exercise of 'back counting' DeBruijn indexes.

_Named_ DeBruijns aren't any better:

```
(\i!0 ->
   (\i!0 i!0 ->
      case
        i!2
        [ (\i!0 i!0 ->
             case
               i!3
               [ (\i!0 i!0 ->
                    force
                      ((\i!0 i!0 -> delay (case i!2 [i!2, (force i!1)]))
                         (equalsInteger i!4 i!2)
                         (delay (equalsString i!3 i!1)))) ]) ])
     (constr 0 [42, i!1])
     (constr 0 [42, i!1]))
  "Hello"
```

However, `Name`s _are_ better:

```
(\i-0 ->
   (\i-1 i-2 ->
      case
        i-1
        [ (\i-3 i-4 ->
             case
               i-2
               [ (\i-5 i-6 ->
                    force
                      ((\i-7 i-8 -> delay (case i-7 [i-7, (force i-8)]))
[...]

PR #970, opened 2026-05-27
New backend

kozross:
Solves #939 partially. Given that we have no reason to use DeBruijns if that's not our final target, we can potentially eliminate a lot of issues, including hoisting (we can do this analytically now), as well as improving our IR.


Plutarch/Backend/Arity.hs
```diff
@@ -0,0 +1,140 @@
+module Plutarch.Backend.Arity (
+  Arity (..),
+  getBuiltinArity,
+) where
+
+import PlutusCore qualified as PLC
+
+data Arity = NoArity | Arity Word
```
gnumonik, line 18: It might be useful to leave a comment explaining what this is. I would *guess* that `NoArity` is something that gets used in a context where you're looking at the arity of a term and not a builtin, and is for things like fully-applied functions or constants or whatnot, but I'm not *certain*.

Plutarch/Backend/AST.hs
```diff
@@ -0,0 +1,483 @@
...
+  vmSingleton,
+ )
+import PlutusCore (Some, ValueOf)
+import PlutusCore qualified as PLC
+
+newtype Hash = Hash Int
+  deriving (Eq, Ord) via Int
+  deriving stock (Show)
+
+data Multiplicity
```
gnumonik, line 110: Again, please explain what this is with a brief comment. I don't believe `Multiplicity` is a standard term (or if it is I don't recall seeing it used this way). I think I know what it means but I may well forget before the next time I need to look at it :P

Plutarch/Backend/AST.hs
```diff
@@ -0,0 +1,483 @@
...
+  = MultiplicityOne Hash
+  | MultiplicityMany Hash
+  deriving stock (Show)
+
+data Liftability = Trivial | NonTrivial
```
gnumonik, line 124: Again, I know what this means (and `Trivial` is probably obvious), but it would be helpful to know _precisely_ what `NonTrivial` is meant to indicate. If there's an enumerable set of criteria for non-triviality then it would be better to enumerate them

Plutarch/Backend/AST.hs
```diff
@@ -0,0 +1,483 @@
+{-# LANGUAGE NoPartialTypeSignatures #-}
+
+module Plutarch.Backend.AST (
```
gnumonik, line 22: A top-level comment outlining (very generally) what exactly this AST is for would probably be quite valuable for future maintainers

Plutarch/Backend/AST.hs
```diff
@@ -0,0 +1,483 @@
...
+      mId <- gets (Bimap.lookupR h . fst)
+      maybe act (pure . AnId) mId
+    newBind :: Hash -> ANFBind () -> State (Bimap Id Hash, IntMap (ANFBind ())) Ref
+    newBind h bind = do
+      firstAvailable <- gets (maybe 0 ((+ 1) . fst) . IntMap.lookupMax . snd)
+      let asId = Id firstAvailable
+      modify (bimap (Bimap.insert asId h) (IntMap.insert firstAvailable bind))
+      pure . AnId $ asId
+
+fromRawTerm :: RawTerm () -> AST Hash
```
gnumonik, line 181: I really think this needs at least a brief explanation of how it works. AT THE VERY LEAST you should link the paper somewhere here. This is tricky for me to fully comprehend and I've had you explain it to me *and* I've read the paper!

Plutarch/Backend/AST.hs
```diff
@@ -0,0 +1,483 @@
...
+getFresh ::
+  forall (a :: Type) (m :: Type -> Type).
+  (MonadState a m, Num a) =>
+  m a
+getFresh = do
+  fresh <- get
+  modify (+ 1)
+  pure fresh
+
+findVarUsage :: Word64 -> Maybe PosTree -> RawTerm () -> Maybe (Hash, Bool)
```
gnumonik, line 326: Needs a brief comment explaining what it does - I forgot what the `Bool` indicates and it's not super obvious to me from context.

Plutarch/Backend/PosTree.hs
```diff
@@ -0,0 +1,64 @@
+module Plutarch.Backend.PosTree (
```
gnumonik, line 14: Either add a top level module comment briefly explaining what this does *or* just link the paper. 

We have to at least point at the paper somewhere in here. It would be borderline impossible for someone who wasn't aware of the paper to figure out what any of this is.

Plutarch/Backend/PosTree.hs
```diff
@@ -0,0 +1,64 @@
...
+  hashWithSalt = defaultHashWithSalt
+  {-# INLINEABLE hash #-}
+  hash = \case
+    PHere -> hash (0 :: Int)
+    POne t -> hash (1 :: Int, t)
+    PTwo ts -> hash (2 :: Int, ts)
+    PMany ts -> hash (3 :: Int, Vector.toList ts)
+    PApplyCase t ts -> hash (4 :: Int, t, NEVector.toList ts)
+
+isLinear :: PosTree -> Bool
```
gnumonik, line 105: Does this mean "the position tree is effectively a List?" If so maybe say that. In general I feel as though if you're writing a standalone predicate function, it's a good practice to say what the predicate is.

Plutarch/Backend/Term.hs
```diff
@@ -0,0 +1,334 @@
...
+  vmMap,
+  vmMergeM,
+  vmSingleton,
+ )
+import PlutusCore (Some, ValueOf)
+import PlutusCore qualified as PLC
+
+type data S
+
+data TermEnv = TermEnv
```
gnumonik, line 125: Uh what's the point of this? If you mentioned it I don't remember. As it's defined it's basically `()` is it not? What point does having this serve?

Plutarch/Backend/VarMap.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE NoPartialTypeSignatures #-}
+
+module Plutarch.Backend.VarMap (
```
gnumonik, line 16: As with the PosTree stuff, please either briefly explain what this in a top level module comment is or link the paper.

plutarch.cabal
gnumonik, line 13: We should really clean this up some day. This looks like someone (Las?) just chucked every extension into the defaults. I sympathize with whoever did it because it's very tempting, but some of these actually suck to have turned on. E.g. `OverloadedLists` forces you to annotate types you shouldn't otherwise need to, `BlockArguments` can give you *very weird* errors if you make an indentation mistake, etc. 

(I know it's not related to your work here I just wanted to type it so we remember it's something to address)

gnumonik (changes requested):
The implementation looks fine as far as I can tell but we're setting future maintainers up for a bad time if we don't contextualize some of this work with comments. A link to the paper would go a long way but a few other things need to be explained, if only briefly, because they are not straightforward (and future maintainers may not have the luxury, like I do, of having you explain things to me).

Plutarch/Backend/Term.hs
kozross: It's meant to be a placeholder for `Config` and `InternalConfig`. We will not need all of those, but we will need _some_, and we might as well have that in place ahead of time.

plutarch.cabal
kozross: I agree, and will probably make an issue to do this at some point.

gnumonik (approved):
I think you've addressed my previous comments. 

The code generator looks solid to me - I'd probably write things in basically the same way. I would be astonished if there weren't _some_ subtle bug somewhere but structurally it makes sense and a review of the implementation did not uncover any non-subtle bugs.

PR #973, opened 2026-06-09
Optimize codegen to be two linear passes

### LINKED ISSUE #972: Simplify new backend compiler
Our current system is far too fiddly, and prone to breakage. We've built ANF, let's use it!

kozross:
Closes #972. Our new code generator now makes use of demand analysis, which allows us to reduce the original graph-traversal-and-cache algorithm to two linear passes:

* The first pass collects both direct dependency count and the last (meaning, closest-to-toplevel) computation requiring any dependency;
* The second pass compiles code from dependencies to dependents, following the order of the ANF.

Because we have demand analysis available, and the ANF guarantees a dependency-friendly order, we can perform `let`-binding 'as required' upon encountering any ANF bind, without backtracking or having to work out of order.

### DIFF
--- a/Plutarch/Backend/Compile.hs
+++ b/Plutarch/Backend/Compile.hs
@@ -4,23 +4,22 @@ module Plutarch.Backend.Compile (
   toUPLCTerm,
 ) where
 
-import Control.Monad (foldM)
+import Control.Monad (foldM, guard)
 import Control.Monad.RWS.CPS (
-  MonadReader,
   MonadState (get),
   RWS,
   asks,
   evalRWS,
-  gets,
   modify,
   runRWS,
  )
+import Control.Monad.Reader (ReaderT, runReaderT)
 import Control.Monad.ST (ST, runST)
-import Data.Foldable (for_, traverse_)
+import Data.Foldable (foldl', for_, traverse_)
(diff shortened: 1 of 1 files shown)


Plutarch/Backend/Compile.hs
```diff
@@ -152,36 +330,54 @@ doFixpointAnalysis = NEVector.ifoldl' go Set.empty
...
+  NeverDemanded <> x = x
+  x <> NeverDemanded = x
+  Demanded (Id i) count1 <> Demanded (Id j) count2 = Demanded (Id $ max i j) (count1 + count2)
+
+instance Monoid Demand where
+  mempty = NeverDemanded
+
+-- For each bind, check:
+--
+-- - How many times it is required as a direct dependency and
```
gnumonik: I would perhaps move this comment to the declaration of `Demand`, it wasn't _immediately_ obvious to me that this was (basically) a description of what the datatype itself means and not a description of the process here (since most of the "process" is offloaded to the semigroup instance anyway and this just orchestrates things with a mutable vector)

gnumonik (approved):
One small comment (on the placement of a comment), other than that this seems sensible to me (after figuring out how the demand analysis works).

PR #980, opened 2026-06-11
Curry up RawTerm lambdas and applications, remove Liftability

### LINKED ISSUE #978: Simplify `RawTerm`, move coalescing to `AST`
The following case demonstrates position trees not being correctly adjusted:

```
  Case 4: OK
    Case: \(x :: Natural) (y :: Natural) -> (pbuiltin AddInteger) x y
    1. Does Case 4 compile?
    Successfully compiled!
    RLamAbs
      ()
      [ Just (PApplyCase Nothing [ Just PHere ]) , Just PHere ]
      (RApply
         ()
         (RBuiltin () AddInteger)
         [ RVar () Argument , RVar () Argument ])
```

This stems from us attempting to coalesce adjacent lambdas and applications at the `RawTerm` stage, rather than later. We should keep `RawTerm` closer to the eDSL (and simpler) to avoid issues like this. We can then do coalescing later, probably at the `AST` stage.

kozross:
Closes #978 . This also eliminates `Liftability` from `AST`. It ultimately isn't terribly useful, as the ANF already does as much lifting as we could want or need.

### DIFF
--- a/Plutarch/Backend/ANF.hs
+++ b/Plutarch/Backend/ANF.hs
@@ -49,7 +49,6 @@ import Plutarch.Backend.AST (
     ASTLeaf
   ),
   Hash,
-  Liftability,
   Multiplicity,
  )
 import Plutarch.Backend.AST qualified as AST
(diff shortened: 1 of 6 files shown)


gnumonik (approved):
Nothing jumps out to me as problematic here & it makes sense that liftability isn't a useful concept anymore after reviewing the previous PR.

PR #981, opened 2026-06-11
Trivial constant inlining

### LINKED ISSUE #975: Inline trivial constants
Currently, any constant used more than once will be `let`-bound. This isn't really worth it in the following cases:

* `()`
* `Bool`
* `Integer` when less than 256

The code generator should be modified to ignore use count for any such constant. This is an easy static check.

kozross:
Closes #975. Also replaces the explicit existentials in `punsafeCase` and `punsafeConstr` with a wrapper `SomeTerm`, as this is easier to work with.

### DIFF
--- a/Plutarch/Backend/Compile.hs
+++ b/Plutarch/Backend/Compile.hs
@@ -8,7 +8,7 @@ module Plutarch.Backend.Compile (
   toUPLCTerm,
 ) where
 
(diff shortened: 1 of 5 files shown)


gnumonik (approved):
This is fine, only comment is the stuff I sent you on slack about `SomeTerm` (which it looks like you just responded to)

PR #983, opened 2026-06-11
Convert demand analysis to annotate ANF

### LINKED ISSUE #982: Change demand analysis to annotate ANF
Given that every ANF bind must have a demand analysis, and that the codegen function is already complex, breaking this out into a pass seems sensible.

kozross:
Fixes #982.


gnumonik (approved):
Nothing seems amiss here, as far as I can tell.

PR #987, opened 2026-06-12
Koz/976

### LINKED ISSUE #976: Eta-reduce builtins under lambdas
Currently, code like `\x y -> apply (builtin AddInteger) x y` will compile literally. This is inefficient, as we can just compile `builtin AddInteger` instead. 

While we could do this by a static analysis, it's probably easier to do this at codegen time by inspecting `ANFLam`. Once complicating factor is that some builtins are 'polymorphic', requiring `force` nodes. This means we must both 'look through' them and also preserve them when needed.

kozross:
Closes #976 . Also provides some basic `PInteger` bits.

### DIFF
--- a/Plutarch/Backend/AST.hs
+++ b/Plutarch/Backend/AST.hs
@@ -31,6 +31,7 @@ module Plutarch.Backend.AST (
 ) where
 
 import Control.Applicative ((<|>))
+import Control.Monad (guard)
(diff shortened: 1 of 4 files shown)


gnumonik (approved):
Only question was the one I asked you on slack (which you answered). Nothing else jumps out as potentially problematic to me.

PR #989, opened 2026-06-15
Resolve possible RCompiled clashes

### LINKED ISSUE #977: Ensure we don't have name collisions with any precompiled code chunks
Currently, a precompiled code chunk is treated as an opaque blob. This can be a problem, as we may accidentally generate a name for a variable or bind that clashes with a name in that chunk. We already do a reserved name check for variable hashes, so this shouldn't pose too much of a problem: we already crawl the entire AST.

One minor hiccup might arise if we have a variable hash clash with a precompiled code chunk variable `Unique`. While this is _exceedingly_ unlikely, it's not outright impossible, and we need a strategy to solve this. Giving each code chunk a unique number and offsetting their uniques by that much (while reserving) is probably adequate. Alternatively, we can do something similar to what hash tables do to resolve collisions.

kozross:
Closes #977 . Also includes code to demonstrate that this works (as well as how bad it is). As part of this, I've added `pcompiled` and `punsafeCompiled` to allow access to embedding compiled code.

### DIFF
--- a/Plutarch/Backend/Compile.hs
+++ b/Plutarch/Backend/Compile.hs
@@ -19,7 +19,8 @@ import Control.Monad.RWS.CPS (
  )
 import Control.Monad.Reader (MonadReader, runReaderT)
 import Control.Monad.ST (runST)
-import Data.Foldable (foldl')
+import Control.Monad.State.Strict (put, runStateT)
+import Data.Foldable (foldl', for_)
 import Data.Kind (Type)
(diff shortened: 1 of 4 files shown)


gnumonik (approved):
I don't see any issues here. I'm _pretty sure_ that bits of compiled code can't refer to variables defined in other bits of compiled code, but even if they can, what you have here _should_ rewrite consistently so that wouldn't be a problem (afaict).

PR #990, opened 2026-06-15
Better treatment of subtyping

### LINKED ISSUE #974: Better treatment of primitive types
Plutarch _essentially_ has representational polymorphism embedded at the type level and reinforced in several ways. This is a good idea, but it's done in a strange way. The new backend is our chance to do better.

kozross:
Closes #974 . This defines a more usefully-named relationship, along with nice error messages.


gnumonik (approved):
Everything looks fine to me. I was curious about why we're using type families instead of classes for a few things here, but that's not materially important, and anyway I asked you on slack (respond at your leisure, it's not urgent.)

PR #993, opened 2026-06-17
More subtyping, better `PlutusType`

### LINKED ISSUE #991: Better story for `PlutusType`
Follow-on from #990 . `PlutusType` overbundles a range of concepts, and the work in #990 was designed to disentangle some of it. We should do the rest, ideally with some examples of how it would behave.

kozross:
Closes #991 partially.


gnumonik (approved):
I looked over this as closely as I could and did not discover any issues.

PR #995, opened 2026-06-17
Always inline builtins

### LINKED ISSUE #994: Always inline builtins
Currently, given code like this:

```haskell
-- Some conceptual shortcuts used
psumOfSquares :: Term s (PInteger :--> PInteger :--> PInteger)
psumOfSquares = plam $ \x y -> paddInteger # (pmultiplyInteger # x # x) # (pmultiplyInteger # y # y)
```

The new backend would attempt to compile something like so:

```
\x y -> let times = PLC.MultiplyInteger 
          in let ySquared = times y y
            in let xSquared = times x x
              in PLC.AddInteger xSquared ySquared
```

Which, after the `let`-transform becomes something like

```
\x y -> (\times -> PLC.AddInteger (times x x) (times y y)) PLC.MultiplyInteger
```

This is larger than it needs to be. If we just inlined builtins in all cases, we would instead generate:

```
\x y -> PLC.AddInteger (PLC.MultiplyInteger x x) (PLC.MultiplyInteger y y)
```

This makes the code smaller at no performance cost, and also makes builtin eta-reduction analysis easier. Unlike the solution taken for #975 , however, the better option is to eliminate `LBuiltin` nodes entirely, and instead extend `Ref` to also allow the possibility of a builtin. This way, we can eliminate this problem by construction, which wasn't possible or desirable for constants in general, as large constants might be worth `let`-binding for code size reasons.

kozross:
Closes #994. In the end, changing `ANF` to remove builtin nodes entirely proved awkward, as then, when asked to compile (the eta-reduction to) a single builtin, the compilation pipeline explodes. While we could modify `ANF` to deal with this eventuality, it's both annoying enough and rare enough to not really be worth it.

### DIFF
--- a/Plutarch/Backend/ANF.hs
+++ b/Plutarch/Backend/ANF.hs
@@ -241,12 +241,12 @@ data Demand
     @since wip
     -}
     Demanded Id Word64
-  | {- | Some constants should never be @let@-bound. This means their demand
+  | {- | Some things should never be @let@-bound. This means their demand
     analysis is trivial.
 
     @since wip
     -}
-    TrivialConstant
+    Trivial
   deriving stock
     ( -- | @since wip
       Eq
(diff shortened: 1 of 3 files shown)


gnumonik (approved):
Again, no issues here.

PR #996, opened 2026-06-17
Eliminate applications of identity

### LINKED ISSUE #992: Eliminate identity function calls
Currently, the code generator will produce calls to `id`, possibly even `let`-binding it. This is pointless and a waste of code, and we should eliminate it.

The easiest place to do this is at the ANF stage. This means that we can quickly identify if we have `id` anywhere, and then fix _all_ of its call sites in one go.

kozross:
Closes #992 .


gnumonik (approved):
Did not see any issues.

PR #998, opened 2026-06-18
Propagate error nodes

### LINKED ISSUE #997: Propagate `error` into applications, constr and case
Follow-on from #995 . Currently, `f foo bar error` will get compiled literally, as will `constr 2 error x` and `case error [handlerOne, handlerTwo]`. This is pointless, as the only result in each of these cases will be `error`.

Likewise, it is possible that `error` will be `let`-bound. This is doubly bad: not only does it inflate the code, it can actually change its semantics by `error`ing earlier than it should!

The changes we need to make are as follows:

* Inform demand analysis that `error` should never be `let`-bound; and
* Make sure that when compiling, we propagate `error` through applications, `constr` (in any position) or `case` (for the scrutinee only)

kozross:
Closes #997 .


gnumonik (approved):
This looks fine, I raised the only issue on slack and you addressed my concerns adequately there.

PR #1001, opened 2026-06-22
Koz/999

### LINKED ISSUE #999: `pcompose` primitive and specialized function composition pipeline
Right now, function composition is fairly unpleasant: every composition adds a lambda and two applications, even if part of another pipeline, making it generally unworkable. With the new backend, we can have a `pcompose` primitive which mark such pipelines with a designated `RawTerm` constructor, allowing supercombinator analysis on the back end.

A challenge with this is any functions that aren't (at least) affine, as we potentially have questions of inlining to consider.

kozross:
Closes #999 . The pipeline is currently rather crude, but it's there, and is _technically_ a bit better than defining composition manually.

This also fixes a bug in the code generator where, if we ended up needing the same `let`-bind more than once in a single ANF bind, the generator would emit multiple identical `let`-bind transforms.

### DIFF
--- a/Plutarch/Backend/Compile.hs
+++ b/Plutarch/Backend/Compile.hs
@@ -39,6 +39,7 @@ import Plutarch.Backend.ANF (
   ANFBind (
     ANFApply,
     ANFCase,
+    ANFCompose,
     ANFConstr,
     ANFDelay,
     ANFFix,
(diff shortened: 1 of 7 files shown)


Plutarch/Backend/Compile.hs
```diff
@@ -405,15 +444,31 @@ mkFixpointNames acc i = do
...
+
 -- A read-only environment for compilation. Contains:
 --
--- - All ANF binds (with demand analysis)
--- - All unique names reserved for these binds, in the same order
--- - Unique name pairs for each fixpoint we have to compile
+
+-- * All ANF binds (with demand analysis)
+
+-- * All unique names reserved for these binds, in the same order
```
gnumonik: Could you double check this comment to make sure it aligns with what's actually in `CompileEnv`? It looks like the first two (`All ANF binds` and `All unique names reserved for these binds`) are smashed together in `ceBinds` and that makes it kind of hard to follow what everything else in the comments actually explains in `CompileEnv`. 

(It's not hard to figure it but it'd be better to have the comments line up exactly with the fields of the record IMO)

Plutarch/Backend/PosTree.hs
```diff
@@ -123,6 +139,10 @@ isLinear = \case
     Just tLinearity -> case sequenceA_ ts of
       Nothing -> tLinearity
       Just _ -> False
+  PCompose ts -> case NEVector.uncons ts of
+    (x, xs) -> case Vector.foldl' go (fmap isLinear x) xs of
```
gnumonik: Incredibly minor but shouldn't this be something like `foldl' (go . isLinear) xs`. Or I guess the types don't line up there (I think?) but there's no harm in shoving the `isLinear` check into `go`. GHC *probably* optimizes this to that anyway but I'm not 100% sure it does.

Plutarch/Backend/Term.hs
```diff
@@ -268,6 +278,50 @@ pfix f = Term $ do
...
+  Term s (c :--> a) ->
+  Term s (c :--> b)
+pcompose f g = Term $ do
+  (fvm, ft) <- asRawTerm f
+  (gvm, gt) <- asRawTerm g
+  case (ft, gt) of
+    (RCompose _ compsF, RCompose _ compsG) -> do
+      let lenFs = NEVector.length compsF
+      let lenGs = NEVector.length compsG
+      let extendedFVM = vmMap (extendComposeEnd lenGs) fvm
```
gnumonik, line 303: Again a minor thing, but if you don't use `extendComposeEnd / extendComposeStart / mergeCompose` outside of this function (and afaict you don't), I think you should `where` bind them so the connection between the two functions is clearer structurally / at a glance. Improves readability at least. 

(There are some other places where I think you have separate helpers that are only used in one place and unlikely to be used elsewhere, but I forgot to note them.)

gnumonik: ...also doing that makes it easier to review lol

gnumonik (changes requested):
Just a few minor things. Overall it seems solid to me. 

I am mainly requesting changes so you can take a look at the comment I flagged. The stylistic things are obviously not that important.

Plutarch/Backend/PosTree.hs
kozross: No, because that won't compile. `fmap isLinear x` is the starting point of the fold.

gnumonik (approved):
Looks good now.

PR #1005, opened 2026-06-23
PlutarchType

### LINKED ISSUE #1000: New basis for the 'is a Plutarch type' hierarchy
`PMatch` currently acts as the baseline of what can be considered a 'valid' Plutarch type. However, this is problematic: for any non-compositional type (that is, any type with no meaningful host language structure, like `PInteger`), its method is completely trivial.

A better approach would be to make _application_ the baseline. This would also allow `PRepresentation` to become an associated type.

kozross:
Closes #1000 .


gnumonik (approved):
Went over this carefully and did not see anything amiss.

PR #1008, opened 2026-06-25
Remaining primitives

### LINKED ISSUE #1006: Fill out the rest of the default universe
Currently, there are some types missing from `Plutarch.Primitive` that happen to be in the default universe. We should fix this.

kozross:
Closes #1006 .


gnumonik (approved):
Do not see any issues. (Since I noticed some of this stuff was missing when I tried to write pretty printer "tests" I had a pretty good idea of what this should look like, and it does look like that)

PR #1010, opened 2026-06-29
PLiftable

### LINKED ISSUE #1007: Port `PLiftable` to new backend
This will be quite similar to what already exists, but due to the changes in the base hierarchy (and representations), some details will be different too.

kozross:
Closes #1007.

### DIFF
--- a/Plutarch/Primitive/Liftable.hs
+++ b/Plutarch/Primitive/Liftable.hs
@@ -38,6 +38,7 @@ import PlutusTx qualified as PTx
 data ReprError
   = UnexpectedNegative Integer
   | ByteOutOfBounds Integer
+  | UnexpectedZero
   deriving stock
     ( -- | @since wip
       Eq
--- a/Plutarch/Primitive/Liftable.hs
+++ b/Plutarch/Primitive/Liftable.hs
@@ -1,6 +1,9 @@
-{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
 
 module Plutarch.Primitive.Liftable (
+  ReprError (..),
+  PlutusRepresentable (..),
   LiftError (..),
   PLiftable (..),
   pconstant,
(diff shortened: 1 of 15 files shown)


Plutarch/Primitive/Liftable.hs
```diff
@@ -0,0 +1,213 @@
...
+1. @'reprToHask' '.' 'haskToRepr'@ @=@ @'Right'@
+
+@since wip
+-}
+class
+  PLC.DefaultUni `PLC.Contains` AsPlutus a =>
+  PlutusRepresentable (a :: Type)
+  where
+  type AsPlutus a :: Type
+  type AsPlutus a = a
```
gnumonik, line 101: Is this actually a sensible default? I honestly don't know. 

I'm guessing it's like this for deriving machinery or something like that?

Plutarch/Primitive/Liftable.hs
```diff
@@ -0,0 +1,213 @@
...
+data LiftError
+  = DidNotEvaluate EvalError
+  | NotAConstant
+  | WrongConstantType
+  deriving stock
+    ( -- | @since wip
+      Show
+    )
+
+{- | = Laws
```
gnumonik: I would add *some* kind of comment that explains, at some level of generality, how this class works / what its purpose is.

gnumonik: Specifically you need to say at least enough that someone can understand how to write an instance. I don't think you necessarily have to explain in a comment here how this works _structurally_ (i.e. how it relates to the rest of Plutarch), but it'd be nice to have a blurb about what `AsPlutus` means.

Plutarch/Primitive/Liftable.hs
```diff
@@ -0,0 +1,213 @@
...
+  reprToHask = traverse reprToHask
+
+-- | @since wip
+instance (PlutusRepresentable a, PlutusRepresentable b) => PlutusRepresentable (a, b) where
+  type AsPlutus (a, b) = (AsPlutus a, AsPlutus b)
+  haskToRepr = bimap haskToRepr haskToRepr
+  reprToHask = bitraverse reprToHask reprToHask
+
+-- | @since wip
+class
```
gnumonik, line 192: This is something of an aside, but would this break if you didn't have the `PlutusRepresentable (AsHaskell a)` constraint? You couldn't use the methods without it so I understand why it's here but does it _have_ to be here explicitly? (I'm just curious)

Plutarch/Primitive/Liftable.hs
```diff
@@ -0,0 +1,213 @@
...
+  PLC.DefaultUni `PLC.Contains` AsPlutus a =>
+  PlutusRepresentable (a :: Type)
+  where
+  type AsPlutus a :: Type
+  type AsPlutus a = a
+  haskToRepr :: a -> AsPlutus a
+  default haskToRepr :: a ~ AsPlutus a => a -> AsPlutus a
+  haskToRepr = id
+  reprToHask :: AsPlutus a -> Either ReprError a
+  default reprToHask :: a ~ AsPlutus a => AsPlutus a -> Either ReprError a
```
gnumonik, line 106: Actually I think you should document `reprToHask` too, because it's not straightforwardly intuitive. 

I _think_ that it should always be _possible_ to return a `Right` here, correct? So the `Left` is meant to indicate something like a case where (I guess) `a` is "smaller than" `AsPlutus a` (such as in the instance below for `Natural`).

gnumonik, line 106: Actually looking at the `Natural` example makes me think you need an additional constructor to `ReprError` that takes a `String` param or something and allows users to write their own custom errors. 

Someone could (I think?) want to write their own instance for a custom type here and it's a bit awkward that the only errors they would have access to without doing so are specific to bytestrings or naturals or whatever.

gnumonik (changes requested):
Not all of my comments are super important but please make the documentation changes & either change `ReprError` in the way I suggested or explain to me (on slack is fine) why I'm wrong and it isn't necessary.

Plutarch/Primitive/Liftable.hs
kozross: This allows one-liner derivations for any type that is already part of the default universe: you can see this in some of the instances below.

Plutarch/Primitive/Liftable.hs
kozross: Agreed. I should really do this for _both_ of them.

Plutarch/Primitive/Liftable.hs
kozross: It's kind of not sensible to _not_ have it. The whole point is that you can only lift things that either belong to the default universe directly, or by way of some 'smaller' type. Without this constraint, this type class loses all meaning. _Technically_ this constraint is redundant, as no method makes use of it, but we definitely need it there for moral reasons.

Plutarch/Primitive/Liftable.hs
kozross: The whole idea behind `reprToHask` is that is allows us to have 'smaller' stand-ins for bits of the default universe. You can see that all `ReprError`s indicate that the given term of the representation in the universe is not something `a` can represent. 

Agreed on the docs though.

kozross: I left a comment on Slack to that effect. I had considered something much more powerful than this, but decided not to include it. Given that you've raised this, it means I should.

Plutarch/Primitive/Liftable.hs
gnumonik: Ok, that's fine, I was just wondering whether it really was essential or it was just "moral".

gnumonik (approved):
Looks good.

PR #1012, opened 2026-06-29
AST Printer

gnumonik:
As before, looking at the output is the most important thing. It should be hooked up to all the tests now. 

We cannot really use operators here due to the need to print the annotations "inline", which also forces some other less than ideal formatting choices upon us.


Plutarch/Backend/AST.hs
```diff
@@ -153,6 +168,15 @@ data Leaf (ann :: Type)
       Eq
     )
 
+instance Pretty (Leaf ann) where
```
kozross, line 172: `@since wip`

Plutarch/Backend/AST.hs
```diff
@@ -193,6 +217,58 @@ data AST (ann :: Type)
       Eq
     )
 
+-- For the sake of consistency this uses the same formatting as the overlapping instance,
+-- although this *could* be prettier if we wanted it to be.
+instance {-# OVERLAPPABLE #-} Pretty (AST ()) where
```
kozross, line 225: `@since wip`

Plutarch/Backend/AST.hs
```diff
@@ -193,6 +217,58 @@ data AST (ann :: Type)
...
+        <+> align (indent 2 (blockList . map pretty . NEVector.toList $ args))
+    ASTConstr _ cix args -> "Constr" <> brackets (pretty cix) <+> blockList (pretty <$> Vector.toList args)
+    ASTCase _ scrut handlers ->
+      "case"
+        <+> pretty scrut
+        <+> hardline
+        <> indent 2 (blockList . map pretty . NEVector.toList $ handlers)
+    ASTCompose _ args -> "Compose" <+> align (blockList (pretty <$> NEVector.toList args))
+
+instance {-# OVERLAPS #-} Pretty ann => Pretty (AST ann) where
```
kozross, line 247: `@since wip`

kozross (changes requested):

PR #1014, opened 2026-07-07
Numerical hierarchy

### LINKED ISSUE #1011: Port numerical hierarchy
Follow-on from #1007 . Now that we have all the necessary machinery to have constants again, as well as to construct useful values, we might as well also bring this over.

kozross:
Closes #1011. Not quite complete with all laws, as we currently lack generalized notions of equality and ordering.

### DIFF
--- a/Plutarch/Numeric/Additive.hs
+++ b/Plutarch/Numeric/Additive.hs
@@ -9,17 +9,18 @@ module Plutarch.Numeric.Additive (
 ) where
 
 import Data.Kind (Type)
-import Data.Vector.NonEmpty qualified as NEVector
 import Plutarch.Backend.S (S)
 import Plutarch.Backend.Term (
   Term,
   perror,
-  pfix,
   plam',
-  punsafeCase,
   punsafeCoerce,
   punsafeConstant,
-  toSomeTerm,
+ )
+import Plutarch.Numeric.Helpers (
+  pexpBySquaring,
+  pione,
+  pizero,
  )
 import Plutarch.Primitive.Apply (
   PlutarchType (PRepresentation),
@@ -43,8 +44,6 @@ import Plutarch.Primitive.BuiltinFun (
   pequalsInteger,
   plessThanEqualsInteger,
   pmultiplyInteger,
-  pquotientInteger,
-  premainderInteger,
   psubtractInteger,
  )
 import Plutarch.Primitive.Function ((:-->))
@@ -96,7 +95,7 @@ class PAdditiveSemigroup a => PAdditiveMonoid (a :: S -> Type) where
     pif
       (pequalsInteger # pizero # pcoerce n)
       pzero
-      (pexpBySquaring padd # x # pcoerce n)
+      (pscalePositive # x # punsafeCoerce n)
 
 -- | @since wip
 instance PAdditiveMonoid PInteger where
@@ -169,32 +168,3 @@ infix 6 #+
 x #- y = pminus # x # y
 
 infix 6 #-
-
--- Helpers
-
-pexpBySquaring ::
-  forall (a :: S -> Type) (s :: S).
-  PlutarchType a =>
-  Term s (a :--> a :--> a) ->
-  Term s (a :--> PInteger :--> a)
-pexpBySquaring f = plam' $ \x -> pfix $ \self -> plam' $ \i ->
-  -- We know that `i` cannot be non-positive here.
-  pif
-    (pequalsInteger # i # pione)
-    x
-    ( let two = punsafeConstant (PLC.someValue @Integer 2)
-          stepDown = self #$ pquotientInteger # i # two
-          squared = f # stepDown # stepDown
-       in -- Because the remainder by two can only be 0 or 1, we can use
-          -- `punsafeCase` here for speed.
-          punsafeCase (premainderInteger # i # two)
-            . NEVector.cons (toSomeTerm squared)
-            . NEVector.singleton
-            $ toSomeTerm (f # squared # x)
-    )
-
-pizero :: forall (s :: S). Term s PInteger
-pizero = punsafeConstant (PLC.someValue @Integer 0)
-
-pione :: forall (s :: S). Term s PInteger
-pione = punsafeConstant (PLC.someValue @Integer 1)
--- a/Plutarch/Numeric/Additive.hs
+++ b/Plutarch/Numeric/Additive.hs
@@ -1,0 +1,200 @@
+{-# LANGUAGE UndecidableSuperClasses #-}
+
+module Plutarch.Numeric.Additive (
+  PAdditiveSemigroup (..),
+  PAdditiveMonoid (..),
+  PAdditiveGroup (..),
+  (#+),
+  (#-),
+) where
+
+import Data.Kind (Type)
+import Data.Vector.NonEmpty qualified as NEVector
+import Plutarch.Backend.S (S)
+import Plutarch.Backend.Term (
+  Term,
+  perror,
+  pfix,
+  plam',
+  punsafeCase,
+  punsafeCoerce,
+  punsafeConstant,
+  toSomeTerm,
+ )
+import Plutarch.Primitive.Apply (
+  PlutarchType (PRepresentation),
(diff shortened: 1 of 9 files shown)


Plutarch/Numeric/Additive.hs
```diff
@@ -0,0 +1,170 @@
...
+import Plutarch.Primitive.Function ((:-->))
+import Plutarch.Primitive.Numeric (PInteger, PNatural, PPositive)
+import PlutusCore qualified as PLC
+
+-- | @since wip
+class PlutarchType a => PAdditiveSemigroup (a :: S -> Type) where
+  padd :: forall (s :: S). Term s (a :--> a :--> a)
+  default padd ::
+    forall (s :: S).
+    PAdditiveSemigroup (PRepresentation a) => Term s (a :--> a :--> a)
```
gnumonik, line 60: Shouldn't this be a class constraint? Or does that break things? I _thought_ you could do it as a class constraint anyway. 

It's probably not a big deal but I _thought_ there was something a week or two ago what was roughly analogous to this (i.e. a situation where you could choose to put a constraint on the PRepresentation in the method or the class) and you chose a class constraint. 

If it has to be the way it is here, could you explain why (so that I may learn something)?

Plutarch/Numeric/Additive.hs
```diff
@@ -0,0 +1,170 @@
...
+  padd = pbls12_381_G2_add
+  pscalePositive = plam' $ \x -> plam' $ \p ->
+    pbls12_381_G2_scalarMul # punsafeCoerce p # x
+
+-- | @since wip
+class PAdditiveSemigroup a => PAdditiveMonoid (a :: S -> Type) where
+  pzero :: forall (s :: S). Term s a
+  default pzero ::
+    forall (s :: S).
+    PAdditiveMonoid (PRepresentation a) => Term s a
```
gnumonik, line 93: Comment for PAdditiveSemigroup applies here too

Plutarch/Numeric/Additive.hs
```diff
@@ -0,0 +1,170 @@
...
+instance PAdditiveMonoid PInteger where
+  pzero = pizero
+  pscaleNatural = punsafeCoerce pmultiplyInteger
+
+-- | @since wip
+instance PAdditiveMonoid PNatural
+
+-- | @since wip
+instance PAdditiveMonoid PBLS12_381_G1_Element where
+  pzero = perror -- TODO: pconstant
```
gnumonik: For purely logistical reasons it'd be better to not TODO these now, I just have a hunch that after a month of working on something else, we'll probably forget we need to add something here.

Plutarch/Numeric/Additive.hs
```diff
@@ -0,0 +1,170 @@
...
+  pscaleNatural = plam' $ \x -> plam' $ \n ->
+    pbls12_381_G1_scalarMul # pcoerce n # x
+
+-- | @since wip
+instance PAdditiveMonoid PBLS12_381_G2_Element where
+  pzero = perror -- TODO: pconstant
```
gnumonik: See previous comment

Plutarch/Numeric/Euclidean.hs
```diff
@@ -0,0 +1,81 @@
...
+-}
+class (PZeroable a, PMultiplicativeMonoid a) => PEuclidean (a :: S -> Type) where
+  pdiv :: Term s (a :--> PNonZero a :--> a)
+  pmod :: Term s (a :--> PNonZero a :--> a)
+  pgcd :: Term s (a :--> PNonZero a :--> a)
+  pdegree :: Term s (a :--> PNatural)
+
+-- | @since wip
+instance PEuclidean PNatural where
+  pdiv = punsafeCoerce pquotientInteger
```
gnumonik, line 56: Again, super minor, but can we not use `pcoerce` instead here? Type system should know that Naturals are Integers "under the hood", shouldn't it?

Plutarch/Numeric/Helpers.hs
```diff
@@ -0,0 +1,57 @@
...
+import PlutusCore qualified as PLC
+
+-- | @since wip
+pexpBySquaring ::
+  forall (a :: S -> Type) (s :: S).
+  PlutarchType a =>
+  Term s (a :--> a :--> a) ->
+  Term s (a :--> PInteger :--> a)
+pexpBySquaring f = plam' $ \x -> pfix $ \self -> plam' $ \i ->
+  -- We know that `i` cannot be non-positive here.
```
gnumonik, line 41: How do we know this? It's not immediately obvious to me

Plutarch/Numeric/Multiplicative.hs
```diff
@@ -0,0 +1,81 @@
...
+  PNatural,
+  PPositive,
+ )
+
+-- | @since wip
+class PlutarchType a => PMultiplicativeSemigroup (a :: S -> Type) where
+  pmultiply :: forall (s :: S). Term s (a :--> a :--> a)
+  default pmultiply ::
+    forall (s :: S).
+    PMultiplicativeSemigroup (PRepresentation a) => Term s (a :--> a :--> a)
```
gnumonik, line 35: See comment on PAdditiveSemigroup

Plutarch/Numeric/Multiplicative.hs
```diff
@@ -0,0 +1,81 @@
...
+
+-- | @since wip
+instance PMultiplicativeSemigroup PPositive
+
+-- | @since wip
+class PMultiplicativeSemigroup a => PMultiplicativeMonoid (a :: S -> Type) where
+  pone :: forall (s :: S). Term s a
+  default pone ::
+    forall (s :: S).
+    PMultiplicativeMonoid (PRepresentation a) => Term s a
```
gnumonik, line 55: See comment on PAdditiveSemigroup (I'm just flagging all of these in case the change I suggested there works, obviously ignore these if it does not work)

Plutarch/Numeric/Zeroable.hs
```diff
@@ -0,0 +1,72 @@
...
+import Plutarch.Primitive.Function ((:-->))
+import Plutarch.Primitive.Numeric (PInteger, PNatural, PPositive)
+
+{- | = Laws
+
+1. @ptoNonZero x x pcoerce@ @=@ @x@
+
+@since wip
+-}
+class
```
gnumonik, line 66: Could you briefly explain (in English) what this class is for? I get the law and I get the methods but I don't get the _point_

gnumonik (changes requested):
More questions than change requests, but it's important I understand this so I'd like you to answer the questions :P

Plutarch/Numeric/Additive.hs
kozross: Good catch, I _can_ actually fix this now. I ... forgot, only proving your point further.

Plutarch/Numeric/Euclidean.hs
kozross: `pcoerce` is covariant, whereas this has to be contravariant.

Plutarch/Numeric/Helpers.hs
kozross: The only places we ever call it ensure this. This is very much an internal method, but I think I should state my assumption.

Plutarch/Numeric/Zeroable.hs
kozross: This type class mostly designates a 'zerofree equivalent' of a type, to be used in the definition of `PEuclidean`'s `pdiv` and `pmod` methods. This way, we can ensure that division-by-zero cannot occur. 

The method is morally `a -> Maybe (NonZero a)`, but because we don't have a 'primitive' or even singular notion of `Maybe`, I have to CPS it.

Plutarch/Numeric/Helpers.hs
gnumonik: Ok. Yeah, say that explicitly. The comment makes it seem like this function itself (not where we happen to call it) provides some kind of guarantee, which isn't the case.

gnumonik (approved):
Looks good to me now.

PR #1018, opened 2026-07-08
UPLC Pretty Printer

gnumonik:
As with the previous ones, please look at the test output and suggest improvements. This is partly adapted from what we had in Covenant, but I made a bunch of aesthetic improvements and had to change the way we handle names. 

I went with `#` for `Apply`, `<_>` for `delay`, and `!` for `force` here largely to reduce the number of brackets and parens. 

Subjectively I think this is more readable than the alternatives but if you disagree, I'll go with what you prefer w/o complaints.

### DIFF
--- a/Plutarch/Utils/Pretty.hs
+++ b/Plutarch/Utils/Pretty.hs
@@ -6,6 +6,7 @@ module Plutarch.Utils.Pretty (
   taggedNode,
   prettyAnnotated,
   blockList,
+  prettyUPLC,
 ) where
 
 import Control.Lens.Plated
@@ -20,9 +21,12 @@ import Prettyprinter (
   Doc,
   Pretty (pretty),
   align,
+  angles,
   brackets,
+  encloseSep,
   group,
   hardline,
+  hsep,
   indent,
   list,
   parens,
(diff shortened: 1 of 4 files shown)


Plutarch/Utils/Pretty.hs
```diff
@@ -140,3 +150,68 @@ prettyAnnotated getAnn annHandler nodePrinter x = annHandler (pretty a) $ nodePr
...
+  ([], rest) -> prettyNoBind rest
+  (letBinds, rest) ->
+    let pRest = "in" <+> prettyNoBind rest
+     in align . vsep . reverse $ (pRest : letBinds)
+  where
+    -- if it's `arg` it came from Plutarch code and we just use the "compact pretty hash" as the visible name directly,
+    -- but if it's something else then it came from a blob of compiled UPLC and we need to make the text part visible
+    prettyName :: Name -> Doc ann
+    prettyName (Name txt (Unique u))
+      | txt == "arg" = pretty . compactReadableVar . fromIntegral $ u
```
kozross, line 165: `arg` is not the only name we could potentially generate. From memory, there is at least `bind` or something similar for `let`.

kozross (changes requested):

PR #1022, opened 2026-07-13
Koz/1017

### LINKED ISSUE #1017: `PSOP` and TH derivations for type classes
We now have enough stuff in place to consider porting this over.

kozross:
Closes #1017 . This provides `PSOP`, as well as TH derivations for SOP encodings that produce `PlutarchType`, `PMatch`, `PCon` and `PEq`, with some codegen tests.

As part of this, I also discovered a nasty code generator bug which could leave dangling references for `let`-binds. This required a change to how we do demand analysis, as well as modifications for the code generator. I deliberately didn't change any APIs (only implementations) to limit the damage.

### DIFF
--- a/Plutarch/Backend/ANF.hs
+++ b/Plutarch/Backend/ANF.hs
@@ -23,7 +23,7 @@ module Plutarch.Backend.ANF (
   getANFBindAnn,
 ) where
 
-import Control.Monad.ST (runST)
+import Control.Monad.ST (ST, runST)
 import Control.Monad.State.Strict (
   State,
   execStateT,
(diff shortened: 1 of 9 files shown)


Plutarch/Backend/ANF.hs
```diff
@@ -351,82 +356,137 @@ data Demand
...
-      MVector.write mv i . ANFCase mempty scrut $ handlers
-    ANFCompose _ components -> do
-      traverse_ (updateDemandAt mv i) components
-      MVector.write mv i . ANFCompose mempty $ components
-  v <- Vector.unsafeFreeze mv
-  pure . ANF bm . NEVector.unsafeFromVector $ v
+  countMV <- MVector.replicate len (0 :: Int)
+  neededVarsMV <- MVector.replicate len Set.empty
+  -- Collect 'raw' statistics on use count, needed variables and binding sites
+  bindingSites <- flip execStateT Map.empty $
```
gnumonik: Small thing: Could this be split out into a helper function in the `where` bindings? This is just pretty big for the RHS of `<-` in a `do` block and makes it somewhat hard to read.

Plutarch/TH/Helpers.hs
```diff
@@ -0,0 +1,156 @@
...
+bindToName = \case
+  PlainTV n _ -> n
+  KindedTV n _ _ -> n
+
+{- | Check that every field of the given data constructor is @'Term'@-wrapped.
+Error if not, or if given a nested @forall@ or GADT.
+
+@since wip
+-}
+checkFieldsAreTerms :: Con -> Q ()
```
gnumonik, line 99: It's been a while since I've written TH: Will type synonyms break this check or are they resolved at this point?

I get that we don't use them but it wouldn't be totally absurd to think someone might.

Plutarch/TH/Strategy.hs
```diff
@@ -0,0 +1,58 @@
...
+import Plutarch.TH.Helpers (checkFieldsAreTerms, checkTyName)
+import Plutarch.TH.SOP (deriveSOP)
+
+{- | Specifies a representation choice for the data type you want derivations
+for. This will determine both what instances are generated, and also what the
+generated instances will look like.
+
+@since wip
+-}
+data Strategy
```
gnumonik, line 22: Eventually there will be other strategies, right? I don't understand the point of the `Strategy` type if there's only one, though maybe I'm missing something.

gnumonik (approved):
Ok I think I've looked at this about as closely as I can and everything seems sensible to me. I just had a few questions which I don't think should block the PR, so I'm approving.

Plutarch/TH/Helpers.hs
kozross: I genuinely don't know the answer to that. I can add a check that would reject type synonyms entirely at least.

Plutarch/TH/Strategy.hs
kozross: Yep, there will be. `SOP` is just the first. I've already got a PR up with another (#1027 ).

gnumonik (approved):

PR #1026, opened 2026-07-17
Track actual multiplicity

### LINKED ISSUE #1025: Make `Multiplicity` an exact count
Some of the transformations we want to do (particularly tree grafting for compositions) requires us knowing an exact use count, rather than merely 'zero, one or many'.

kozross:
Closes #1025 . This allows some significant simplifications in parts of `AST.hs`, and arguably has a clearer name to boot.


gnumonik (approved):
This is all fine

PR #1027, opened 2026-07-21
Data derivations

### LINKED ISSUE #1024: Data-encoding TH derivation
Follow-on from #1022 .

kozross:
Closes #1024 , depends on #1022 .

### DIFF
--- a/Plutarch/TH/DataPlutus.hs
+++ b/Plutarch/TH/DataPlutus.hs
@@ -7,17 +7,21 @@ module Plutarch.TH.DataPlutus (
 
 import Control.Monad (replicateM)
 import Data.Foldable (foldl', foldrM, for_, traverse_)
+import Data.Traversable.WithIndex (itraverse)
 import Data.Vector (Vector)
 import Data.Vector qualified as Vector
 import Data.Vector.NonEmpty qualified as NEVector
 import Language.Haskell.TH (
   Bang,
   BndrVis,
+  Body (NormalB),
   Con (InfixC, NormalC, RecC),
   Dec,
-  Exp (AppE, ConE, LamE, VarE),
+  Exp (AppE, CaseE, ConE, LamE, LitE, VarE),
+  Lit (IntegerL),
+  Match (Match),
   Name,
-  Pat (VarP),
+  Pat (ConP, VarP),
   Q,
   TyVarBndr,
   Type (AppT, ConT),
@@ -27,11 +31,13 @@ import Plutarch.Backend.Term (plam')
 import Plutarch.Primitive.Apply (PlutarchType (PRepresentation))
 import Plutarch.Primitive.BuiltinFun (pheadList)
 import Plutarch.Primitive.CanData (PCanData)
+import Plutarch.Primitive.Con (PCon (pcon'))
 import Plutarch.Primitive.Data (PAsData, PData)
 import Plutarch.Primitive.Eq (PEq (peq))
 import Plutarch.Primitive.Match (PMatch (pmatch'))
 import Plutarch.Primitive.Pair (PBPair (PBPair))
 import Plutarch.TH.Helpers (conToName, fullTypeName, getArity, mkContextOf)
+import PlutusCore qualified as PLC
 
 -- | @since wip
 deriveDataPlutus :: Vector (TyVarBndr BndrVis) -> Name -> Vector Con -> Q [Dec]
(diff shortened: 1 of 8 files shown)


Plutarch/TH/DataPlutus.hs
```diff
@@ -0,0 +1,200 @@
...
+      handlers <- foldrM (\e acc -> [e|NEVector.cons (toSomeTerm $(pure e)) $(pure acc)|]) start handlersRest
+      [e|punsafeCase $(pure (VarE tagName)) $(pure handlers)|]
+    mkHandler :: Name -> Name -> (Name, Word) -> Q Exp
+    mkHandler fieldsName contName (conName, arity) = case arity of
+      0 -> [e|$(pure (VarE contName)) $(pure (ConE conName))|]
+      1 -> [e|$(pure (VarE contName)) ($(pure (ConE conName)) (punsafeCoerce $(pure (VarE fieldsName))))|]
+      _ -> do
+        headTails <- replicateM (fromIntegral $ arity - 1) ((,) <$> newName "h" <*> newName "t")
+        argEs <- toArgEs headTails
+        let final = AppE (VarE contName) . foldl' AppE (ConE conName) $ argEs
```
gnumonik, line 119: Not super important: If you're going to (have to?) write out the TH AST directly I think you should have a comment indicating what the expression is supposed to look like in the splice. It makes it easier to review and modify in the future.

Plutarch/TH/DataPlutus.hs
```diff
@@ -0,0 +1,200 @@
...
+    toArgEs = \case
+      [] -> pure []
+      [(h, t)] -> (:) <$> [e|punsafeCoerce $(pure (VarE h))|] <*> ((: []) <$> [e|punsafeCoerce (pheadList @PData # $(pure (VarE t)))|])
+      (h, _) : rest -> (:) <$> [e|punsafeCoerce $(pure (VarE h))|] <*> toArgEs rest
+    go :: Exp -> Name -> [(Name, Name)] -> Q Exp
+    go acc remaining = \case
+      [] -> pure acc
+      (h, t) : rest -> do
+        let plams =
+              AppE (VarE 'plam')
```
gnumonik, line 149: See previous comment.

Plutarch/TH/DataPlutus.hs
```diff
@@ -0,0 +1,200 @@
...
+    ctx = pure . mkContextOf ''PCanData $ tyVars
+
+checkFieldIsWrapped :: Con -> Q ()
+checkFieldIsWrapped = \case
+  NormalC name fields -> for_ fields (go name)
+  RecC name fields -> for_ fields (goNamed name)
+  InfixC lhs name rhs -> do
+    go name lhs
+    go name rhs
+  _ -> fail "Unexpected constructor type found. If you see this message, report as a bug."
```
gnumonik, line 206: I think this is fine, but IIRC (you might want to check, this is based on a very old recollection), the "flavor" of constructor is fully determined syntactically, so that if you defined a type like 

```
data SillyGadt where 
  SillyOne :: Int -> SillyGadt 
  SillyTwo :: SillyGadt
```

It would read as a GADT to TH even though it doesn't _need_ to be one. I'm pretty sure this is how it worked at some point at least. 

In theory you could make this work for "constructors of GADTs that don't actually need to be GADTs". It's _probably_ not worth the effort. However, since I think that HLint (or HLS? one of the two) sometimes suggests GADT syntax where you don't need it, it might be good to at least add a more specific warning for the `RecGadtC` and `GadtC` cases telling them to rewrite it as a normal data type. 

(If you do that somewhere else that I haven't gotten to yet then ignore this. I'm guessing you check for GADT-ness elsewhere independently of this function and the error should be wherever that is)

gnumonik: If I remember right, it likes to suggest GADT syntax for Plutarch type declarations in particular.

gnumonik (approved):
A few small suggestions but nothing incredibly substantive. I'll leave it up to you whether to make any changes.

Plutarch/TH/DataPlutus.hs
kozross: That's a _very_ tricky check to do. I would have to dig through the GADT declaration to check if it traps any constraints. Furthermore, this can be quite subtle, as it might do so by using variables that don't occur in the tyvars bound by the type constructor.

HLint makes no such suggestion IME - in fact, it actually does the _opposite_. If HLS suggests GADT syntax in inappropriate places, it's a bug in HLS, and I don't think we have to work around it.

PR #1030, opened 2026-07-22
Enum derivation strategy

### LINKED ISSUE #1028: Enum TH derivation
Follow-on from #1024. This would allow types with no fields in any 'arm' to encode as `Integer`s.

kozross:
Follow-on from #1024 , closes #1028 . Also fixes a bug in the `default` method implementations for `PCanData`.

### DIFF
--- a/Plutarch/Primitive/CanData.hs
+++ b/Plutarch/Primitive/CanData.hs
@@ -57,7 +57,7 @@ class PlutarchType a => PCanData (a :: S -> Type) where
     forall (s :: S).
     PCanData (PRepresentation a) =>
     Term s (PAsData a) -> Term s a
-  pfromData = punsafeCoerce
+  pfromData = punsafeCoerce . pfromData @(PRepresentation a) . punsafeCoerce
   ptoData ::
     forall (s :: S).
     Term s a -> Term s (PAsData a)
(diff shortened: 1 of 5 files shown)


Plutarch/Primitive/BuiltinFun.hs
```diff
@@ -397,6 +404,13 @@ pmkPairData ::
   Term s (PData :--> PData :--> PBPair PData PData)
 pmkPairData = punsafeBuiltin PLC.MkPairData
 
+-- | @since wip
+pnilData ::
+  forall (a :: S -> Type) (s :: S).
+  PData `PCanRepresent` a =>
+  Term s (PBList a)
+pnilData = papp (punsafeBuiltin PLC.MkNilData) (punsafeConstant . PLC.someValue @() $ ())
```
gnumonik: If we pre-evaluate this doesn't matter, but: Wouldn't it be better to lift a constant empty list? I had to check the denotation in Plutus to remember why it needed the `()` and [IOG](https://github.com/IntersectMBO/plutus/blob/f3ac51d95aec5317ee84b8a88de2f100ddf11425/plutus-core/plutus-core/src/PlutusCore/Default/Builtins.hs#L1858-L1860) seems to say there's no reason to use the builtin ever.

gnumonik (approved):
No major issues, feel free to respond to my one comment if you like but if I'm not around or you're short on time then it's fine to merge afaict.

Plutarch/Primitive/BuiltinFun.hs
kozross: That's a good point actually. Easy fix too.

PR #1032, opened 2026-07-22
DataList derivation strategy

### LINKED ISSUE #1031: DataList derivation strategy
Follow-on from #1028 . This will encode one-'arm' data types with `PAsData` wrapped fields as a builtin list.

kozross:
Closes #1031 . The `PMatch`-generating code has also been simplified considerably in both the new derivation and the one for `DataPlutus`.


gnumonik (approved):
As far as I can tell, everything here is solid.

PR #1035, opened 2026-07-26
Mogensen-Scott derivation

### LINKED ISSUE #1033: Mogensen-Scott derivation strategy
This is made more complicated by the fact that the RHS of a type instance cannot have `forall`s. For example, the following will not compile:

```haskell
data PMaybe (a :: S -> Type) (s :: S) = 
    PNothing | 
    PJust (Term s a)

instance PlutarchType (PMaybe a) where
    -- This will complain about an illegal polymorphic type
    type PRepresentation (PMaybe a) = forall r . (r :--> (a :--> r) :--> r)
```

We can avoid this problem in two ways:

* 'Hide' the `forall` behind a `newtype` (similarly to how `PForall` does it). This is horribly unreadable and quite confusing.
* Make any MS-encoded type fundamental, and then handle its representation by a lambda implicitly. This is a bit less confusing, but has potential to go a lot more wrong.

kozross:
Closes #1033 .

### DIFF
--- a/Plutarch/TH/Helpers.hs
+++ b/Plutarch/TH/Helpers.hs
@@ -14,6 +14,7 @@ module Plutarch.TH.Helpers (
   mkUncons,
   conToFieldTypes,
   mkPLam,
+  mkAnonPLam,
 ) where
 
 import Data.Foldable (foldl', for_)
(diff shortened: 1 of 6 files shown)


gnumonik (approved):
As far as I can tell this is OK (but ugh it's really hard to be sure with TH)

PR #1037, opened 2026-07-26
Fix TH visibility issues in derivations

### LINKED ISSUE #1036: Fix TH derivation code to not require in-scope identifiers from `Backend` and `Primitive` modules
Currently, some derivations require 'dragging in' code that really ought not to be visible globally. This is a side effect of using `[d|` and `[e|` quasiquoters, as these do not scope names that are visible locally at time of definition. 

This can be fixed in a somewhat tedious way by using manual splices of the relevant identifiers, at the cost of some readability.

kozross:
Closes #1036 . #1035 should merge before this, as the MS derivations haven't been dealt with.

### DIFF
--- a/Plutarch/TH/DataList.hs
+++ b/Plutarch/TH/DataList.hs
@@ -37,6 +37,9 @@ import Plutarch.TH.Helpers (
   hasNoFields,
   mkContextOf,
   mkUncons,
+  pmkConsE,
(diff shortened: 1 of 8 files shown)


gnumonik (approved):
Don't see how any changes here could break anything, pretty straightforward

PR #1042, opened 2026-07-28
Change result type of pgcd

### LINKED ISSUE #1041: `pgcd`'s result type is too general
Given that `pgcd` must be called with at least one nonzero argument, it cannot ever return `0`. Thus, its return type should reflect this: it's currently `a` and should be `PNonZero a`.

kozross:
Closes #1041 . This newfound honesty in the type signature also avoids `punsafeCoerce` use, as well as being clearer regarding the constraints that an instance should follow.


gnumonik (approved):
Looks good

PR #1043, opened 2026-07-29
Make MS derivations generalize in tyvars

### LINKED ISSUE #1040: Mogensen-Scott derivation strategy doesn't play nice with rep polymorphism
Consider the following data type definition:

```haskell
data PMaybe a s = PNothing | PJust (Term s a)

deriveFor ''PMaybe MogensenScott
```

Logically, it stands to reason that `PRepresentation (PMaybe PNatural) ~ PMaybe PInteger`. However, this isn't currently the case: the `MogensenScott` derivation strategy makes every type it's used on fundamental.

kozross:
Closes #1040 .

### DIFF
--- a/Plutarch/TH/MS.hs
+++ b/Plutarch/TH/MS.hs
@@ -27,7 +27,7 @@ import Language.Haskell.TH (
   Pat (ConP, VarP, WildP),
   Q,
   TyVarBndr,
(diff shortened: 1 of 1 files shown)


gnumonik (approved):
This one also looks fine.

PR #1044, opened 2026-07-29
Pretty RawTerm

gnumonik:
Subsumes the open "Prettier UPLC" PR and substantially reworks its contents, so we should probably just merge this and delete that.

### DIFF
--- a/Plutarch/Backend/Term.hs
+++ b/Plutarch/Backend/Term.hs
@@ -69,6 +69,8 @@ import Data.Can (Can (Eno, Non, One, Two))
 import Data.Kind (Type)
 import Data.List (find)
 import Data.Map.Merge.Strict (WhenMatched, zipWithAMatched)
+import Data.Set (Set)
+import Data.Set qualified as Set
 import Data.Text (Text)
 import Data.These (These (That, These, This))
 import Data.These.Combinators (justHere, justThere)
@@ -212,17 +214,20 @@ prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
     Right (varMap, term) -> case runRWS (withVarMap varMap $ go term) mempty 0 of
       (res', _, _) -> res'
   where
+    mapMaybeSet :: forall (a :: Type). Ord a => (a -> Maybe a) -> Set a -> Set a
+    mapMaybeSet f = Set.foldl' (\acc x -> case f x of Just y -> Set.insert y acc; _ -> acc) mempty
+
     withVarMap ::
       forall (r :: Type).
       VarMap ->
-      RWS (Vector (PosTree, Text)) () Integer r ->
-      RWS (Vector (PosTree, Text)) () Integer r
+      RWS (Set (PosTree, Text)) () Integer r ->
+      RWS (Set (PosTree, Text)) () Integer r
     withVarMap vm act = do
       let posTrees = vmFold (\acc _ new -> Vector.snoc acc new) mempty vm
-      namedTrees <- Vector.mapM (\pt -> nextVarName >>= \nm -> pure (pt, nm)) posTrees
+      namedTrees <- Set.fromList . Vector.toList <$> Vector.mapM (\pt -> nextVarName >>= \nm -> pure (pt, nm)) posTrees
       local (namedTrees <>) act
 
-    nextVarName :: RWS (Vector (PosTree, Text)) () Integer Text
+    nextVarName :: RWS (Set (PosTree, Text)) () Integer Text
     nextVarName = do
       s <- gets compactReadableVar
       modify' (+ 1)
@@ -232,22 +237,22 @@ prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
       forall (a :: Type).
       PosTree ->
       (PosTree -> Maybe PosTree) ->
-      RWS (Vector (PosTree, Text)) () Integer a ->
-      RWS (Vector (PosTree, Text)) () Integer (Doc ann, a)
+      RWS (Set (PosTree, Text)) () Integer a ->
+      RWS (Set (PosTree, Text)) () Integer (Doc ann, a)
     withBoundVar pt localF cont = do
       varName <- nextVarName
       let pVar = pretty varName
-      (pVar,) <$> local (Vector.cons (pt, varName) . mapMaybe1 localF) cont
+      (pVar,) <$> local (Set.insert (pt, varName) . mapMaybe1 localF) cont
 
-    resolveThisVar :: RWS (Vector (PosTree, Text)) () Integer (Doc ann)
+    resolveThisVar :: RWS (Set (PosTree, Text)) () Integer (Doc ann)
     resolveThisVar = do
       pts <- ask
       case snd <$> find (\x -> fst x == PHere) pts of
         Nothing -> error "free variable in Term, this should be impossible (also: Implement better error handling in prettyTerm)"
         Just thisVar -> pure (pretty thisVar)
 
-    mapMaybe1 :: forall c d. (c -> Maybe c) -> Vector (c, d) -> Vector (c, d)
-    mapMaybe1 f = Vector.mapMaybe (\(x, y) -> (,) <$> f x <*> pure y)
+    mapMaybe1 :: forall (c :: Type) (d :: Type). (Ord c, Ord d) => (c -> Maybe c) -> Set (c, d) -> Set (c, d)
+    mapMaybe1 f = mapMaybeSet (\(x, y) -> (,) <$> f x <*> pure y)
 
     downL :: PosTree -> Maybe PosTree
     downL = \case
@@ -294,7 +299,7 @@ prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
       [Doc ann] ->
       Maybe PosTree ->
       RawTerm () ->
-      RWS (Vector (PosTree, Text)) () Integer ([Doc ann], Doc ann)
+      RWS (Set (PosTree, Text)) () Integer ([Doc ann], Doc ann)
     goLambda acc mpt = \case
       RLamAbs _ cMPT cBody -> case mpt of
         Nothing -> do
@@ -311,7 +316,7 @@ prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
           (thisVar, body) <- withBoundVar pt downOne $ go otherBody
           pure (reverse (thisVar : acc), body)
 
-    goApp :: RawTerm () -> RWS (Vector (PosTree, Text)) () Integer (NonEmptyVector (Doc ann))
+    goApp :: RawTerm () -> RWS (Set (PosTree, Text)) () Integer (NonEmptyVector (Doc ann))
     goApp = \case
       RApply _ f arg -> do
         xs <- local (mapMaybe1 downL) $ goApp f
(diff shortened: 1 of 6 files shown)


Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
 
 type role Term nominal representational
 
+instance Pretty (Term s a) where
+  pretty = prettyTerm
+
+prettyTerm :: forall (s :: S) (px :: S -> Type) (ann :: Type). Term s px -> Doc ann
+prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
+  (res, _, _) -> case res of
+    Left e -> error (show e)
```
kozross: I think in this situation it might be better to indicate the error in a prettyprinted way, rather than just throwing up with `error`.

Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
...
+
+    resolveThisVar :: RWS (Vector (PosTree, Text)) () Integer (Doc ann)
+    resolveThisVar = do
+      pts <- ask
+      case snd <$> find (\x -> fst x == PHere) pts of
+        Nothing -> error "free variable in Term, this should be impossible (also: Implement better error handling in prettyTerm)"
+        Just thisVar -> pure (pretty thisVar)
+
+    mapMaybe1 :: forall c d. (c -> Maybe c) -> Vector (c, d) -> Vector (c, d)
+    mapMaybe1 f = Vector.mapMaybe (\(x, y) -> case f x of Just x' -> Just (x', y); _ -> Nothing)
```
kozross: Isn't this just `Vector.mapMaybe (\(x, y) -> (,) <$> f x <*> pure y)`?

Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
...
+
+    downL :: PosTree -> Maybe PosTree
+    downL = \case
+      PTwo xs -> case xs of
+        This l -> pure l
+        That _ -> Nothing
+        These l _ -> pure l
```
kozross: This is `justThis`.

Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
...
+      _ -> Nothing
+
+    downR :: PosTree -> Maybe PosTree
+    downR = \case
+      PTwo xs -> case xs of
+        This _ -> Nothing
+        That r -> pure r
+        These _ r -> pure r
```
kozross: This is `justThat`.

Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
...
+    isSmall = \case
+      RVar {} -> True
+      RConstant {} -> True
+      RError {} -> True
+      RBuiltin {} -> True
+      RDelay _ inner -> isSmall inner
+      RForce _ inner -> isSmall inner
+      _ -> False
+
+    goLambda :: [Doc ann] -> Maybe PosTree -> RawTerm () -> RWS (Vector (PosTree, Text)) () Integer ([Doc ann], Doc ann)
```
kozross: Break this across several lines.

Plutarch/Utils/Pretty.hs
```diff
@@ -123,19 +136,45 @@ compactReadableVar n
...
 
--- first two args are the left and right block separator
+_customLine :: Doc ann
+_customLine = flatAlt hardline ""
+
+-- arg is a "tag" for the AST node type, next two two args are the left and right block separator
+block' :: forall (ann :: Type). Doc ann -> Doc ann -> Doc ann -> Doc ann -> Doc ann
+block' lbl l' r d = group $ flatAlt multiLine oneLine
+  where
+    l = l' <> lbl
```
kozross, line 149: Signature please.

Plutarch/Utils/Pretty.hs
```diff
@@ -151,15 +190,143 @@ prettyAnnotated getAnn annHandler nodePrinter x = annHandler (pretty a) $ nodePr
...
+        l
+          <> cxt
+          <> hardline
+          <> indent 2 body
+          <> myLine
+          <> r
+    oneLineNoParens = "\\" <> hsep vars <+> "->" <+> body
+
+-- | Don't pass an empty list into this
+appLike :: forall (ann :: Type). Doc ann -> PrintMode -> Bool -> [Doc ann] -> Doc ann
```
kozross: Any chance this could take a `NonEmpty`?

Plutarch/Utils/Pretty.hs
```diff
@@ -151,15 +190,143 @@ prettyAnnotated getAnn annHandler nodePrinter x = annHandler (pretty a) $ nodePr
...
+          <+> var
+          <+> "="
+          <+> align (group bind)
+          <> hardline
+          <> "in"
+          <+> align (group body)
+          <> myLine
+          <> r
+
+    oneLineNoParens = "let" <+> var <+> "=" <+> bind <+> "in" <+> body
```
kozross, line 290: Signatures.

Plutarch/Utils/Pretty.hs
```diff
@@ -151,15 +190,143 @@ prettyAnnotated getAnn annHandler nodePrinter x = annHandler (pretty a) $ nodePr
...
+        else
+          align . group $
+            l
+              <> "case"
+              <+> hardline
+              <> indent 2 (align . vcat $ [scrut, customList handlers])
+              <> myLine
+              <> r
+
+    oneLineNoParens = "case" <+> scrut <+> oneLineList handlers
```
kozross, line 322: Signatures.

Plutarch/Backend/Term.hs
```diff
@@ -192,6 +200,175 @@ newtype Term (s :: S) (a :: S -> Type)
 
 type role Term nominal representational
 
+instance Pretty (Term s a) where
+  pretty = prettyTerm
+
+prettyTerm :: forall (s :: S) (px :: S -> Type) (ann :: Type). Term s px -> Doc ann
+prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
+  (res, _, _) -> case res of
```
kozross, line 212: What happens if someone requests a prettyprint of a `Term` that isn't closed? You have to make use of the `VarMap` that its computation produces to prettyprint it, because otherwise, you won't know what the free vars are called (or even where they are).

kozross (changes requested):

Plutarch/Utils/Pretty.hs
gnumonik: Yah it can

Plutarch/Backend/Term.hs
```diff
@@ -206,10 +208,20 @@ instance Pretty (Term s a) where
 prettyTerm :: forall (s :: S) (px :: S -> Type) (ann :: Type). Term s px -> Doc ann
 prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
   (res, _, _) -> case res of
-    Left e -> error (show e)
-    Right (_, term) -> case runRWS (go term) mempty 0 of
+    Left e -> "Encountered an error when attempting to pretty print a term: " <> pretty (show e)
+    Right (varMap, term) -> case runRWS (withVarMap varMap $ go term) mempty 0 of
```
kozross, line 214: Why not just change the reader environment from `mempty` to whatever the `mapM` in `withVarMap` would produce?

Plutarch/Backend/Term.hs
```diff
@@ -206,10 +208,20 @@ instance Pretty (Term s a) where
...
       (res', _, _) -> res'
   where
+    withVarMap ::
+      forall (r :: Type).
+      VarMap ->
+      RWS (Vector (PosTree, Text)) () Integer r ->
+      RWS (Vector (PosTree, Text)) () Integer r
+    withVarMap vm act = do
+      let posTrees = vmFold (\acc _ new -> Vector.snoc acc new) mempty vm
+      namedTrees <- Vector.mapM (\pt -> nextVarName >>= \nm -> pure (pt, nm)) posTrees
```
kozross: This presumes the var names in the varmap are both contiguous and start at 0. The second is a safe bet, the first is not.

Plutarch/Utils/Pretty.hs
```diff
@@ -210,8 +215,8 @@ lambdaTemplate mode vars body = case mode of
     oneLineNoParens = "\\" <> hsep vars <+> "->" <+> body
 
 -- | Don't pass an empty list into this
-appLike :: forall (ann :: Type). Doc ann -> PrintMode -> Bool -> [Doc ann] -> Doc ann
-appLike op mode funIsSmall funList@(fun : args) = case mode of
+appLike :: forall (ann :: Type). Doc ann -> PrintMode -> Bool -> NonEmptyVector (Doc ann) -> Doc ann
+appLike op mode funIsSmall funList@(NEVector.uncons -> (fun, args)) = case mode of
```
kozross: I'd prefer a `let` over a view pattern here.

Plutarch/Utils/Pretty.hs
```diff
@@ -224,33 +229,34 @@ appLike op mode funIsSmall funList@(fun : args) = case mode of
             l
               <+> fun
               <> hardline
-              <> indent 2 (vcat (map (op <+>) args))
+              <> indent 2 (vcat (map (op <+>) $ toList args))
```
kozross: If at all possible, it's better to use the `toList` provided by the specific concrete type than the `Foldable` type class method. It's probably not an issue here, but `toList` has a default in `Foldable` which might not be efficient (and in fact, probably isn't), and you're relying on the library author to properly overload it.

If you're working over an arbitrary `Foldable`, you can't avoid this, but in this case, you can and should.

test/Main.hs
```diff
@@ -427,6 +430,18 @@ main =
             let t = toUPLCTerm anf'
             step $ "UPLC:\n" <> toPrettyString t
             pure ()
+    , testCaseSteps "Case 20 (not a real test, always passes)" $ \step -> do
```
kozross: That's not technically true - it could fail to compile depending on what we mess up later.

kozross (changes requested):

Plutarch/Backend/Term.hs
gnumonik: I guess I could do that but I prefer to keep all of the monadic stuff in a single computation so that I don't have to "run" the monad and then pass the state right back into another monadic computation, which introduces the possibility for a mistake. 

I dunno, I don't have any strong feelings about this but I thought about doing what you suggested for 10 secs when I was writing it and then decided this way is marginally better due to eliminating the possibility for at least one mistake.

Plutarch/Backend/Term.hs
```diff
@@ -212,17 +214,20 @@ prettyTerm t = case runRWS (runExceptT (asRawTerm t)) TermEnv 0 of
...
       forall (r :: Type).
       VarMap ->
-      RWS (Vector (PosTree, Text)) () Integer r ->
-      RWS (Vector (PosTree, Text)) () Integer r
+      RWS (Set (PosTree, Text)) () Integer r ->
+      RWS (Set (PosTree, Text)) () Integer r
     withVarMap vm act = do
       let posTrees = vmFold (\acc _ new -> Vector.snoc acc new) mempty vm
-      namedTrees <- Vector.mapM (\pt -> nextVarName >>= \nm -> pure (pt, nm)) posTrees
+      namedTrees <- Set.fromList . Vector.toList <$> Vector.mapM (\pt -> nextVarName >>= \nm -> pure (pt, nm)) posTrees
```
kozross: You probably could do this as a `Vector.foldM` more clearly.

kozross (approved):

PR #1046, opened 2026-07-29
Drop plutarch-orphanage's duplicate Arbitrary1 NonEmptyList instance

colll78:
`plutarch-orphanage`'s `PlutusLedgerApi.V1.Orphans.Value` defines:

```haskell
-- This is frankly a bizarre omission
instance Arbitrary1 NonEmptyList where
```

QuickCheck has since agreed with that comment and added the instance to `Test.QuickCheck.Modifiers`. With a current QuickCheck both exist, so the local definition is a duplicate-instance error and `plutarch-orphanage` fails to compile.

This removes the local instance (and the now-unused `NonEmptyList (NonEmpty)` import) rather than CPP-guarding it, on the grounds that the upstream definition supersedes it and the two are semantically the same.

**Trade-off worth a maintainer's call:** this makes QuickCheck's instance load-bearing, so it raises the effective minimum QuickCheck version. If you would rather keep a lower bound, the alternative is a CPP guard plus a `build-depends` constraint — happy to switch to that if preferred.

Verified: Plutarch's full test suite passes (1063/1063) on GHC 9.6.6 with this applied.

### DIFF
--- a/plutarch-orphanage/src/PlutusLedgerApi/V1/Orphans/Value.hs
+++ b/plutarch-orphanage/src/PlutusLedgerApi/V1/Orphans/Value.hs
@@ -29,7 +29,6 @@ import Test.QuickCheck (
   CoArbitrary,
   Function (function),
   Gen,
-  NonEmptyList (NonEmpty),
   NonZero (NonZero),
   Positive (Positive),
   chooseBoundedIntegral,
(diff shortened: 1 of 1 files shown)


kozross:
@colll78 - thanks for that! I think we'll bump minimum version, as it seems this doesn't cause you any compilation issues.

plutarch-orphanage/src/PlutusLedgerApi/V1/Orphans/Value.hs
```diff
@@ -309,19 +308,11 @@ instance Function PLA.TokenName where
...
-  {-# INLINEABLE liftShrink #-}
-  liftShrink shrinkInner (NonEmpty ell) =
-    NonEmpty <$> case ell of
-      [] -> []
-      (x : xs) -> (:) <$> shrinkInner x <*> liftShrink shrinkInner xs
+-- Note: this module used to define `instance Arbitrary1 NonEmptyList`, calling
+-- its absence "frankly a bizarre omission". QuickCheck now provides that
+-- instance itself (Test.QuickCheck.Modifiers), so keeping this one is a
+-- duplicate-instance error. Removed rather than CPP-guarded because the
+-- upstream definition supersedes it.
```
kozross, line 315: This comment can go I think.

kozross (changes requested):

PR #1055, opened 2026-08-03
Helpers cleanup

### LINKED ISSUE #1052: Consolidate all helpers into a dedicated module group
Currently, helper modules are scattered all over the place. We should put them in a single dedicated location.

kozross:
Closes #1052


gnumonik (approved):
Personally I like 'Utils' more than 'Helpers' cuz 2 chars shorter but we can go with 'Helpers' if you like that more.

PR #1057, opened 2026-08-03
Pin QuickCheck to 2.18 at least, remove duplicate instance

kozross:
Supercedes #1046 .


gnumonik (approved):
On the assumption that you have good reasons for the pin there's nothing here that could really be wrong.

PR #1068, opened 2026-08-05
Evaluation golden support

### LINKED ISSUE #1053: Simplified output goldens
Follow-on from #1048 . Currently, goldens output all of the following:

* `Term`
* AST
* ANF both before and after demand analysis
* UPLC

This is useful for us, but less useful for a Plutarch user. Furthermore, the current goldens are non-evaluating, which is again fine for compiler development, but less fine for script development.

We need a 'simplified' version, which produces the following:

* `Term`
* UPLC pre-evaluation
* UPLC post-evaluation

The evaluation budget should be as large as possible here - the goal is to see what happens, not whether it'll fit.

kozross:
Closes #1053 . Also fixes placement of prettyprinting helpers that got overlooked in #1052 .


Plutarch/Test/Golden.hs
```diff
@@ -79,6 +80,52 @@ plutarchGolden testDescription testName t =
...
+* The 'Term' itself;
+* The UPLC resulting from compiling the 'Term'; and
+* The UPLC resulting from evaluating the compiled 'Term'.
+
+All but the first will produce an error if the 'Term' fails to compile, and
+the last will produce an error if the compiled 'Term' fails to evaluate.
+
+= Important note
+
+The caveats regarding naming given for 'plutarchGolden' also apply to this
```
gnumonik, line 96: This isn't important but what happens if the names do clash? Do both tests even get run?

gnumonik (approved):
This seems fine. My only question is somewhat tangential and shouldn't block merging.

PR #1069, opened 2026-08-06
Tracing support

### LINKED ISSUE #1066: Port tracing support
Currently, tracing is not possible at all. This requires a few pieces:

* Support in `TermEnv` to set a tracing level
* Tracing primitives that can be disabled by tracing level

kozross:
Closes #1066 . Also extends `plutarchGolden` and `plutarchGoldenEval` with versions that allow choosing the `TermEnv`.


gnumonik (approved):
Seems fine to me.

PR #1076, opened 2026-08-10
Optimizations

### LINKED ISSUE #1072: Add the option to perform UPLC optimizations using Plutus Core's helper
Follow-on from #1069 . The original backend includes this capability, and we should too, as a post-processing pass. This should be flag-controlled, as in some cases (such as when debugging), we may want to have the UPLC output correlate closely with what we wrote.

kozross:
Closes #1072 .


Plutarch/Backend/Term.hs
```diff
@@ -158,19 +159,57 @@ data TracingMode
...
+Currently, \'external optimizations\' consist of the following:
+
+* Pre-evaluation of the generated code; followed by
+* The UPLC optimizer provided in Plutus Core.
+
+These are labelled as \'external\' as Plutarch does not control them.
+
+@since wip
+-}
+data OptimizationMode = OnlyInternal | InternalExternal
```
gnumonik, line 173: I'm guessing this will evolve into something more sophisticated once we have more optimizations? These aren't the most intuitive constructor names for optimization options but that's probably fine b/c it'll change later.

gnumonik (approved):
I think this is all sensible. We probably want a more sophisticated interface for configuring optimizations at some point, but obviously it wouldn't have much of a point now.

PR #1078, opened 2026-08-10
plutarchGoldenAll

### LINKED ISSUE #1077: Goldens of two different types targeting the same files clash
This can be most clearly seen when you have `plutarchGolden` and `plutarchGoldenEval` running in parallel: tests fail due to the underlying file being held by one with the other trying to concurrently write to it. This should be fixed by either allowing parameterization of exactly which stage(s) to run, or by having a 'combined' version that does everything.

kozross:
Closes #1077 . Also replaces all existing golden tests using _both_ `plutarchGolden` and `plutarchGoldenEval` to `plutarchGoldenAll`.

### DIFF
--- a/Plutarch/Test/Golden.hs
+++ b/Plutarch/Test/Golden.hs
@@ -3,6 +3,7 @@
 module Plutarch.Test.Golden (
   plutarchGolden,
   plutarchGoldenEval,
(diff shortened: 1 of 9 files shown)


gnumonik (approved):
Seems fine.

PR #1080, opened 2026-08-16
Improved demand analysis

### LINKED ISSUE #1023: Introduce a notion of 'hard' and 'soft' dependencies to demand analysis
Follow-on from #1022 . Due to the changes required to fix the demand analysis bug present in #1022, closed sub-computations are now lifted 'as high as possible', meaning the toplevel. Furthermore, currently, demand analysis assumes that _every_ dependency represented in the ANF is 'hard', meaning 'always required'. 

This can potentially cause issues. Consider the following Plutarch:

```haskell
plam' $ \b -> pif b branch1 branch2
```

Furthermore, suppose that `branch1` depends on (and uses multiple times) a closed subcomputation `s1`, and `branch2` depends on (and uses multiple times) a closed subcomputation `s2`, and neither of these dependencies is shared between both `branch`es. Currently, we would generate the equivalent of 

```
\b -> let s1Bind = s1
          s2Bind = s2
        in case b [branch1, branch2]
```

After a `let`-transform, we get

```
\b -> (\s1Bind s2Bind -> case b [branch1, branch2]) s1 s2
```

As UPLC is strict, this now forces evaluation of _both_ `s1` and `s2`. This is excessive: we know that, depending on `b`, we will only ever need _one_ of `s1` and `s2`. A better solution would be to compile 

```
\b -> pif b (let s1Bind = s1 in branch1) (let s2Bind = s2 in branch2)
```

which would generate

```
\b -> case b [(\s1Bind -> branch1) s1, (\s2Bind -> branch2) s2]
```

We can see (and do) this here because we know that `branch1` and `branch2` are 'soft' dependencies: we _may_ need to evaluate them, but we don't necessarily. Thus, demand analysis should only hoist 'up to softness'.

kozross:
Closes #1023 . In the end, 'hard' and 'soft' dependency analysis was unnecessary, as a join-point-based approach was easier.


Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
       Nothing -> acc
-      Just site -> min acc site
+      Just (Id site) -> min acc site
+    digDependency :: Id -> Int -> Id
+    digDependency target currId =
+      let currBind = binds NEVector.! currId
+       in case currBind of
+            -- We should absolutely never make it down here. Give it a default
+            -- for the completeness checker.
+            ANFLeaf _ -> Id currId
```
gnumonik: Why not throw an error if it's something that shouldn't ever happen? Then, if it happens, we know something is going wrong (vs having to determine that it's defaulting to some value it shouldn't have). 

(I guess this is a judgment call based on exactly _how_ impossible you judge this case to be, which is a bit hard for me to evaluate w/o spending more time reviewing)

Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
+            -- The _first time_ we encounter Case 2 represents the point at
+            -- which the `let`-bind of `target` _must_ happen. If we see a Case
+            -- 1, the descendant (or possibly one of its descendants) is instead
+            -- a better choice.
+            ANFApply _ f xs ->
+              let start = refToJPR target f
+               in case NEVector.foldl' (combineJPRs target) start xs of
+                    -- This cannot happen: at least one of the descendants must
+                    -- depend on `target` somehow.
+                    NoCandidateFound -> Id currId
```
gnumonik: See previous comment

Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
+                    Ineligible -> Id currId
+                    Descend next -> digDependency target next
+            ANFConstr _ _ fields -> case Vector.uncons fields of
+              -- This cannot happen: at least one of the descendants must depend
+              -- on `target` somehow.
+              Nothing -> Id currId
+              Just (field, fields') -> case Vector.foldl' (combineJPRs target) (refToJPR target field) fields' of
+                -- This cannot happen: at least one of the descendants must
+                -- depend on `target` somehow.
+                NoCandidateFound -> Id currId
```
gnumonik: ibid

Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
+                Ineligible -> Id currId
+                Descend next -> digDependency target next
+            ANFCase _ scrut handlers ->
+              let start = refToJPR target scrut
+               in case NEVector.foldl' (combineJPRs target) start handlers of
+                    -- This cannot happen: at least one of the descendants must
+                    -- depend on `target` somehow.
+                    NoCandidateFound -> Id currId
```
gnumonik: ibid

Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
+                    Ineligible -> Id currId
+                    Descend next -> digDependency target next
+            ANFCompose _ components -> case NEVector.uncons components of
+              (c, cs) -> case Vector.foldl' (combineJPRs target) (refToJPR target c) cs of
+                -- This cannot happen: at least one of the descendants must
+                -- depend on `target` somehow.
+                NoCandidateFound -> Id currId
```
gnumonik: ibid

Plutarch/Backend/ANF.hs
```diff
@@ -409,38 +408,295 @@ data Demand
...
+        traverse_ (updateCountAt countMV) xs
+      ANFConstr _ _ fields -> traverse_ (updateCountAt countMV) fields
+      ANFCase _ scrut handlers -> do
+        updateCountAt countMV scrut
+        traverse_ (updateCountAt countMV) handlers
+      ANFCompose _ components -> traverse_ (updateCountAt countMV) components
+  pure countMV
+  where
     smallEnoughToInline :: Some (ValueOf PLC.DefaultUni) -> Bool
     smallEnoughToInline = \case
```
gnumonik, line 700: This is kind of tangential, but are other constants not small enough to (necessarily benefit if we) inline?

I might have asked you this in a previous review, sorry if I'm repeating myself.

gnumonik: I vaguely remember mentioning empty lists here, surely we could inline those, right?

gnumonik (approved):
Do read the comments but feel free to merge if you don't think they need addressed.

Plutarch/Backend/ANF.hs
kozross: We could realistically inline almost any constant we wanted with little effort. The reason I went with these ones specifically stems from the old backend. Specifically, these exact cases were singled out for always inlining. I suspect code _size_ might have something to do with it.

gnumonik (approved):

Issue #1071, opened 2026-08-06
Parameterize derivations to allow skipping `PEq` and `PDebug`


kozross:
Follow-on from #1070 . Currently, all TH derivations unconditionally attempt to derive every type class they can, including `PEq`. This is fine for most data types, but for data types containing functions, `PEq` makes no sense. Likewise, `PDebug` doesn't make sense either.

This would introduce an additional 'control' data type similar to

```haskell
data DerivationType = AllDerivations | NoEqDebug
```

which would then be an additional argument to `deriveFor`. If given `AllDerivations`, types containing functions would error at compile time; otherwise, the derivations for `PEq` and `PDebug` would be skipped.

kozross:
Migrated to new repo.

Issue #1070, opened 2026-08-06
`PDebug`


kozross:
Follow-on from #1066 . This would be a type class akin to the older `PShow`, with autoderivation support.

PR #1092, opened 2026-08-27
Koz/1090

### LINKED ISSUE #1090: `PEq` derivation is not sensitive to fixpoint types
Consider the following:

```haskell

data PSOPList (a :: S -> Type) (s :: S) = 
   PSOPNil |
   PSOPCons (Term s a) (Term s (PSOPList a))
```

The current derived instance will blindly recurse on the head-and-tail, which means it will generate a term that will loop codegen forever. A similar problem would arise with any `Show`-like thing we want to derive.

This means the Eq generator TH has to perform a check for recursion, and, if detected, should insert the appropriate `pfix` machinery. This is also a good time to verify a type isn't _negative_ recursive, because that shouldn't even be allowed to begin with.

kozross:
Closes #1090. Currently, function types error instead of producing derivations. This is going to be fixed, together with `PDebug` support, in #1071 once #1070 is dealt with.


Plutarch/TH/DataPlutus.hs
```diff
@@ -97,18 +105,25 @@ derivePMatch tyVars tyName cs c =
...
       -- If we have a minimum required number of fields, we extract them _now_,
       -- to avoid having repeated code in our handlers doing the same thing.
       let minArity = Vector.foldl' min arityLast aritiesRest
       -- Offset all arities by the number we 'preload'
       let arityLast' = arityLast - minArity
-      let aritiesRest' = fmap (\x -> x - minArity) aritiesRest
+      let aritiesRest' = fmap (- minArity) aritiesRest
       withPreload contName tagName (arityLast', nameLast) (aritiesRest', namesRest) [] fieldsName minArity
-    withPreload :: Name -> Name -> (Word, Name) -> (Vector Word, Vector Name) -> [Name] -> Name -> Word -> Q Exp
+    withPreload ::
```
gnumonik, line 158: Could you say what this does? It has a _lot_ of _very uninformatively typed_ arguments and the name doesn't make it clear and it's hard enough to read TH even when it's meticulously documented

Plutarch/TH/MS.hs
```diff
@@ -51,18 +55,20 @@ import Plutarch.Primitive.Function ((:-->))
...
-  Just (cs, c) ->
-    if Vector.all hasNoFields constructors
-      then fail "Use the Enum strategy for types with no fields in any 'arm'."
-      else do
-        plutarchTypeDec <- derivePlutarchType tvbs name
-        pmatchDec <- derivePMatch tvbs name cs c
-        pconDec <- derivePCon tvbs name constructors
-        peqDec <- derivePEq tvbs name constructors
-        pure $ plutarchTypeDec <> pmatchDec <> pconDec <> peqDec
+deriveMS :: Vector (TyVarBndr BndrVis) -> Name -> PTypeSum -> Q [Dec]
```
gnumonik, line 73: Could you put an example of a simple MS encoded data type in the comments here? Maybe this is something that'd only be useful to me while reviewing, but having something to refer to would make this easier to follow.

Plutarch/TH/Strategy.hs
```diff
@@ -83,15 +83,15 @@ deriveFor tyName strat = do
...
-          DataPlutus -> deriveDataPlutus tvbs name consAsVec
-          Enum -> deriveEnum tvbs name consAsVec
-          DataList -> deriveDataList tvbs name consAsVec
-          MogensenScott -> deriveMS tvbs name consAsVec
+          Enum -> deriveEnum tvbs name typeStructure
+          DataList -> deriveDataList tvbs name typeStructure
+          DataPlutus -> deriveDataPlutus tvbs name typeStructure
+          SOP -> deriveSOP tvbs name typeStructure
+          MogensenScott -> deriveMS tvbs name typeStructure
     NewtypeD {} -> fail "Newtype derivations not supported at present."
```
gnumonik, line 95: Wait, why? There isn't any reason we can't support them, right? 

Though I guess we might want a newtype _strategy_ eventually and can defer supporting them until we have that (on the assumption that people may want onchain newtypes). 

We did remove all of the stuff that lets you do the "this is a newtype on chain" thing, right? (I.e. where the Plutarch repr is just the repr of the underlying type). We should probably support that somehow eventually.

gnumonik (commented):
Please explain what preloading actually does. The other comments aren't super important.

Plutarch/TH/Strategy.hs
kozross: Agreed, but that's something I wanted to kick down the road.

gnumonik (approved):
Seems good now.

